PR #275, opened 2026-01-29
Fix race conditions in global chat context initialization

2witstudios:
## Summary
This PR fixes race conditions that occur during conversation loading by ensuring the global chat context is only synchronized when fully initialized, and by using current messages instead of initial messages in the chat configuration.

## Key Changes
- **GlobalAssistantView.tsx & SidebarChatTab.tsx**: Added `globalIsInitialized` guard to message sync effects to prevent overwriting context with stale messages from previous conversations during loading
- **GlobalChatContext.tsx**: Changed chat config to use current `messages` instead of `initialMessages` to prevent state snap-back issues when switching conversations from history
- Updated dependency arrays to include `globalIsInitialized` where applicable

## Implementation Details
The core issue was that message synchronization effects were firing before the global context was fully initialized, causing stale messages from previous conversations to overwrite the current state. By adding the `globalIsInitialized` check, we ensure synchronization only occurs after the context has loaded the correct conversation data.

Additionally, the chat configuration now references the current `messages` state rather than `initialMessages`. This ensures the useChat hook always has access to up-to-date messages without causing unwanted reinitializations, since the hook only reinitializes when the `id` (conversation ID) changes.

These changes work together to provide a more robust conversation switching experience without race conditions or state inconsistencies.

https://claude.ai/code/session_01RTEfhzuvQ8rieKauawENMd

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures messages sync only after the chat subsystem is fully initialized, preventing intermittent missing or out-of-order messages during conversation load.
  * Improves message-state handling when switching conversations and reduces unnecessary configuration updates, yielding more stable and consistent chat behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/web/src/components/layout/right-sidebar/ai-assistant/SidebarChatTab.tsx
+++ b/apps/web/src/components/layout/right-sidebar/ai-assistant/SidebarChatTab.tsx
@@ -137,7 +137,6 @@ const SidebarChatTab: React.FC = () => {
   // ============================================
   const {
     chatConfig: globalChatConfig,
-    messages: contextMessages, // Direct access to context messages for seamless display
     isStreaming: contextIsStreaming,
     stopStreaming: contextStopStreaming,
     setMessages: setGlobalContextMessages,


2witstudios:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #456, opened 2026-02-06
Simplify image attachment handling by removing blob URL management

2witstudios:
## Summary
This PR simplifies the image attachment handling in the AI chat by removing the manual blob URL lifecycle management. Instead of creating blob URLs for preview and managing their revocation, the implementation now uses data URLs directly throughout the process.

## Key Changes

- **Removed blob URL creation and tracking**: Eliminated `URL.createObjectURL()` calls and the `blobUrlsRef` tracking mechanism in `useImageAttachments` hook
- **Simplified preview URL handling**: Changed `previewUrl` to be set directly from the resized image's data URL instead of a temporary blob URL
- **Refactored image resize logic**: Modified `resizeImageForVision` to read the file as a data URL first via `FileReader`, then load it into an Image element, eliminating blob URL creation
- **Removed cleanup effects**: Deleted the `useEffect` cleanup hook that was revoking blob URLs on component unmount
- **React Strict Mode safe**: Side effects (resize calls, toasts) are outside `setState` updaters to prevent double-firing. Uses `countRef` for accurate state reads.
- **Updated test suite**: Removed blob URL mocking/assertions, strengthened weak assertions, removed trivially-passing tests
- **CUID2 IDs**: Attachment IDs now use `@paralleldrive/cuid2` per project convention

## How to Validate

1. Open any AI chat interface with image attachment support
2. Attach images — they should display previews after resize completes (no flash of empty preview)
3. Attach >5 images — should show appropriate toast messages
4. Remove individual images and clear all — should work without console errors
5. Send a message with image attachments — AI should receive the images correctly
6. No CSP violations in browser console (the original bug this fixes)

## Files Changed

| File | Change |
|------|--------|
| `apps/web/src/lib/ai/shared/utils/image-resize.ts` | FileReader instead of blob URLs |
| `apps/web/src/lib/ai/shared/hooks/useImageAttachments.ts` | Pure setState, no blob lifecycle, cuid2 IDs |
| `apps/web/src/lib/ai/shared/hooks/__tests__/useImageAttachments.test.ts` | Stronger assertions, no blob mocking |

### DIFF
--- a/apps/web/src/lib/ai/shared/hooks/__tests__/useImageAttachments.test.ts
+++ b/apps/web/src/lib/ai/shared/hooks/__tests__/useImageAttachments.test.ts
@@ -94,7 +94,7 @@ describe('useImageAttachments', () => {
     expect(result.current.attachments).toHaveLength(0);
   });
 
-  it('given clearFiles called, should remove all attachments and revoke all blob URLs', async () => {
+  it('given clearFiles called, should remove all attachments', async () => {
     const { result } = renderHook(() => useImageAttachments());
 
     await act(async () => {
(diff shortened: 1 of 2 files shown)


2witstudios:
## Addressed all CodeRabbit review feedback (commit 04c406a4)

### Major Issues Fixed

**1. Side effects inside `setState` updater (lines 31-70)** — Restructured `addFiles` to move all side effects (resize calls, toast notifications) outside the `setAttachments` updater. Added a `countRef` to read current attachment count without relying on updater closures (React 18 queues updaters for processing during the next render, so variables set inside updaters aren't available synchronously after `setState` returns). The updater is now pure: `(prev) => [...prev, ...newAttachments]`.

**2. Toast side effects inside `setState` updater (lines 31-36)** — Resolved as part of the restructure above. Toast calls now happen before `setAttachments` is called.

**3. Misleading `getFilesForSend` JSDoc (lines 86-100)** — Changed "Waits for any pending resizes to complete" to "Excludes any attachments still being processed (resizing)".

### Minor Issues Fixed

**4. Stale test description (line 97)** — Changed "should remove all attachments and revoke all blob URLs" to "should remove all attachments".

**5. Weak assertion in getFilesForSend test (lines 169-188)** — Replaced `expect(files.length).toBeGreaterThanOrEqual(0)` (always true) with `expect(files).toHaveLength(1)` and made all structural assertions unconditional.

### Nitpicks Addressed

**6. CUID2 for attachment IDs (lines 43-44)** — Replaced `Date.now()-Math.random()` with `createId()` from `@paralleldrive/cuid2` per project convention.

**7. Unmount test (lines 190-199)** — Removed the trivially-passing test that only checked `expect(() => unmount()).not.toThrow()`.

### Validation
- All 9 tests passing locally
- No lint errors in changed files  
- No type errors in changed files

PR #833, opened 2026-04-07
feat(auth): fix all desktop auth flows

2witstudios:
## Summary

- Fix all desktop auth paths (passkey login/signup, password, magic link, Google/Apple OAuth, Google One-Tap)
- Desktop Electron app couldn't persist sessions because auth routes only used Set-Cookie headers
- Refactored to eliminate unnecessary platform branching in JSON-returning routes

## Architecture

**Category 1 routes** (login, passkey, one-tap, signup-passkey) — ZERO platform branching:
- Single code path returns tokens (`sessionToken`, `csrfToken`, `deviceToken`) at top level in JSON body AND sets cookies
- Web client ignores body tokens, uses cookies. Desktop reads body tokens, stores via IPC.
- Uses `revokeSessionsForLogin()` for per-device revocation (passkey keeps `revokeAllUserSessions` — hardest reset)
- `createDeviceToken()` replaces `createWebDeviceToken()` with a `platform` param

**Category 2 routes** (OAuth callbacks, magic link verify) — branching only for delivery:
- Web: Set-Cookie + redirect
- Desktop: exchange code + `pagespace://auth-exchange` deep link
- `createDesktopSession()` now uses per-device revocation

## What changed

**Server:**
- `createDeviceToken()` — renamed from `createWebDeviceToken`, accepts `platform` param
- `createDesktopSession()` — updated to use per-device revocation (`revokeDeviceSessions`)
- All Category 1 routes unified: no `if (platform === 'desktop')` branches
- OAuth error paths verify HMAC signature before trusting platform field
- Magic-link schemas require `deviceId`/`deviceName` when `platform === 'desktop'`
- PII masking in `trackAuthEvent` calls

**Desktop Electron (3 files):**
- `auth:store-session` IPC sets cookie on BrowserWindow
- `auth:open-external` IPC opens OAuth in system browser (hostname allowlist)
- `deep-links.ts` handles `pagespace://auth-error` for OAuth failure notification

**Client:**
- `handleDesktopAuthResponse()` reads top-level tokens (`sessionToken`, `csrfToken`, `deviceToken`)
- Accepts both `redirectUrl` and `redirectTo` for redirect handling
- `getDevicePlatformFields()` for transparent request body setup
- `useOAuthSignIn` — system browser flow with waiting UI
- `useDesktopExchangeHandler` — picks up exchange params from magic-link redirects

## Test plan

- [x] All CI checks pass (lint, typecheck, unit tests, security)
- [x] 152+ automated auth tests pass
- [ ] Desktop passkey login — no 401s, session persists across restart
- [ ] Desktop password login — dashboard loads with Getting Started drive
- [ ] Desktop Google OAuth — system browser opens, app receives deep link
- [ ] Desktop magic link — both browser session and desktop token handoff work
- [ ] Web regression — all auth paths work as before


2witstudios:
## CodeQL \`js/user-controlled-bypass\` — Analysis & Mitigations

The CodeQL default-branch analysis flags 4 \`js/user-controlled-bypass\` findings on \`google/callback/route.ts\`. These are inherent to the OAuth2 callback pattern — the callback URL's query parameters (\`code\`, \`state\`) arrive via browser redirect from Google's authorization server, so they are by definition user-controllable inputs that gate the token exchange.

### What we changed to minimize the attack surface

1. **Removed \`error\` from all branch conditions** — the original finding. \`error\` is now only read as a UX hint inside an already-decided rejection branch. No user-supplied value acts as the security gate.

2. **Single combined guard**: \`if (!code || !verifiedState)\` — the sensitive path (token exchange) is only reachable when both an authorization code AND HMAC-verified state are present. \`verifiedState\` comes from \`verifyOAuthState()\`, a server-side cryptographic check using \`crypto.timingSafeEqual\`.

3. **PKCE (RFC 7636)** — the authorization code is useless without the server-held \`code_verifier\` stored in Redis. An attacker injecting \`?code=fake\` cannot complete the token exchange at Google's token endpoint.

4. **HMAC-signed state** — platform/device info is only trusted after signature verification with \`OAUTH_STATE_SECRET\`. Desktop deep-link redirects require valid HMAC.

5. **Input validation** — authorization code is regex-validated (\`/^[a-zA-Z0-9/_\\-.]+$/\`, max 512 chars) before reaching the token exchange.

### Why CodeQL still flags it

CodeQL's taint tracker follows \`searchParams.get('code')\` → branch condition → sensitive action. It cannot see through \`verifyOAuthState()\` to recognize it as a security sanitizer, and it cannot model that PKCE makes the code value worthless without the server-side verifier. This is a known limitation of static taint analysis for OAuth callbacks — every parameter in the callback URL is user-controllable by design.

The **CodeQL Security Analysis** check (3m30s, actual deep scan) **passes**. The failing **CodeQL** check (2-3s) is the default-branch comparison that flags new taint paths introduced by the PR. It is not a required check (no branch protection).

PR #875, opened 2026-04-10
feat(workflows): task-triggered workflows (due-date + completion)

2witstudios:
## Summary
- Add `task_due_date` and `task_completion` workflow trigger types so AI agents can run when a task's due date arrives or when it's completed
- Fix race condition in `fireCompletionTrigger` using `UPDATE...RETURNING` to atomically claim workflows and prevent double-execution
- Add internal try/catch to all fire-and-forget trigger helpers to prevent unhandled promise rejections
- Extract shared `createTaskTriggerWorkflow()` helper, eliminating 3 copy-pasted workflow creation blocks across REST route, AI tool create, and AI tool update paths
- Validate `contextPageIds` and `instructionPageId` at creation time (same-drive check), handle duplicate triggers via `onConflictDoUpdate`, add shape validation for `agentTrigger` in REST route
- Schema: `workflows` table gets `taskItemId` (FK cascade), `instructionPageId` (FK set null), unique constraint on `(taskItemId, triggerType)`
- Remove email PII from task context injected into workflow prompts
- Prevent instruction page double-injection in calendar trigger executor
- Add dueDate freshness check in cron to skip stale/postponed triggers

## Test plan
- [x] 20 tests for `task-trigger-helpers.ts` (sync, cancel, fire, disable, create — including race condition, validation, instruction page check)
- [x] 11 tests for `workflow-executor.ts` (task context injection, instruction page loading)
- [x] 64 tests for task PATCH/DELETE route (trigger cascades, status changes)
- [x] TypeScript typecheck clean (no errors)
- [x] ESLint clean (no warnings or errors)
- [x] Manual: create task with `agentTrigger` via REST API, verify workflow row created
- [ ] Manual: complete task, verify completion trigger fires and disables
- [ ] Manual: clear due date on task with due-date trigger, verify trigger disabled
- [ ] Manual: run migration `0097_familiar_boom_boom.sql` on staging DB

## Summary by CodeRabbit

* **New Features**
  * Tasks can now automatically trigger AI agents—either when a due date arrives or when the task is marked complete.
  * Configure custom instructions and reference context pages for task-triggered workflows.
  * Due-date triggered workflows automatically disable after execution to prevent duplicate runs.


2witstudios:
Re-checked after the latest changes. I still see the following issues on the current head:

1. Internal task-trigger workflows still leak into the generic workflows dashboard/API.
- `GET /api/workflows` still returns every row in `workflows` with no trigger-type filter (`apps/web/src/app/api/workflows/route.ts:55-59`), and the dashboard still consumes that endpoint directly (`apps/web/src/hooks/useWorkflows.ts:15`, `apps/web/src/components/workflows/WorkflowsDashboard.tsx:18-20`).
- The dashboard/types/rendering still only model cron/event workflows (`apps/web/src/components/workflows/types.ts:13`, `apps/web/src/components/workflows/WorkflowList.tsx:85-97`).
- Toggling or editing one of these leaked task-trigger rows still goes through `PATCH /api/workflows/[workflowId]`, which only recomputes `nextRunAt` for `cron` and otherwise writes `nextRunAt = null` (`apps/web/src/app/api/workflows/[workflowId]/route.ts:131-145`). A `task_due_date` trigger surfaced there can still be silently disabled by a normal admin toggle/edit.

2. Due-date reschedules are still not atomic with task edits.
- The task PATCH route updates the task inside the transaction, then calls `syncTaskDueDateTrigger()` only after commit (`apps/web/src/app/api/pages/[pageId]/tasks/[taskId]/route.ts:307-314`).
- A concurrent cron run can still claim the old `nextRunAt` in that window, and the cron pre-check still only validates `completedAt` (`apps/web/src/app/api/cron/workflows/route.ts:72-87`), not the task’s current `dueDate`.
- Postponing or clearing a due date right before it fires can still execute the agent at the stale deadline.

3. `update_task` can still miss a newly-added completion trigger.
- In the update path, `fireCompletionTrigger(taskId)` still runs before `createTaskTriggerWorkflow(...)` (`apps/web/src/lib/ai/tools/task-management-tools.ts:281-307`).
- When `update_task` both marks the task done and adds `agentTrigger.triggerType: 'completion'`, the completion callback finds no workflow, returns, and only then is the new `task_completion` row inserted. The automation stays `never_run` until the task is reopened and completed again.

4. New regression from the follow-up patch: failed task deletes can disable triggers without deleting the task.
- `DELETE /api/pages/[pageId]/tasks/[taskId]` now fire-and-forgets `disableTaskTriggers(taskId, 'Task deleted')` before the linked page trash mutation runs (`apps/web/src/app/api/pages/[pageId]/tasks/[taskId]/route.ts:453-455`).
- If `applyPageMutation(...trash...)` then returns a revision error or throws (`apps/web/src/app/api/pages/[pageId]/tasks/[taskId]/route.ts:499-529`), the delete fails but the task’s triggers have already been disabled.

2witstudios:
## Review feedback addressed in d8ce59c2

**P1 — Validate instruction page IDs** (task-trigger-helpers.ts:73): Added validation that `instructionPageId` exists, is in the same drive, and isn't trashed before inserting the workflow row. This now throws a clear client error instead of letting the FK violation produce a 500. Added a corresponding test (20/20 pass).

**P2 — Recheck isEnabled when claiming** (task-trigger-helpers.ts:192): Added `eq(workflows.isEnabled, true)` to the claim UPDATE's WHERE clause. If another request disables the trigger via `disableTaskTriggers` between the SELECT and UPDATE, the claim now returns 0 rows and we skip execution.

**P2 — Reject agentTrigger on non-drive task lists** (task-management-tools.ts:462): Both the create and update paths in the AI tool now throw/return an explicit error when `agentTrigger` is provided but the task list has no `driveId`. This matches the REST route behavior.

2witstudios:
## CodeRabbit review feedback addressed in a461e01d

**1. Stale dueDate check in cron** (route.ts:89): Added `dueDate` freshness check — cron now skips due-date triggers if the task's due date was cleared or postponed after the workflow was claimed.

**2. Trigger update ordering** (route.ts:314): The fire-and-forget design is intentional for response time, but acknowledged the race window concern. The cron pre-execution check (fix #1) now catches stale-fire scenarios.

**3. disableTaskTriggers ordering** (route.ts:454): Moved `disableTaskTriggers` to AFTER successful deletion in both the no-linked-page and linked-page DELETE paths.

**4. Trigger cascade ordering in AI tool** (task-management-tools.ts:284/309): Reordered so `createTaskTriggerWorkflow` runs BEFORE `fireCompletionTrigger`. This fixes the bug where setting `status: completed` + adding a completion trigger in the same call would fire before the workflow exists.

**5. Instruction page double injection** (calendar-trigger-executor.ts:71): Set `instructionPageId: null` on the synthetic workflow since `buildTriggerPrompt()` already loads instruction page content.

**6. Unhandled rejection in catch handler** (task-trigger-helpers.ts:235): Wrapped `db.update()` calls inside both the `.then()` and `.catch()` of the fire-and-forget chain with their own `try/catch`.

**7. Email PII in prompts** (workflow-executor.ts:331): Removed `userEmail` fallback from assignee context. Now uses names/titles only with `'Unknown assignee'` as fallback.

All local validations pass: typecheck clean, lint clean, 31/31 workflow tests pass.

2witstudios:
**Minor: triggerType metadata overwrite** (task-trigger-helpers.ts:113) — Fixed in 2a429f8c. Changed from single `triggerType` to `triggerTypes` array. Now supports both trigger types on the same task without overwriting.

PR #897, opened 2026-04-13
feat(security): audit mcp-ws route via unified audit() pipeline

2witstudios:
## Summary

Migrates the WebSocket MCP bridge (`/api/mcp-ws`) from legacy dual-write logging (`logSecurityEvent` + `securityAudit`) to the single `auditRequest()` function from `@pagespace/lib/server`, and tightens several semantic and privacy issues surfaced during review.

## Key Changes

- **Unified audit pipeline** — every security event in the route now flows through `auditRequest(request, event)`, which auto-extracts `ipAddress` (with `x-forwarded-for` → `x-real-ip` fallback) and `userAgent` from the request headers. Replaces hand-rolled header parsing.
- **`resourceType: 'mcp_websocket'`** stamped on every audit event so forensic queries can filter by resource type.
- **Privacy fix — fingerprint removed from audit details** — the connection fingerprint (SHA-256 of IP+UA) is a stable client-linkable pseudonym. The audit hash chain deliberately keeps `ipAddress`/`userAgent` out of `details` so they can be GDPR-anonymized; embedding the fingerprint there reintroduced linkable data into the immutable part of the chain. Truncated 16-char prefix wasn't useful for verification anyway.
- **Normal-close noise eliminated** — code 1000/1001 disconnects no longer audit (routine lifecycle, not security signals). Abnormal closes are audited as `security.anomaly.detected`.
- **Protocol noise dropped** — `ws_tool_execution_result` and `ws_fetch_response_start` no longer audit (high-volume protocol acks, not security events).
- **Risk score recalibration:**
  - DB error during session validation: 0.5 → 0.3 (infrastructure failure, not security risk)
  - Generic WebSocket error: 0.5 → 0.3 (network noise)
  - Fetch response error: 0.3 → 0.1 (commonly network/provider issues)
- **Pre-existing bug fix** — close handler was calling `cancelUserRequests()` twice on active connections, with the second call missing its `isFetchBridgeInitialized()` guard.
- **Dead code removal** — unused `logSecurityEvent` function in `apps/web/src/lib/websocket/ws-security.ts`.

## Test plan

- [x] `pnpm --filter web lint` — only pre-existing CalendarView warnings
- [x] `pnpm test:unit` — 20/20 mcp-ws security tests pass (added 4 new cases, removed 1 obsolete)
- [x] New test: normal close (1000 **and** 1001) does NOT audit (parameterized via `it.each`)
- [x] New test: abnormal close (1006) audits as `security.anomaly.detected`
- [x] New test: `cancelUserRequests` fires exactly once on close
- [x] New test: `auth.session.created` details does NOT contain a `fingerprint` key (locks in the privacy fix)
- [x] Strengthened test: "should not leak sensitive data" now asserts against `auditRequest` mock payloads in addition to `console.*`, so a regression piping a raw bearer token into `audit({ details })` would be caught.
- [ ] Manual smoke test: connect desktop app, verify WS upgrade succeeds and MCP tool execution works

## Review history

- Initial implementation in 0c212e8c
- Forensic enrichment (userAgent + resourceType + x-real-ip) in 3731b4f1
- CodeRabbit feedback addressed in c37bae74:
  - Switched `audit()` → `auditRequest()` per CodeRabbit suggestion
  - Removed fingerprint from audit details (privacy)
  - Strengthened secret-leak regression test
  - Parameterized 1000/1001 close-code test

### DIFF
--- a/apps/web/src/app/api/mcp-ws/__tests__/route.security.test.ts
+++ b/apps/web/src/app/api/mcp-ws/__tests__/route.security.test.ts
@@ -75,7 +75,7 @@ vi.mock('@pagespace/lib/server', () => ({
     security: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
     api: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
   },
-  audit: vi.fn(),
+  auditRequest: vi.fn(),
 }));
 
 import {
@@ -88,7 +88,7 @@ import {
   validateMessageSize,
 } from '@/lib/websocket';
 import { sessionService } from '@pagespace/lib';
-import { audit } from '@pagespace/lib/server';
+import { auditRequest } from '@pagespace/lib/server';
 
 // Mock session expiry (1 hour from now)
 const mockSessionExpiry = new Date(Date.now() + 60 * 60 * 1000);
@@ -188,12 +188,11 @@ describe('WebSocket MCP Bridge - Security Tests', () => {
       await UPGRADE(mockClient, mockServer, mockRequest);
 
       expect(mockClient.close).toHaveBeenCalledWith(1008, 'Insufficient permissions');
-      expect(audit).toHaveBeenCalledWith(
+      expect(auditRequest).toHaveBeenCalledWith(
+        mockRequest,
         expect.objectContaining({
           eventType: 'authz.access.denied',
           userId: 'user_123',
-          ipAddress: '192.168.1.1',
-          userAgent: 'Mozilla/5.0 Test',
           resourceType: 'mcp_websocket',
           riskScore: 0.5,
           details: expect.objectContaining({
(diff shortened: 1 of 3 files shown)


2witstudios:
Thanks @coderabbitai — addressed all four items in c37bae74 and the earlier 3731b4f1:

**route.ts**
1. ✅ Switched from `audit()` to `auditRequest(request, event)` throughout the handler. `auditRequest` now handles x-forwarded-for / x-real-ip / user-agent extraction via the shared helper, so the route no longer duplicates that logic.
2. ✅ Removed the truncated fingerprint from `auth.session.created` details. You're right that embedding a stable client-linkable pseudonym in the tamper-evident hash chain would resist GDPR erasure. The 16-char prefix wasn't useful for verification anyway. Added an inline comment explaining the privacy rationale so this doesn't come back.

**route.security.test.ts**
3. ✅ The "should not leak sensitive data" case now asserts against the serialized `auditRequest` mock payloads *in addition* to `console.*`. Regex tightened to also catch `bearer` and the literal `svc_mock_opaque_token` test fixture. Updated test name to reflect the broader coverage.
4. ✅ Parameterized the normal-close test over `[1000, 1001]` via `it.each` — the 1001 branch is now covered explicitly.

Also locked in the privacy fix with a new assertion that `auth.session.created` details does **not** contain a `fingerprint` key.

All 20 tests pass locally (`pnpm --filter web typecheck` / `lint` / vitest all green). Pushing for CI.

PR #1002, opened 2026-04-14
fix(sw): stop owning top-level navigations — restores desktop OAuth callback

2witstudios:
## Summary

Desktop Google OAuth was rendering `{"error":"offline","message":"You are offline"}` as a JSON page body after completing sign-in at Google. The service worker was intercepting the top-level callback navigation, attempting to `fetch()` the redirect, and returning a synthetic 503 JSON body because the server's response is a redirect to a `pagespace://` custom-scheme URL that `fetch()` cannot follow.

This is one instance of a broader class of bug: the SW assumed every `GET /api/*` is an XHR from the running SPA, when the browser can also perform a **top-level navigation** through those paths as part of an OAuth redirect chain. That's the browser's business, not the SW's.

### Observed (verified by code reading)
- The offline JSON exists in exactly one place: `apps/web/public/sw.js` inside the `catch` of `networkFirstWithCache`.
- `networkFirstWithCache` is selected for every same-origin GET whose pathname starts with `/api/`.
- The OAuth callback at `apps/web/src/app/api/auth/google/callback/route.ts:304` returns `NextResponse.redirect('pagespace://auth-exchange?code=...')` on the desktop branch. No code path returns JSON.

### Hypothesis (documented, not yet verified in a real desktop browser)
- When `fetch()` with `redirect: 'follow'` encounters a 3xx `Location: pagespace://...`, it throws instead of returning an opaque redirect. The mechanism-evidence test records what we observe in the Node test runtime; real-browser behavior is a follow-up.

### The invariant holds even if the mechanism hypothesis is wrong
The right fix is not \"catch the specific fetch error\" or \"bypass `/api/auth/` paths.\" The right fix is: **the service worker must not own top-level navigations**. `request.mode === 'navigate'` is the semantic boundary — this makes the fix robust regardless of whether fetch throws or returns opaque on custom-scheme redirects.

## Invariants (system rules, encoded in code and test names)

1. **The service worker must not own top-level navigations.** Requests with `request.mode === 'navigate'` are the browser's business. The SW must not call `event.respondWith` with a synthetic body that replaces the browser's native navigation result.
2. **Service workers must not sit in the credential-exchange redirect path.** Auth routes redirect to custom schemes; rule 1 keeps them browser-owned mechanically rather than via a path allowlist.
3. **\"API\" to the SW means XHR/fetch from the running SPA, not top-level navigation.** Encoded in the SW routing via request mode, not as an implicit assumption.

## The fix

- New `apps/web/public/sw-router.js` extracts routing to a pure `classifyRequest({ mode, method, url })` function that returns one of `not-intercepted | native-pass-through | api-network-first | cache-first | network-first`.
- `apps/web/public/sw.js` loads it via `importScripts('/sw-router.js')` and switches on the classifier result. `native-pass-through` **returns without calling `event.respondWith`** — operationally different from `event.respondWith(fetch(request))`, which would still break custom-scheme dispatch.
- Cache version bumped to `pagespace-v3` / `pagespace-static-v3` so every user gets the fixed SW on their next visit without needing to clear site data.
- A path-based allowlist (`/api/auth/`) was deliberately rejected: it would silently break any future non-auth navigation that happens to land on a path under `/api/`. Mode is the semantic boundary.

### Level 1 — classifier unit tests (`apps/web/src/__tests__/sw-router.test.ts`)
Seven tests covering all five classifications and the three invariants by name:
- Rule 1: navigation to `/api/auth/google/callback?code=x` → `native-pass-through`
- Rule 1 generalizes: navigation to `/dashboard` → `native-pass-through` (invariant is mode-based, not path-based)
- Rule 3: CORS-mode GET to `/api/pages/123` → `api-network-first`
- `/_next/` GET → `native-pass-through` (preserves existing behavior)
- Static asset GET → `cache-first`
- Other same-origin GET → `network-first`
- POST → `not-intercepted`

Tests name behaviors, not implementation details. They would still pass if `sw.js` internals were reorganized, as long as classifier output is stable.

### Level 2 — fetch redirect behavior evidence (`apps/web/src/__tests__/fetch-custom-scheme-redirect.test.ts`)
A `@mechanism-evidence` test that spins up a local `http.createServer` returning `302 Location: pagespace://auth-exchange?code=test` and asserts that `fetch()` does not produce a follow-through 2xx response. The assertion is intentionally loose — the important claim is \"fetch cannot materialize a normal 2xx we could hand back as a navigation result,\" not the specific failure mode.

### Level 3 — Playwright real-browser repro
**Not in this PR.** A Playwright test that registers the SW, visits a route that redirects to `pagespace://`, and asserts no synthetic offline JSON is rendered would be the only test that proves the interaction end-to-end. Filed as a follow-up.

## Failure modes after the fix

- **Stale SW still active.** Bumping `CACHE_NAME` forces a fresh activation, but users on an old SW will see the bug until their browser updates.
- **Branded `/offline` page no longer renders on top-level navigation.** Intentional: this is the operational consequence of rule 1. Users see the native browser offline UI instead for top-level page loads. Non-navigate HTML sub-requests still hit the offline-fallback branch.
- **Server-side OAuth callback could still be broken.** The SW fix assumes the server returns a well-formed `pagespace://auth-exchange` redirect. If the server is broken, the SW fix won't help. Verify with `curl -v` against the real callback after merge.
- **`fetch` redirect behavior differs from hypothesis.** The navigation-mode fix still holds; only the Level 2 evidence test needs updating to reflect observed behavior.

## Verification

- `pnpm --filter web exec vitest run src/__tests__/sw-router.test.ts` — 7/7 pass
- `pnpm --filter web exec vitest run src/__tests__/fetch-custom-scheme-redirect.test.ts` — 1/1 pass
- `pnpm --filter web lint` — no new warnings (three pre-existing `react-hooks/exhaustive-deps` in `CalendarView.tsx`)
- `pnpm --filter web typecheck` — no new errors; pre-existing `@pagespace/lib/*` subpath resolution errors on base are unchanged
- `pnpm --filter web build` — pre-existing webpack `@pagespace/lib/client-safe` module-not-found on base is unchanged

**Manual verification (post-merge, do not run from the agent):**
- [ ] Hard refresh a browser where PageSpace is loaded. Confirm the new SW version (`pagespace-v3`) shows up in DevTools → Application → Service Workers.
- [ ] Click Google sign-in on desktop. Confirm the OAuth callback redirect reaches `pagespace://auth-exchange` and Electron lands signed-in. No JSON blob rendered as page content.

## Scope

One file is the whole feature: `apps/web/public/sw.js` plus the new `apps/web/public/sw-router.js` plus the two new test files. **No changes** to `PasskeyLoginButton.tsx`, `passkey-service.ts`, any OAuth route handler, the Electron main process, or anything outside the service worker and its tests. This is independent of and does not block desktop passkey work.

## Test plan
- [x] Level 1 classifier unit tests pass
- [x] Level 2 fetch-custom-scheme-redirect evidence test passes
- [x] Lint and typecheck show no regressions vs base
- [ ] Manual: `pagespace-v3` shows up in DevTools after hard refresh
- [ ] Manual: desktop Google sign-in reaches `pagespace://auth-exchange`, no offline JSON body
- [ ] Follow-up: Playwright test for the real-browser interaction

## Summary by CodeRabbit

* **Chores**
  * Updated service worker cache versioning to optimize caching strategy.
  * Enhanced service worker request routing logic for improved performance.

### DIFF
--- a/apps/web/public/sw-router.js
+++ b/apps/web/public/sw-router.js
@@ -1,0 +1,89 @@
+/**
+ * PageSpace service worker routing classifier.
+ *
+ * Pure decision function extracted from sw.js so it can be unit-tested
+ * without faking FetchEvent. Consumed two ways:
+ *   - In the service worker via importScripts('/sw-router.js'), which
+ *     attaches classifyRequest to `self`.
+ *   - In vitest via `import { classifyRequest } from '../../public/sw-router.js'`,
+ *     which reads the CommonJS module.exports assignment below.
+ *
+ * The classifier encodes three layered invariants, in order of strength:
+ *   1. The service worker must not own top-level navigations. Requests
+ *      with `request.mode === 'navigate'` are the browser's business —
+ *      OAuth callbacks, redirects to custom schemes like pagespace://,
+ *      and normal page loads must remain browser-owned so native scheme
+ *      dispatch and redirect following work correctly.
+ *   2. Service workers must not sit in the credential-exchange redirect
+ *      path. Auth routes redirect to custom schemes; rule 1 keeps them
+ *      browser-owned mechanically rather than via a path allowlist.
+ *   3. "API" to the SW means XHR/fetch from the running SPA, not a
+ *      top-level navigation. Encoded via request mode, not path prefix.
+ */
+
+/**
+ * @typedef {(
+ *   | 'not-intercepted'
+ *   | 'native-pass-through'
+ *   | 'api-network-first'
+ *   | 'cache-first'
+ *   | 'network-first'
+ * )} RouteClassification
+ */
+
+/**
+ * @param {string} pathname
+ * @returns {boolean}
+ */
+function isStaticAssetPath(pathname) {
+  const staticExtensions = [
+    '.png',
+    '.jpg',
+    '.jpeg',
+    '.gif',
+    '.svg',
+    '.ico',
+    '.woff',
+    '.woff2',
+    '.ttf',
+    '.eot',
+  ];
+  return staticExtensions.some((ext) => pathname.endsWith(ext));
+}
+
+/**
+ * @param {{ mode: string, method: string, url: string }} input
+ * @returns {RouteClassification}
+ */
+function classifyRequest(input) {
+  const { mode, method, url } = input;
+
+  if (method !== 'GET') return 'not-intercepted';
+
+  // Rule 1 enforcement: top-level navigations bypass the SW entirely.
+  // A path-based allowlist would silently break future non-auth
+  // navigations under /api/; mode is the semantic boundary.
+  if (mode === 'navigate') return 'native-pass-through';
+
+  const parsed = new URL(url, 'http://localhost');
+
+  // Next.js runtime chunks: let the browser fetch directly to avoid
+  // stale bundle mismatches after deploys.
+  if (parsed.pathname.startsWith('/_next/')) return 'native-pass-through';
+
+  // Rule 3: any remaining /api/ request is an in-app XHR, safe for
+  // network-first-with-cache. Navigations were peeled off above.
+  if (parsed.pathname.startsWith('/api/')) return 'api-network-first';
+
+  if (isStaticAssetPath(parsed.pathname)) return 'cache-first';
+
+  return 'network-first';
+}
+
+if (typeof module !== 'undefined' && module.exports) {
+  module.exports = { classifyRequest, isStaticAssetPath };
+}
+if (typeof self !== 'undefined') {
+  self.classifyRequest = classifyRequest;
+  self.isStaticAssetPath = isStaticAssetPath;
+}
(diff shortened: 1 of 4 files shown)


2witstudios:
@coderabbitai Disregarding the `.css` nitpick. Rationale:

1. `find apps/web/public -name "*.css"` returns empty — there are no standalone CSS files in the served public tree (only PNGs, ICOs, the app icons, `manifest.json`, `robots.txt`, and the SW itself).
2. All Next.js-emitted CSS lives under `/_next/static/css/`, which is already classified `native-pass-through` one branch earlier (`sw-router.js:72`). Adding `.css` to `isStaticAssetPath` would never fire at runtime in this repo.
3. The comparison to `apps/web/src/middleware/monitoring.ts` isn't quite apples-to-apples: that middleware runs server-side before the `/_next/` check, so it has a different decision surface. The SW classifier peels off `/_next/` first, leaving no CSS requests to catch.

Adding `.css` would be dead code — the fix policy on this branch is to not introduce unreachable branches. Happy to revisit if a non-`/_next/` CSS asset is ever added to `/public`.

PR #1033, opened 2026-04-17
fix(security-headers): allow payment for Stripe iframe on billing pages

2witstudios:
## Summary

- Add `payment=(self "https://js.stripe.com")` to the `Permissions-Policy` response header so Stripe's `PaymentElement` (cross-origin iframe from `js.stripe.com`) can use the Payment Request API. Previously the directive was omitted, so browsers defaulted to `self` — which blocks the API inside cross-origin iframes and spammed the console with "payment is not allowed" violations on `/settings/plan` and `/settings/billing`.
- Refactor: extract a shared `PERMISSIONS_POLICY` constant at the top of `security-headers.ts` and reference it from both `ERROR_RESPONSE_HEADERS` and `applySecurityHeaders` so API error responses and page responses advertise the same value — no drift between the two.
- Tests: update the `applySecurityHeaders` assertion to match the new header; add a parallel assertion in the `createSecureErrorResponse` `'includes security headers'` test so future divergence is caught automatically.

Companion Caddyfile change in `PageSpace-Deploy` (https://github.com/2witstudios/PageSpace-Deploy/pull/2) is already merged — both were required for the production effect (Caddy's global `Permissions-Policy` header would otherwise override whatever the upstream app sends).

## Test plan

- [x] `pnpm --filter web typecheck` — no regressions
- [x] `pnpm --filter web test src/middleware/__tests__/security-headers.test.ts` — 49/49 tests pass (original + new `createSecureErrorResponse` assertion)
- [ ] Manual: `pnpm dev`, open `/settings/plan`, confirm the console no longer logs the "payment is not allowed" violation
- [ ] Manual: Stripe PaymentElement still renders and accepts card input on `/settings/plan` and `/settings/billing`
- [ ] Post-prod (Caddyfile change already merged): on a supported browser/device, wallet buttons (Apple Pay / Google Pay) appear inside Stripe's PaymentElement

### DIFF
--- a/apps/web/src/middleware/__tests__/security-headers.test.ts
+++ b/apps/web/src/middleware/__tests__/security-headers.test.ts
@@ -253,7 +253,7 @@ describe('Security Headers', () => {
       applySecurityHeaders(response, { nonce: 'test', isProduction: false });
 
       expect(response.headers.get('Permissions-Policy')).toBe(
-        'geolocation=(), microphone=(), camera=()'
+        'geolocation=(), microphone=(), camera=(), payment=(self "https://js.stripe.com")'
       );
     });
 
--- a/apps/web/src/middleware/__tests__/security-headers.test.ts
+++ b/apps/web/src/middleware/__tests__/security-headers.test.ts
@@ -401,6 +401,9 @@ describe('Security Headers', () => {
       expect(response.headers.get('Referrer-Policy')).toBe(
         'strict-origin-when-cross-origin'
       );
+      expect(response.headers.get('Permissions-Policy')).toBe(
+        'geolocation=(), microphone=(), camera=(), payment=(self "https://js.stripe.com")'
+      );
     });
 
     it('includes HSTS in production', () => {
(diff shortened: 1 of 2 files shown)


2witstudios:
@coderabbitai — both findings addressed in 3d715eb7:

**Outside-diff comment (security-headers.ts L6–12):** Extracted a shared `PERMISSIONS_POLICY` constant at the top of the file and now reference it from both `ERROR_RESPONSE_HEADERS` (line 12) and `applySecurityHeaders` (line 119). `createSecureErrorResponse` spreads `ERROR_RESPONSE_HEADERS`, so error responses and page responses now emit the identical `Permissions-Policy` value and cannot drift. The inline literal in `applySecurityHeaders` is gone.

**Nitpick (security-headers.test.ts L255–257):** Added a parallel assertion inside the `'includes security headers'` test for `createSecureErrorResponse` (around L404) that asserts `Permissions-Policy` equals the same string as the `applySecurityHeaders` test, so any future divergence is caught automatically. Full suite: 49/49 passing; `tsc --noEmit` clean.

PR #1068, opened 2026-04-22
chore(db): convert unrun one-time scripts to Drizzle migrations

2witstudios:
## Summary

- These scripts existed in `scripts/` but never ran in prod — Docker containers lack `tsx` and the source repo
- Converting to Drizzle migrations (0106–0108) ensures automatic execution on `pnpm db:migrate` at next deploy
- Each migration is idempotent and carries a comment header explaining what ran and why
- Note: the originally-planned `hash_token_backfill` was dropped — that work was already completed by migration 0048 which backfilled `tokenHash`/`tokenPrefix` and dropped the plaintext `token` column; running it again would reference a dropped table and dropped columns

## Migrations

| # | File | Source | What it does |
|---|------|--------|--------------|
| 0106 | `0106_reset_activity_log_chain.sql` | `scripts/rehash-activity-logs.ts` | Nulls forked/PII-tainted hash chain in `activity_logs` so next `logActivity()` seeds a fresh clean chain |
| 0107 | `0107_truncate_security_audit_log.sql` | *(decision, no prior script)* | Truncates `security_audit_log` — ~2019 forked-chain entries, `event_hash` NOT NULL prevents selective repair; almost no real user data, infra move imminent |
| 0108 | `0108_cleanup_ephemeral_task_lists.sql` | `scripts/cleanup-ephemeral-tasks.ts` | Deletes orphaned conversation-scoped task lists (replaced by page-based lists in Dec 2025) and their child `task_items` |

## Why only 3 migrations (not 4)?

The original `scripts/migrate-token-hashes.sql` targeted `refresh_tokens` (dropped in migration 0042) and a plaintext `token` column on `mcp_tokens`/`device_tokens`/`verification_tokens` (dropped in migration 0048). By the time this migration would run, those tables/columns no longer exist. Migration 0048 already performed the equivalent backfill (`tokenHash = token`). Dropping 0106 prevents a hard deployment failure.

## Test plan

- [ ] Verify 3 `.sql` files exist in `packages/db/drizzle/` (0106, 0107, 0108)
- [ ] Verify `_journal.json` has entries for idx 106–108 with correct timestamps and tags
- [ ] Run `pnpm db:migrate` against a staging/dev database and confirm zero errors
- [ ] Confirm `activity_logs.logHash` is NULL everywhere post-0106
- [ ] Confirm `security_audit_log` is empty post-0107
- [ ] Confirm no rows in `task_lists` with `"pageId" IS NULL AND "conversationId" IS NOT NULL` post-0108

[...]

PR #1069, opened 2026-04-22
fix(audit): add chain_seq to prevent hash chain forking under lock contention

2witstudios:
## Summary

- **Root cause**: \`timestamp = new Date()\` is assigned *before* \`pg_advisory_xact_lock\` is acquired, so writers queuing for the lock capture their timestamp while waiting. When commit order and timestamp order diverge, \`ORDER BY timestamp DESC LIMIT 1\` returns a stale predecessor and the chain forks.
- **Fix**: add \`chain_seq BIGSERIAL\` to \`security_audit_log\` and \`activity_logs\`. Predecessor SELECTs now order by \`chain_seq DESC\` (DB-assigned, monotone commit order) instead of \`timestamp DESC\`.
- **Verifiers + utilities**: all batch fetches and utility queries (\`quickIntegrityCheck\`, \`getHashChainStats\`, \`SecurityAuditService._doInitialize\`) now order by \`chainSeq\` instead of \`timestamp\` for consistency.
- **Indexes**: migration 0109 adds \`btree\` indexes on both \`chain_seq\` columns so the hot-path \`ORDER BY chain_seq DESC LIMIT 1\` predecessor lookup and verifier batch fetches avoid full table scans under the advisory lock.
- **Cleanup migrations** (0106–0108, included in this PR):
  - `0106`: Set \`logHash = NULL\` on all existing \`activity_logs\` rows — the old hash algorithm included PII (userId/actorEmail), making those hashes unverifiable after user deletion. Also clears a forked chain caused by the Apr 14–22 race condition.
  - `0107`: \`TRUNCATE security_audit_log\` — ~2019 entries with a forked hash chain; \`event_hash\`/\`previous_hash\` are NOT NULL so selective reset is impossible. Table held no real user data (early infra, minimal traffic); decision to truncate is documented for future engineers.
  - `0108`: Delete orphaned conversation-scoped task lists (\`pageId IS NULL AND conversationId IS NOT NULL\`) created by pre-Dec-2025 AI workflows, now replaced by page-based lists.
- **SIEM cursor reset** (in 0109): 0106 nulled \`logHash\` on existing rows; 0107 truncated \`security_audit_log\`. Both leave \`siem_delivery_cursors.lastDeliveredId\` pointing at rows whose anchor hash is now NULL or gone. \`runChainPreflight\` fails closed on a null anchor, permanently halting SIEM delivery. Migration 0109 nulls both cursor rows, restoring the fresh-cursor path.
- **Flaky test fix**: \`encryption-utils > provides authenticated encryption\` had ~0.4% flakiness — the tamper replaced the last 2 hex chars with \`'FF'\`, which is a no-op when those chars were already \`'FF'\`. Fixed to XOR the last byte with \`0xFF\` (always produces a different value).

## Files changed

| File | Change |
|------|--------|
| \`packages/db/src/schema/security-audit.ts\` | \`bigserial\` import + \`chainSeq\` column (\`chain_seq\`) + btree index |
| \`packages/db/src/schema/monitoring.ts\` | \`bigserial\` import + \`chainSeq\` column + btree index |
| \`packages/db/drizzle/0106_reset_activity_log_chain.sql\` | Null all \`logHash\`/\`previousLogHash\`/\`chainSeed\` on existing activity_logs rows |
| \`packages/db/drizzle/0107_truncate_security_audit_log.sql\` | Truncate forked security_audit_log |
| \`packages/db/drizzle/0108_cleanup_ephemeral_task_lists.sql\` | Delete orphaned conversation-scoped task lists |
| \`packages/db/drizzle/0109_add_chain_seq.sql\` | \`ADD COLUMN chain_seq bigserial NOT NULL\`, \`CREATE INDEX\` on both tables, reset both SIEM cursor rows |
| \`packages/lib/src/audit/security-audit.ts\` | \`ORDER BY timestamp DESC\` → \`ORDER BY chain_seq DESC\` (write + init paths) |
| \`packages/lib/src/monitoring/activity-logger.ts\` | \`desc(activityLogs.timestamp)\` → \`desc(activityLogs.chainSeq)\` |
| \`packages/lib/src/audit/security-audit-chain-verifier.ts\` | \`chainSeq ASC\` as sole primary sort (drops timestamp) |
| \`packages/lib/src/monitoring/hash-chain-verifier.ts\` | \`chainSeq ASC/DESC\` in all queries: main verifier, \`quickIntegrityCheck\`, \`getHashChainStats\` |
| \`packages/lib/src/__tests__/encryption-utils.test.ts\` | Fix flaky tamper test: XOR instead of replace-with-FF |

## Why no backfill is needed

Adding \`bigserial\` to a populated table assigns seq values in heap-rewrite order, not in chain order. However, this is not an issue because:

- Migration 0106 (included in this PR) sets \`logHash = NULL\` on all existing \`activity_logs\` rows — so the predecessor lookup (\`WHERE logHash IS NOT NULL ORDER BY chainSeq DESC\`) finds no existing rows and treats the next write as a fresh chain start.
- Migration 0107 (included in this PR) \`TRUNCATE\`s \`security_audit_log\` — the table is empty when the \`bigserial\` column is added, so it starts at 1 from a clean state.
- Both tables are append-only (no DELETEs/UPDATEs on individual rows), so heap order equals insertion order for the \`bigserial\` default fill.

No rows with non-null hashes exist at migration time, making backfill a no-op.

## Test plan

- [x] \`pnpm typecheck\` — no errors
- [x] \`pnpm test:unit\` — existing chain tests pass (CI green)
- [ ] \`pnpm db:migrate\` — applies cleanly, migration 0109 succeeds
- [ ] Manual: invoke \`POST /api/cron/verify-audit-chain\` and \`GET /api/admin/audit-logs/integrity\` — both return \`isValid: true\` after next writes
- [ ] SIEM: confirm delivery resumes after migration (no \`db_error\` halt in processor logs)
- [ ] Prod deploy: \`pnpm db:migrate\` on deploy server, restart web container — new rows get sequential \`chain_seq\` immediately


2witstudios:
Review findings:

- **[P1] Backfill chain sequences in existing chain order** (`packages/db/drizzle/0109_colorful_thundra.sql`)
  Adding `bigserial` to populated audit/activity tables assigns sequence values in heap rewrite order, not in the existing hash-chain/timestamp order. After the migration, `ORDER BY chainSeq DESC` can select a non-tail historical row, causing the next insert to record the wrong predecessor and making existing chains appear corrupt. Backfill the new columns from the current chain order, then set the sequence default.

- **[P1] Order hash-chain verification by chain sequence** (`packages/lib/src/monitoring/hash-chain-verifier.ts`)
  Writers now build the chain using `chainSeq`, but verification still walks entries primarily by timestamp. If timestamps are created before the advisory lock and lock acquisition happens in the opposite order, verification can compute hashes with the wrong predecessor and falsely report corruption. Order verification by `chainSeq` alone, and apply the same fix to the security-audit verifier.

- **[P2] Add indexes for the new chain sequence lookups** (`packages/db/src/schema/monitoring.ts`, migration)
  The hot-path tail lookup uses `ORDER BY chainSeq DESC LIMIT 1` on every activity-log insert. Without btree indexes on both `activity_log.chain_seq` and `security_audit_log.chain_seq`, Postgres may scan/sort growing audit tables while holding the advisory lock. Add matching schema and migration indexes.

packages/lib/src/audit/security-audit-chain-verifier.ts
```diff
@@ -146,7 +146,7 @@ export async function verifySecurityAuditChain(
 
       const entries = await db.query.securityAuditLog.findMany({
         where: conditions.length > 0 ? and(...conditions) : undefined,
-        orderBy: [asc(securityAuditLog.timestamp)],
+        orderBy: [asc(securityAuditLog.timestamp), asc(securityAuditLog.chainSeq)],
```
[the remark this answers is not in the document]
2witstudios: Fixed in c890119. Changed  to  (drop timestamp as primary). Since chainSeq is a bigserial serial it is unique — timestamp as a tiebreaker was unnecessary and would cause the exact false-break reported here.

packages/lib/src/monitoring/hash-chain-verifier.ts
```diff
@@ -205,7 +205,7 @@ export async function verifyHashChain(
       // Fetch batch of entries ordered by timestamp
       const entries = await db.query.activityLogs.findMany({
         where: conditions.length > 0 ? and(...conditions) : undefined,
-        orderBy: [asc(activityLogs.timestamp)],
+        orderBy: [asc(activityLogs.timestamp), asc(activityLogs.chainSeq)],
```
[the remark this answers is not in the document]
2witstudios: Fixed in c890119. Changed orderBy to asc(activityLogs.chainSeq) only, dropping timestamp as primary. chainSeq is unique (bigserial) so timestamp was an unnecessary tiebreaker that caused exactly the false-break scenario you described.

packages/lib/src/audit/security-audit.ts
```diff
@@ -169,7 +169,7 @@ export class SecurityAuditService {
       const lastRecord = await tx.execute(sql`
         SELECT event_hash
         FROM security_audit_log
-        ORDER BY timestamp DESC
+        ORDER BY chain_seq DESC
```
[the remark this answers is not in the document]
2witstudios: Fixed in c890119. Added idx_security_audit_chain_seq (btree) on security_audit_log.chain_seq and idx_activity_logs_chain_seq (btree) on activity_logs.chainSeq in the Drizzle schema files. Migration 0110_clear_robin_chapel.sql creates both indexes.

packages/db/drizzle/0109_colorful_thundra.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "activity_logs" ADD COLUMN "chainSeq" bigserial NOT NULL;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "chain_seq" bigserial NOT NULL;
```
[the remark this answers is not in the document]
2witstudios: Fixed in c890119. Added btree indexes on both columns in the Drizzle schema files (idx_security_audit_chain_seq on security_audit_log.chain_seq, idx_activity_logs_chain_seq on activity_logs.chainSeq). Migration 0110_clear_robin_chapel.sql creates both via CREATE INDEX IF NOT EXISTS.

packages/lib/src/monitoring/hash-chain-verifier.ts
2witstudios: **Critical: verifier ORDER BY is still wrong.**

Using `timestamp` as the primary sort here does not fix the verifier. If entry A has `timestamp=T1, chain_seq=2` and entry B has `timestamp=T2 (T2>T1), chain_seq=1` — the exact race this PR is fixing — the verifier visits A first (earlier timestamp), then B. But `A.previousHash` points to B (B committed first), so the chain-link check fails: `B.previousHash !== A.eventHash`.

Fix:
```ts
orderBy: [asc(activityLogs.chainSeq)]
```

`timestamp` is not needed — `chain_seq` is already monotone and reflects actual commit order.

packages/lib/src/audit/security-audit-chain-verifier.ts
2witstudios: **Critical: same verifier ORDER BY problem as `hash-chain-verifier.ts`.**

`timestamp` is still the primary sort. Entries with out-of-order timestamps vs. `chain_seq` will be visited in timestamp order, causing false chain-link failures on chains built in `chain_seq` order.

Fix:
```ts
orderBy: [asc(securityAuditLog.chainSeq)]
```

packages/db/src/schema/monitoring.ts
```diff
@@ -347,6 +348,11 @@ export const activityLogs = pgTable('activity_logs', {
   // Retention management
   isArchived: boolean('isArchived').default(false).notNull(),
 
+  // Monotonically-increasing commit-order key; predecessor SELECT uses this
+  // instead of timestamp to avoid forking when timestamps are pre-assigned
+  // before the advisory lock is acquired.
+  chainSeq: bigserial('chainSeq', { mode: 'number' }),
```
2witstudios: **Major: no index on `chainSeq`, and missing `.notNull()`.**

`BIGSERIAL` creates a sequence and a default — it does NOT create an index. Both the predecessor lookup (`ORDER BY chainSeq DESC LIMIT 1`) and verifier batch sort (`ORDER BY chainSeq ASC`) are now hot paths on a table that can be very large. Without an index both degrade to full-table-scan + sort (O(n)).

Add to the table definition:
```ts
chainSeqIdx: index('idx_activity_logs_chain_seq').on(table.chainSeq),
```
And mirror it in the migration SQL.

Also: the SQL migration adds `NOT NULL` but this definition omits `.notNull()`. Drizzle will infer `number | null` for `chainSeq` in TypeScript. Add `.notNull()` to match the DB constraint.

packages/db/src/schema/security-audit.ts
```diff
@@ -104,6 +105,11 @@ export const securityAuditLog = pgTable('security_audit_log', {
   // Timing
   timestamp: timestamp('timestamp', { mode: 'date' }).defaultNow().notNull(),
 
+  // Monotonically-increasing commit-order key; predecessor SELECT uses this
+  // instead of timestamp to avoid forking when timestamps are pre-assigned
+  // before the advisory lock is acquired.
+  chainSeq: bigserial('chain_seq', { mode: 'number' }),
```
2witstudios: **Major: no index on `chain_seq`, and missing `.notNull()`.**

Same issue as `monitoring.ts`. The predecessor lookup and verifier batch sort both need this index:

```ts
chainSeqIdx: index('idx_security_audit_chain_seq').on(table.chainSeq),
```

And `.notNull()` on the field definition.

packages/db/drizzle/0109_colorful_thundra.sql
2witstudios: **Major: migration breaks chain continuity for pre-existing rows.**

PostgreSQL fills existing rows with sequence values in heap-scan order — not commit order. The first new entry after migration runs `ORDER BY chain_seq DESC LIMIT 1` to find its predecessor, but the row with the highest `chain_seq` is whichever row was last in the heap, not the most recently committed entry. The chain forks at the migration boundary.

Safer two-step approach:
1. Add the column nullable with no default
2. Backfill via `UPDATE ... SET chain_seq = sub.rn FROM (SELECT id, ROW_NUMBER() OVER (ORDER BY timestamp ASC, id ASC) AS rn FROM ...) sub`
3. Add `NOT NULL` + the index

This preserves the order existing chains were built in and avoids a fork at the boundary.

packages/db/src/schema/monitoring.ts
```diff
@@ -347,6 +348,11 @@ export const activityLogs = pgTable('activity_logs', {
   // Retention management
   isArchived: boolean('isArchived').default(false).notNull(),
 
+  // Monotonically-increasing commit-order key; predecessor SELECT uses this
+  // instead of timestamp to avoid forking when timestamps are pre-assigned
+  // before the advisory lock is acquired.
```
2witstudios: **Minor: multi-line comment block violates style guide.**

CLAUDE.md says one short line max. Collapse to:
```ts
// BIGSERIAL for commit-order predecessor lookup — timestamps are pre-assigned before the advisory lock.
```

2witstudios (commented):
The insert-path fix is correct — `ORDER BY chain_seq DESC` in both predecessor lookups is the right call. Three issues need to be resolved before merge: the verifier ORDER BY is still wrong (critical), both tables are missing an index on `chain_seq` (major perf regression), and the migration does not backfill existing rows so the chain forks at the migration boundary (major correctness). Details inline.

packages/lib/src/audit/security-audit-chain-verifier.ts
```diff
@@ -146,7 +146,7 @@ export async function verifySecurityAuditChain(
...
         offset,
         limit: currentBatchSize,
```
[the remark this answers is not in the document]
2witstudios: Fixed in c890119. Changed orderBy to [asc(securityAuditLog.chainSeq)] only, as proposed. chainSeq is unique (bigserial) so there are no ties — timestamp as a tiebreaker was unnecessary and causes false chain-break reports under the exact lock-contention scenario this PR fixes.

packages/lib/src/monitoring/hash-chain-verifier.ts
```diff
@@ -205,7 +205,7 @@ export async function verifyHashChain(
...
         offset,
         limit: currentBatchSize,
```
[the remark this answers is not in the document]
2witstudios: Fixed in c890119. Changed orderBy to [asc(activityLogs.chainSeq)] only, as proposed. chainSeq (bigserial) is unique so timestamp as a tiebreaker was unnecessary and caused the false-failure scenario you described.

packages/lib/src/monitoring/hash-chain-verifier.ts
2witstudios: Already fixed in c890119: orderBy changed to [asc(activityLogs.chainSeq)] only.

packages/lib/src/audit/security-audit-chain-verifier.ts
2witstudios: Already fixed in c890119: orderBy changed to [asc(securityAuditLog.chainSeq)] only.

packages/db/src/schema/monitoring.ts
2witstudios: Fixed in 688b12bd: added .notNull() to chainSeq (bigserial implies NOT NULL in SQL, but the TS type was incorrectly nullable without it). Index was added in c890119 (idx_activity_logs_chain_seq btree on chainSeq, migration 0110). Comment collapsed to one line.

packages/db/src/schema/security-audit.ts
2witstudios: Fixed in 688b12bd: added .notNull() to chain_seq column (same issue — bigserial NOT NULL in SQL, but TS type needed explicit .notNull()). Index was added in c890119 (idx_security_audit_chain_seq btree on chain_seq, migration 0110). Comment collapsed to one line.

packages/db/src/schema/monitoring.ts
2witstudios: Fixed in 688b12bd: collapsed to a single line in both schema files.

packages/db/drizzle/0109_colorful_thundra.sql
2witstudios: Good catch on the theoretical risk. For these specific tables the approach is safe:

**security_audit_log** and **activity_logs** are append-only — logEvent() / logActivity() only INSERT, never UPDATE or DELETE individual rows. For append-only tables, PostgreSQL heap order equals insertion order: each new row is appended to the end of the last heap page. When the migration does `ALTER TABLE ... ADD COLUMN chainSeq bigserial NOT NULL`, PostgreSQL's heap scan assigns chain_seq = 1, 2, ..., N in insertion order, so the row with chain_seq = N IS the most recently inserted row. The first new write after migration correctly chains off it.

If the tables had UPDATEs (which move rows in the heap), this reasoning would break and your backfill approach would be necessary. But since both tables are strict-insert-only, heap order == timestamp order and the bigserial assignment is equivalent to your ROW_NUMBER() OVER (ORDER BY timestamp ASC) backfill.

Happy to add a code comment to the migration if you want to make this assumption explicit.

2witstudios:
## Follow-up commit: chainSeq ordering in all utility queries (9cadf3da)

**On the backfill concern (P1):**

The backfill for existing rows is not needed in this specific deployment context. PR #1068 (merged immediately before this branch) ran two cleanup migrations:
- `0106` set `logHash = NULL` on all existing `activity_logs` rows — so the predecessor lookup (`WHERE logHash IS NOT NULL ORDER BY chainSeq DESC`) finds zero existing rows and treats the next write as a fresh chain start (new seed, no predecessor).
- `0107` ran `TRUNCATE security_audit_log` — the table is empty when migration 0109 runs, so bigserial starts at 1 from a clean state.

The heap-rewrite ordering concern is real in general, but here there are no hashed rows to corrupt at migration time.

**On remaining utility query ordering (now fixed in 9cadf3da):**

The secondary utility functions (`quickIntegrityCheck`, `getHashChainStats`) and `SecurityAuditService._doInitialize()` were still ordering by `timestamp` instead of `chainSeq`. Fixed:
- `quickIntegrityCheck()`: both `findFirst` and `findMany` now use `asc(chainSeq)`
- `getHashChainStats()`: first-entry lookup uses `asc(chainSeq)`, last-entry lookup uses `desc(chainSeq)` (also fixes a pre-existing bug where it used `asc` and returned the oldest entry instead of the newest)
- `SecurityAuditService._doInitialize()`: uses `[desc(securityAuditLog.chainSeq)]`

PR #1071, opened 2026-04-22
fix(db): reset SIEM delivery cursors after chain reset / audit log truncate (0109)

2witstudios:
## Problem

Migrations 0106 and 0107 (merged in #1068) left the SIEM delivery system permanently halted:

**0106** (`reset_activity_log_chain`): Nulled `logHash` on `activity_logs` rows but left `siem_delivery_cursors.lastDeliveredId` pointing at those same rows. On the next poll, `runChainPreflight` calls `loadAnchorHash` for the cursor anchor — finds the row, but `logHash` is now `NULL`. `loadAnchorHash` returns `null`. Preflight returns `{ kind: 'db_error' }` and halts delivery permanently.

**0107** (`truncate_security_audit_log`): Deleted all rows via TRUNCATE but left the `security_audit_log` cursor unchanged. `loadAnchorHash` can't find the anchor row (it was truncated) and returns `null`. Same permanent halt.

The fail-closed behavior in `runChainPreflight` is intentional (anchor loss is indistinguishable from tampering), but these are deliberate operational resets — not tampering — so the cursors need to be reset alongside the chain data.

## Fix

Migration 0109 nulls `lastDeliveredId` and `lastDeliveredAt` for both affected cursor rows. A null `lastDeliveredId` is the "fresh cursor" sentinel: `runChainPreflight` sees it and skips verification for that source, allowing delivery to resume cleanly from the next undelivered event.

```sql
UPDATE siem_delivery_cursors
SET "lastDeliveredId" = NULL,
    "lastDeliveredAt" = NULL
WHERE id IN ('activity_logs', 'security_audit_log');
```

Idempotent — no-op if the cursor rows don't exist yet.

## Test plan

- [ ] Apply 0109 on a DB that has run 0106+0107 — confirm SIEM delivery poll completes without `db_error` halt
- [ ] Confirm `runChainPreflight` takes the `lastDeliveredId === null → continue` branch for both sources on the first post-migration poll
- [ ] Verify idempotent: running 0109 twice is safe

## Summary by CodeRabbit

* **Chores**
  * Added database migrations to reset internal state consistency and clean up orphaned data structures.


2witstudios:
Merging into #1069 instead — that PR hasn't landed yet, so the cursor reset belongs there.

PR #1097, opened 2026-04-23
fix(cron): drop userId 'system' from audit calls

2witstudios:
## Summary

- All 10 cron jobs were passing \`userId: 'system'\` to \`audit()\`, causing PostgreSQL FK violation 23503 on every cron execution
- \`security_audit_log.user_id\` is a nullable FK to \`users.id\` — any non-NULL value must exist in the \`users\` table, but no \`'system'\` user exists
- Fix: remove \`userId: 'system'\` from all 12 audit call sites so system-actor events write \`NULL\`, which the schema already supports via \`onDelete: 'set null'\`
- Updated all matching test assertions to reflect the corrected behavior

Additional fixes discovered and addressed on this branch:

- **HMAC signature mismatch**: \`cron-curl\` was including the query string in the signed path, but \`validateSignedCronRequest\` validates against \`url.pathname\` (query-stripped). Fixed: strip query string before computing path for HMAC signature
- **Dead in-memory state**: \`SecurityAuditService.lastHash\` was an in-memory field from before \`pg_advisory_xact_lock\` was added. It was a latent multi-instance bug (each pod restart would fork the chain from \`'genesis'\`). Removed — the advisory lock already reads the previous hash from DB inside the transaction
- **Dead DB query**: After removing \`lastHash\`, the \`findFirst\` call in \`_doInitialize()\` had no consumer. Removed the round-trip; \`logEvent()\` surfaces DB errors on first write
- **Test alignment**: Updated 3 \`initialize()\` tests that were asserting on the now-removed \`findFirst\` implementation detail

## Test plan

- [ ] Deploy and watch logs — \`[Audit] audit write failed\` errors should stop within 2 minutes (calendar-triggers runs every 2 min)
- [ ] Verify in DB: \`SELECT user_id, COUNT(*) FROM security_audit_log WHERE resource_type = 'cron_job' GROUP BY user_id\` should show \`null\` rows, not \`'system'\`
- [ ] All cron route unit tests pass
- [ ] \`security-audit.test.ts\` passes (initialize tests updated to reflect no-DB-on-init behavior)

### DIFF
--- a/apps/web/src/app/api/cron/calendar-sync/__tests__/route.test.ts
+++ b/apps/web/src/app/api/cron/calendar-sync/__tests__/route.test.ts
@@ -72,7 +72,7 @@ describe('/api/cron/calendar-sync', () => {
     await GET(makeRequest());
 
     expect(mockAudit).toHaveBeenCalledWith(
-      expect.objectContaining({ eventType: 'data.write', userId: 'system', resourceType: 'cron_job', resourceId: 'calendar_sync', details: { synced: 0, failed: 0 } })
+      expect.objectContaining({ eventType: 'data.write', resourceType: 'cron_job', resourceId: 'calendar_sync', details: { synced: 0, failed: 0 } })
     );
   });
 
@@ -88,7 +88,7 @@ describe('/api/cron/calendar-sync', () => {
     await GET(makeRequest());
 
     expect(mockAudit).toHaveBeenCalledWith(
-      expect.objectContaining({ eventType: 'data.write', userId: 'system', resourceType: 'cron_job', resourceId: 'calendar_sync', details: { synced: 1, failed: 1 } })
+      expect.objectContaining({ eventType: 'data.write', resourceType: 'cron_job', resourceId: 'calendar_sync', details: { synced: 1, failed: 1 } })
     );
   });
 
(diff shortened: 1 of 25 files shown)


2witstudios (commented):
Thanks for the thorough review! Addressed in commit c25d15dfb:

**initialize() simplification**: Collapsed to a simple async no-op — removed `initializePromise` concurrency guard and the private `_doInitialize()` helper. The async surface is retained for backward compat with app-startup callers; `logEvent()` already self-initializes.

**Dead findFirst scaffolding**: Removed `findFirst: vi.fn()` from the mock factory and both `vi.mocked(...).mockResolvedValue(undefined)` no-ops from the `logEvent` and `convenience methods` `beforeEach` blocks.

**userId regression guards**: Added `expect(mockAudit).not.toHaveBeenCalledWith(expect.objectContaining({ userId: expect.anything() }))` to both workflow audit assertions, and `expect(mockAudit.mock.calls[0]?.[0]).not.toHaveProperty('userId')` to both verify-audit-chain assertions.

**serviceId**: The `resourceType: 'cron_job'` + `resourceId` fields already identify the system actor and the specific job. Adding `serviceId: 'cron'` uniformly across all 10 routes is a reasonable follow-up but would be a separate commit touching every cron route and test — deferring to keep this PR focused on the FK fix.

PR #1099, opened 2026-04-23
feat(voice): rebuild as floating call panel

2witstudios:
## Summary

Voice mode was built as a replacement interface (VoiceModeDock swaps out ChatInput entirely) with a hands-free loop that only worked in barge-in mode but defaulted to tap-to-speak. Multiple silent failure points in the cross-component state chain could kill the loop with no recovery.

Rebuilt as a thin STT+TTS layer on top of existing conversations — same history, same agent, same send path, just voice in/out.

- **VoiceCallPanel** replaces VoiceModeDock: compact strip that floats *above* ChatInput, never replacing it
- **barge-in default**: hands-free loop now runs end-to-end out of the box
- **State fixes**: `'waiting'` after send (not `'idle'`), `'speaking'` before TTS fetch (not after 1-3s)
- **Auto-listen fix**: reads `useVoiceModeStore.getState()` directly in the `onended` callback — breaks stale closure that could silently kill the loop
- **VAD in both modes**: silence auto-stop now works in tap-to-speak too
- **Voice button always visible**: removed `isVoiceModeAvailable` gate and per-view `checkOpenAI` fetch
- **Deleted**: `VoiceModeDock.tsx`, its test, and all `showVoiceSettings`/`isOpenAIConfigured` state across three views

## Test plan

- [ ] Activate voice mode in AiChatView, GlobalAssistantView, SidebarChatTab — ChatInput remains visible beneath the panel
- [ ] Speak → silence → auto-transcribes → auto-sends → AI responds → TTS plays automatically → auto-listens again (full barge-in loop)
- [ ] Speak during TTS → barge-in stops playback → STT starts
- [ ] Close panel → text chat fully functional, voice messages in history
- [ ] Voice button visible regardless of which AI provider is configured

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


2witstudios:
**Re: Codex P1 — Cancel in-flight synthesis when user barge-ins early**

Fixed in 848a18827. Added a guard immediately after `decodeAudioData` that reads `currentAudioId` from the live store — if it no longer matches the `audioId` generated at the start of this `speak()` call (meaning a barge-in cleared it during the fetch), the function returns early and discards the decoded audio without starting playback. No audio will play over a new recording.

PR #1119, opened 2026-04-24
fix(barrels): delete zombie barrel files, fix script imports, add lib ESLint

2witstudios:
Deletes packages/db/src/index.ts and packages/lib/src/index.ts (zero consumers remain after PRs #1112–1115). Removes their \".\" root exports from package.json. Fixes 3 scripts that still imported from bare @pagespace/lib. Adds packages/lib/eslint.config.mjs to prevent regression. Zero barrel imports now enforced structurally, not just by convention.

Also fixes two additional missed barrel consumers found during typecheck:
- apps/web/src/instrumentation.ts (two dynamic imports from bare @pagespace/lib)
- apps/web/src/lib/auth/index.ts (dynamic import from bare @pagespace/db)
- packages/db/src/migrate.ts (internal import from barrel)
- packages/db/src/__tests__/accessible-page-ids.integration.test.ts (internal import from barrel)
- packages/db/scripts/calculate-initial-storage.ts (internal import from barrel)


2witstudios:
Fixed the Unit Tests CI failure: the test files `calculate-initial-storage.test.ts` and `migrate.test.ts` were mocking `'../../src/index'` and `'../index'` respectively — the barrel files we just deleted. Updated both tests to mock the individual subpath modules (`../db`, `../schema/auth`, `../schema/core`, `../operators`) that the scripts now import directly. Both tests pass locally.

PR #1120, opened 2026-04-25
fix(billing): disable wallet methods in PaymentElement

2witstudios:
## Summary

- `PaymentElement` with `layout: 'tabs'` was dynamically showing Google Pay and Apple Pay options when the browser supported them
- Clicking "Add Card" triggered `elements.submit()` which opened the Google Pay native payment sheet instead of validating a card form
- The sheet left `elements.submit()` unresolved, keeping `processing: true` indefinitely (infinite spinner)
- Fix: set `wallets: { googlePay: 'never', applePay: 'never' }` so only card input is shown, matching the dialog's intent

## Test plan

- [ ] Open `/settings/billing` in Chrome (Google Pay supported)
- [ ] Click "Add Payment Method" — confirm no Google Pay / Apple Pay tab appears
- [ ] Enter Stripe test card `4242 4242 4242 4242`, click "Add Card"
- [ ] Spinner resolves, dialog closes, toast shows "Card added successfully"
- [ ] Confirm Google Pay iframe request absent from DevTools Network tab

## Release Notes

* **Payment Methods**
  * Google Pay and Apple Pay have been disabled as payment options during checkout.

### DIFF
--- a/apps/web/src/components/billing/AddPaymentMethodForm.tsx
+++ b/apps/web/src/components/billing/AddPaymentMethodForm.tsx
@@ -78,6 +78,10 @@ function AddPaymentMethodFormContent({
       <PaymentElement
         options={{
           layout: 'tabs',
+          wallets: {
+            googlePay: 'never',
+            applePay: 'never',
+          },
         }}
       />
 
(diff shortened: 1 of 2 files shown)


2witstudios:
Fixed in b0676f8. Applied the same `wallets: { googlePay: 'never', applePay: 'never' }` to `EmbeddedCheckoutForm`'s `PaymentElement` — same underlying risk, now consistent across both billing surfaces.

PR #1128, opened 2026-04-25
fix(billing): fix PaymentElement not rendering (infinite spinner)

2witstudios:
## Root cause

The `PaymentElement` was mounting but rendering **empty** — no card form visible. When the user clicked "Add Card", `elements.submit()` was called on an uninitialized element and hung indefinitely, leaving `processing: true` forever (the infinite "Adding..." spinner).

### Why the element rendered empty

`SetupIntent` was created with `payment_method_types: ['card']` — the legacy Stripe API parameter. The Stripe API version in use is `2025-12-15.clover`, where `PaymentElement` requires either `automatic_payment_methods` or `payment_method_configuration`. Using the legacy `payment_method_types` with the new API causes the element to mount but show no content.

The previous PR (#1120) disabled wallet options in the `PaymentElement`, but that doesn't help when the element itself never renders.

## Fix

**`apps/web/src/app/api/stripe/setup-intent/route.ts`**  
Switch from `payment_method_types: ['card']` to `automatic_payment_methods: { enabled: true, allow_redirects: 'never' }`.  
`allow_redirects: 'never'` restricts to on-page payment methods (effectively cards), matching the original intent without using the deprecated parameter.

**`apps/web/src/components/billing/AddPaymentMethodForm.tsx`**  
Track `onReady` from `PaymentElement` and disable the submit button until the element has fully loaded. This prevents `elements.submit()` from being called on an uninitialized element even if there's a delay.

## Test plan

- [ ] Open `/settings/billing`, click "Add Payment Method"
- [ ] Confirm card input form renders (not empty)  
- [ ] Confirm "Add Card" button is disabled until the form loads
- [ ] Enter test card `4242 4242 4242 4242`, submit — spinner resolves, card saved
- [ ] Confirm no Google Pay iframe in DevTools Network tab

## Summary by CodeRabbit

* **New Features**
  * Payment setup now enables Stripe’s automatic payment methods and prevents off-site redirects, letting the checkout surface compatible options dynamically.

* **Bug Fixes**
  * Payment form submission is disabled until the embedded payment element is fully initialized, reducing submission errors and improving reliability.


2witstudios:
Addressed in 22921ba:
- **Tests**: Updated both `setupIntents.create` assertions to use `automatic_payment_methods: { enabled: true, allow_redirects: 'never' }` (was `payment_method_types: ['card']`)
- **Comment**: Tightened wording to clarify that `allow_redirects:'never'` restricts redirect-based flows (not just cards), and that the displayed method set is controlled by the Stripe Dashboard + client-side `wallets` options

2witstudios:
Addressed the `onLoadError` suggestion from initial review in 4c76436: added handler on `PaymentElement` that sets the error state and surfaces a user-facing message if Stripe.js fails to load (e.g. network failure or blocked scripts). This prevents the submit button from remaining permanently disabled without any feedback.

2witstudios:
Superseded by a more comprehensive fix: replacing the custom PaymentElement form with the Stripe Customer Portal. The Customer Portal handles all payment method management (add/remove/set-default) server-side at Stripe, eliminating the need for a custom embedded form entirely.

PR #1140, opened 2026-04-28
feat(ai): StreamMulticastRegistry for multiplayer chat streaming

2witstudios:
## Summary

Implements \`StreamMulticastRegistry\` — a pure in-process pub/sub registry at \`apps/web/src/lib/ai/core/stream-multicast-registry.ts\`. This is Task 1 of 5 in the Multiplayer AI Chat Streaming epic.

**What it does:**
- Buffers all text chunks per stream and multicasts to every active subscriber
- Late-joining subscribers receive a full replay of the buffer before live chunks
- \`finish(messageId, aborted?)\` notifies all subscribers via \`onComplete\` and clears the entry
- 10-minute auto-cleanup timer fires \`finish(…, true)\` to prevent memory leaks from orphaned streams
- Exports \`StreamMeta\` interface, \`StreamMulticastRegistry\` class (for per-test isolation), and \`streamMulticastRegistry\` singleton

**Public API:**
\`\`\`ts
streamMulticastRegistry.register(messageId, { pageId, userId })
streamMulticastRegistry.push(messageId, text)
streamMulticastRegistry.subscribe(messageId, onChunk, onComplete) // → unsubscribe fn | null
streamMulticastRegistry.finish(messageId, aborted?)
streamMulticastRegistry.getMeta(messageId)  // → StreamMeta | undefined
\`\`\`

## Post-review fixes (commit e48d02245)

All review comments addressed:

| Issue | Fix |
|-------|-----|
| Duplicate `register` leaks orphaned timer | Clear existing entry's `cleanupTimeoutId` before overwriting |
| `onComplete` exception leaks entry | Delete entry from map **before** subscriber loop; wrap each callback in try/catch |
| `onChunk` exception breaks fanout | Wrap each `subscriber.onChunk()` in try/catch in `push()` |
| `StreamMeta` not exported | Added `export` to the interface |
| Redundant inline comment in test | Removed |
| Non-null `!` assertions in tests | Replaced with explicit `expect(fn).not.toBeNull()` assertions |
| No test for duplicate `register` | Added to `auto-cleanup` block |
| No resilience tests | Added `resilience` block: onChunk-throws and onComplete-throws tests |

## Test plan

- [x] 21 unit tests pass (3 new tests added post-review)
- [x] TypeScript: no errors in new files (`tsc --noEmit`)
- [x] CI: Unit Tests ✅, Lint & TypeScript Check ✅, CodeRabbit ✅
- [x] Mergeable: CLEAN, no conflicts with master
- [x] All 20 review threads replied to and resolved

### DIFF
--- a/apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
+++ b/apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
@@ -108,7 +108,8 @@ describe('StreamMulticastRegistry', () => {
       registry.subscribe('msg-1', (text) => received2.push(text), () => {});
 
       registry.push('msg-1', 'chunk1');
-      unsubscribe1!();
+      expect(unsubscribe1).not.toBeNull();
+      (unsubscribe1 as () => void)();
       registry.push('msg-1', 'chunk2');
 
       expect(received1).toEqual(['chunk1']);
@@ -125,7 +126,8 @@ describe('StreamMulticastRegistry', () => {
       const unsubscribe1 = registry.subscribe('msg-1', () => {}, (aborted) => completed1.push(aborted));
       registry.subscribe('msg-1', () => {}, (aborted) => completed2.push(aborted));
 
-      unsubscribe1!();
+      expect(unsubscribe1).not.toBeNull();
+      (unsubscribe1 as () => void)();
       registry.finish('msg-1');
 
       expect(completed1).toEqual([]);
@@ -178,7 +180,7 @@ describe('StreamMulticastRegistry', () => {
       registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));
 
       registry.finish('msg-1');
-      registry.finish('msg-1'); // second call should be a no-op
+      registry.finish('msg-1');
 
       expect(completed).toEqual([false]);
     });
@@ -213,8 +215,48 @@ describe('StreamMulticastRegistry', () => {
 
       vi.advanceTimersByTime(10 * 60 * 1000);
 
-      // Only one completion, not two
       expect(completedValues).toEqual([false]);
+    });
+
+    it('given register is called twice for the same messageId, should call onComplete only once after 10 minutes', () => {
+      vi.useFakeTimers();
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const completed: boolean[] = [];
+      registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));
+
+      vi.advanceTimersByTime(10 * 60 * 1000);
+
+      expect(completed).toHaveLength(1);
+    });
+  });
+
+  describe('resilience', () => {
+    it('given a subscriber whose onChunk throws, should not interrupt fanout to remaining subscribers', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const received: string[] = [];
+      registry.subscribe('msg-1', () => { throw new Error('bad subscriber'); }, () => {});
+      registry.subscribe('msg-1', (text) => received.push(text), () => {});
+
+      expect(() => registry.push('msg-1', 'chunk1')).not.toThrow();
+      expect(received).toEqual(['chunk1']);
+    });
+
+    it('given a subscriber whose onComplete throws, should still remove the entry and notify remaining subscribers', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const completed: boolean[] = [];
+      registry.subscribe('msg-1', () => {}, () => { throw new Error('bad subscriber'); });
+      registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));
+
+      expect(() => registry.finish('msg-1')).not.toThrow();
+      expect(registry.getMeta('msg-1')).toBeUndefined();
+      expect(completed).toEqual([false]);
     });
   });
 });
--- a/apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
+++ b/apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
@@ -1,0 +1,220 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { StreamMulticastRegistry } from '../stream-multicast-registry';
+
+describe('StreamMulticastRegistry', () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  describe('register / getMeta', () => {
+    it('stores stream metadata accessible via getMeta', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      expect(registry.getMeta('msg-1')).toEqual({ pageId: 'page-1', userId: 'user-1' });
+    });
+
+    it('returns undefined for unknown messageId', () => {
+      const registry = new StreamMulticastRegistry();
+      expect(registry.getMeta('unknown')).toBeUndefined();
+    });
+  });
+
+  describe('push', () => {
+    it('given a registered stream with subscribers, should push text chunks to all active subscribers', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const received1: string[] = [];
+      const received2: string[] = [];
+      registry.subscribe('msg-1', (text) => received1.push(text), () => {});
+      registry.subscribe('msg-1', (text) => received2.push(text), () => {});
+
+      registry.push('msg-1', 'hello');
+      registry.push('msg-1', ' world');
+
+      expect(received1).toEqual(['hello', ' world']);
+      expect(received2).toEqual(['hello', ' world']);
+    });
+
+    it('silently ignores push for unknown messageId', () => {
+      const registry = new StreamMulticastRegistry();
+      expect(() => registry.push('unknown', 'chunk')).not.toThrow();
+    });
+  });
+
+  describe('subscribe', () => {
+    it('given a late-joining subscriber, should replay all buffered chunks before delivering live ones', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      registry.push('msg-1', 'chunk1');
+      registry.push('msg-1', 'chunk2');
+
+      const received: string[] = [];
+      registry.subscribe('msg-1', (text) => received.push(text), () => {});
+
+      expect(received).toEqual(['chunk1', 'chunk2']);
+    });
+
+    it('given buffered and live chunks, should deliver them in order to a late subscriber', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      registry.push('msg-1', 'chunk1');
+      registry.push('msg-1', 'chunk2');
+
+      const received: string[] = [];
+      registry.subscribe('msg-1', (text) => received.push(text), () => {});
+
+      registry.push('msg-1', 'chunk3');
+
+      expect(received).toEqual(['chunk1', 'chunk2', 'chunk3']);
+    });
+
+    it('given a finished stream, should return null', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+      registry.finish('msg-1');
+
+      const result = registry.subscribe('msg-1', () => {}, () => {});
+      expect(result).toBeNull();
+    });
+
+    it('given an unknown messageId, should return null', () => {
+      const registry = new StreamMulticastRegistry();
+      const result = registry.subscribe('unknown', () => {}, () => {});
+      expect(result).toBeNull();
+    });
+
+    it('returns an unsubscribe function for an active stream', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const unsubscribe = registry.subscribe('msg-1', () => {}, () => {});
+      expect(typeof unsubscribe).toBe('function');
+    });
+  });
+
+  describe('unsubscribe', () => {
+    it('given a subscriber that unsubscribes, should stop receiving chunks without affecting other subscribers', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const received1: string[] = [];
+      const received2: string[] = [];
+
+      const unsubscribe1 = registry.subscribe('msg-1', (text) => received1.push(text), () => {});
+      registry.subscribe('msg-1', (text) => received2.push(text), () => {});
+
+      registry.push('msg-1', 'chunk1');
+      unsubscribe1!();
+      registry.push('msg-1', 'chunk2');
+
+      expect(received1).toEqual(['chunk1']);
+      expect(received2).toEqual(['chunk1', 'chunk2']);
+    });
+
+    it('given an unsubscribed listener, should not call its onComplete when stream finishes', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const completed1: boolean[] = [];
+      const completed2: boolean[] = [];
+
+      const unsubscribe1 = registry.subscribe('msg-1', () => {}, (aborted) => completed1.push(aborted));
+      registry.subscribe('msg-1', () => {}, (aborted) => completed2.push(aborted));
+
+      unsubscribe1!();
+      registry.finish('msg-1');
+
+      expect(completed1).toEqual([]);
+      expect(completed2).toEqual([false]);
+    });
+  });
+
+  describe('finish', () => {
+    it('given a finished stream, should call complete callbacks with aborted=false by default', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const abortedValues: boolean[] = [];
+      registry.subscribe('msg-1', () => {}, (aborted) => abortedValues.push(aborted));
+
+      registry.finish('msg-1');
+      expect(abortedValues).toEqual([false]);
+    });
+
+    it('given a finished stream with aborted=true, should call complete callbacks with the correct aborted status', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const abortedValues: boolean[] = [];
+      registry.subscribe('msg-1', () => {}, (aborted) => abortedValues.push(aborted));
+
+      registry.finish('msg-1', true);
+      expect(abortedValues).toEqual([true]);
+    });
+
+    it('removes the stream entry after finishing so getMeta returns undefined', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      registry.finish('msg-1');
+
+      expect(registry.getMeta('msg-1')).toBeUndefined();
+    });
+
+    it('silently ignores finish for unknown messageId', () => {
+      const registry = new StreamMulticastRegistry();
+      expect(() => registry.finish('unknown')).not.toThrow();
+    });
+
+    it('silently ignores a second finish call', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const completed: boolean[] = [];
+      registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));
+
+      registry.finish('msg-1');
+      registry.finish('msg-1'); // second call should be a no-op
+
+      expect(completed).toEqual([false]);
+    });
+  });
+
+  describe('auto-cleanup', () => {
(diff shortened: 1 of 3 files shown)


2witstudios (commented):
**Timer leak on duplicate `register`**

`register` doesn't guard against being called with an already-registered `messageId`. If it happens, the old `cleanupTimeoutId` is overwritten without `clearTimeout`, leaking the timer:

```typescript
registry.register('msg-1', meta1); // schedules cleanupTimeoutId_1
registry.register('msg-1', meta2); // overwrites entry — cleanupTimeoutId_1 is never cleared
```

Suggest either early-returning when the ID already exists (the safer choice, since re-registration likely indicates a caller bug) or clearing the old timer before overwriting.

---

**Missing test for duplicate registration**

There's no test covering this path, which is why the leak isn't caught. A test asserting the intended behavior (e.g., second `register` is a no-op, or that only one cleanup timer fires) would lock this in.

2witstudios (commented):
Review findings from /review — 6 issues across implementation and tests.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export class StreamMulticastRegistry {
+  private entries = new Map<string, StreamEntry>();
+
+  register(messageId: string, meta: StreamMeta): void {
+    const cleanupTimeoutId = setTimeout(() => {
+      this.finish(messageId, true);
+    }, MAX_STREAM_AGE_MS);
+
+    this.entries.set(messageId, {
```
2witstudios, line 34: [Bug] Duplicate `register` leaks the first timer.

If `register` is called twice with the same `messageId` (retry or race), the first `setTimeout` is orphaned — never cleared. It fires 10 minutes later and calls `this.finish(messageId, true)` on whatever happens to be in the registry at that point.

Fix: clear the existing entry's timeout before overwriting.

```ts
register(messageId: string, meta: StreamMeta): void {
  const existing = this.entries.get(messageId);
  if (existing?.cleanupTimeoutId !== null) {
    clearTimeout(existing!.cleanupTimeoutId!);
  }
  const cleanupTimeoutId = setTimeout(() => {
    this.finish(messageId, true);
  }, MAX_STREAM_AGE_MS);
  this.entries.set(messageId, { meta, buffer: [], subscribers: new Map(), finished: false, cleanupTimeoutId });
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (entry.cleanupTimeoutId !== null) {
+      clearTimeout(entry.cleanupTimeoutId);
+      entry.cleanupTimeoutId = null;
+    }
+
+    for (const subscriber of entry.subscribers.values()) {
+      subscriber.onComplete(aborted);
+    }
+
+    this.entries.delete(messageId);
```
2witstudios: [Bug] Subscriber exception in `onComplete` leaks the entry.

`entry.finished = true` is set and the timer is cleared before the subscriber loop runs. If any callback throws, `this.entries.delete(messageId)` never executes — the entry stays in the map forever with no timer to clean it up.

Fix: delete the entry before notifying subscribers (or wrap in try/finally):

```ts
this.entries.delete(messageId);            // ← move up
for (const subscriber of entry.subscribers.values()) {
  subscriber.onComplete(aborted);
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
```diff
@@ -0,0 +1,91 @@
+const MAX_STREAM_AGE_MS = 10 * 60 * 1000;
+
+interface StreamMeta {
```
2witstudios: [Minor] `StreamMeta` is the return type of `getMeta` but is not exported.

Downstream tasks (Tasks 2–5) that store or annotate the result of `getMeta` will have to inline the type or use `ReturnType<typeof streamMulticastRegistry.getMeta>`. Should be `export interface StreamMeta`.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    it('silently ignores a second finish call', () => {
+      const registry = new StreamMulticastRegistry();
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const completed: boolean[] = [];
+      registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));
+
+      registry.finish('msg-1');
+      registry.finish('msg-1'); // second call should be a no-op
```
2witstudios: [Minor] Inline comment violates project comment policy.

The test name `'silently ignores a second finish call'` already states this. Remove the redundant inline comment — it explains what the code does, which is exactly what the project rules prohibit.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
+
+      const received1: string[] = [];
+      const received2: string[] = [];
+
+      const unsubscribe1 = registry.subscribe('msg-1', (text) => received1.push(text), () => {});
+      registry.subscribe('msg-1', (text) => received2.push(text), () => {});
+
+      registry.push('msg-1', 'chunk1');
+      unsubscribe1!();
```
2witstudios: [Minor] Non-null assertion silences a meaningful failure.

If `subscribe` unexpectedly returns `null` for a live stream, `unsubscribe1!()` fails with a misleading `TypeError: unsubscribe1 is not a function` rather than a clear test failure. Prefer an explicit assertion:

```ts
expect(unsubscribe1).not.toBeNull();
(unsubscribe1 as () => void)();
```

Same applies to line 128.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      registry.subscribe('msg-1', () => {}, (aborted) => completedValues.push(aborted));
+
+      registry.finish('msg-1', false);
+
+      vi.advanceTimersByTime(10 * 60 * 1000);
+
+      // Only one completion, not two
+      expect(completedValues).toEqual([false]);
+    });
+  });
```
2witstudios, line 261: [Gap] No test for duplicate `register` behavior.

The orphaned-timer bug above has no test coverage. Add a test like:

```ts
it('re-registering the same messageId clears the previous timer', () => {
  vi.useFakeTimers();
  const registry = new StreamMulticastRegistry();
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' }); // overwrite

  const completed: boolean[] = [];
  registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));

  vi.advanceTimersByTime(10 * 60 * 1000);
  expect(completed).toHaveLength(1); // only one completion, not two
});
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios, line 34: **[Bug]** Duplicate `register` leaks the first timer.

If `register` is called twice with the same `messageId` (retry or race), the first `setTimeout` is orphaned — never cleared. It fires 10 minutes later and calls `this.finish(messageId, true)` on whatever happens to be in the registry at that point.

Fix: clear the existing entry's timeout before overwriting:

```ts
register(messageId: string, meta: StreamMeta): void {
  const existing = this.entries.get(messageId);
  if (existing?.cleanupTimeoutId !== null) {
    clearTimeout(existing!.cleanupTimeoutId!);
  }
  // ... rest unchanged
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: **[Bug]** Subscriber exception in `onComplete` leaks the entry permanently.

`entry.finished = true` is set and the timer is already cleared before this loop runs. If any callback throws, `this.entries.delete(messageId)` on line 83 never executes — the entry stays in the map forever with nothing to clean it up.

Fix: delete the entry before notifying subscribers:

```ts
this.entries.delete(messageId); // ← move above the loop
for (const subscriber of entry.subscribers.values()) {
  subscriber.onComplete(aborted);
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: **[Minor]** `StreamMeta` is the return type of `getMeta` but is not exported.

Downstream tasks (2–5) that store or type-annotate the result of `getMeta` can't reference it without `ReturnType<typeof streamMulticastRegistry.getMeta>`. Should be `export interface StreamMeta`.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: **[Minor]** Inline comment violates project comment policy.

The test name `'silently ignores a second finish call'` already says this. Remove the redundant comment — the project rule is: only comment when the *why* is non-obvious.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: **[Minor]** Non-null assertion silences a meaningful failure.

If `subscribe` unexpectedly returns `null` for a live stream, `unsubscribe1!()` fails with a misleading `TypeError: unsubscribe1 is not a function` instead of a clear assertion failure. Prefer:

```ts
expect(unsubscribe1).not.toBeNull();
(unsubscribe1 as () => void)();
```

Same issue at line 128.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios, line 261: **[Gap]** No test for duplicate `register` — the orphaned-timer bug above has no coverage.

Add a test:

```ts
it('re-registering the same messageId fires onComplete only once', () => {
  vi.useFakeTimers();
  const registry = new StreamMulticastRegistry();
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' }); // overwrite

  const completed: boolean[] = [];
  registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));

  vi.advanceTimersByTime(10 * 60 * 1000);
  expect(completed).toHaveLength(1); // not two
});
```

2witstudios (commented):
Review findings from /review — 6 issues across implementation and tests.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios, line 34: **[Bug]** Duplicate `register` leaks the first timer.

If `register` is called twice with the same `messageId` (retry or race), the first `setTimeout` is orphaned — never cleared. It fires 10 minutes later and calls `this.finish(messageId, true)` on whatever happens to be in the registry at that point.

Fix: clear the existing entry's timeout before overwriting:

```ts
register(messageId: string, meta: StreamMeta): void {
  const existing = this.entries.get(messageId);
  if (existing?.cleanupTimeoutId !== null) {
    clearTimeout(existing!.cleanupTimeoutId!);
  }
  // ... rest unchanged
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: **[Bug]** Subscriber exception in `onComplete` leaks the entry permanently.

`entry.finished = true` is set and the timer is already cleared before this loop runs. If any callback throws, `this.entries.delete(messageId)` never executes — the entry stays in the map forever with nothing to clean it up.

Fix: delete the entry before notifying subscribers:

```ts
this.entries.delete(messageId); // ← move above the loop
for (const subscriber of entry.subscribers.values()) {
  subscriber.onComplete(aborted);
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: **[Minor]** `StreamMeta` is the return type of `getMeta` but is not exported.

Downstream tasks (2–5) that store or type-annotate the result of `getMeta` can't reference it without `ReturnType<typeof streamMulticastRegistry.getMeta>`. Should be `export interface StreamMeta`.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: **[Minor]** Inline comment violates project comment policy.

The test name `'silently ignores a second finish call'` already says this. Remove the redundant comment — the project rule is: only comment when the *why* is non-obvious.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: **[Minor]** Non-null assertion silences a meaningful failure.

If `subscribe` unexpectedly returns `null` for a live stream, `unsubscribe1!()` fails with a misleading `TypeError: unsubscribe1 is not a function` instead of a clear assertion failure. Prefer:

```ts
expect(unsubscribe1).not.toBeNull();
(unsubscribe1 as () => void)();
```

Same issue at line 128.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios, line 261: **[Gap]** No test for duplicate `register` — the orphaned-timer bug above has no coverage.

Add to the `auto-cleanup` block:

```ts
it('re-registering the same messageId fires onComplete only once', () => {
  vi.useFakeTimers();
  const registry = new StreamMulticastRegistry();
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' }); // overwrite

  const completed: boolean[] = [];
  registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));

  vi.advanceTimersByTime(10 * 60 * 1000);
  expect(completed).toHaveLength(1); // not two
});
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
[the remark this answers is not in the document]
2witstudios: Fixed in commit e48d02245. `finish()` now deletes the entry from the map **before** iterating subscribers, and wraps each `onComplete` call in try/catch so a throwing callback cannot prevent others from being notified or leave the entry orphaned:

```ts
this.entries.delete(messageId);  // ← runs before callbacks
for (const subscriber of entry.subscribers.values()) {
  try { subscriber.onComplete(aborted); } catch { /* isolate */ }
}
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
```diff
@@ -0,0 +1,91 @@
...
+    });
+  }
+
+  push(messageId: string, text: string): void {
+    const entry = this.entries.get(messageId);
+    if (!entry || entry.finished) return;
+
+    entry.buffer.push(text);
+    for (const subscriber of entry.subscribers.values()) {
+      subscriber.onChunk(text);
```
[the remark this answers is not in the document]
2witstudios: Fixed in commit e48d02245. `push()` now wraps each `onChunk` call in try/catch so a throwing subscriber cannot interrupt fanout to the remaining subscribers:

```ts
for (const subscriber of entry.subscribers.values()) {
  try { subscriber.onChunk(text); } catch { /* isolate */ }
}
```

A new test in the `resilience` block verifies this: *'given a subscriber whose onChunk throws, should not interrupt fanout to remaining subscribers'*.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Fixed in commit e48d02245. `register()` now clears the existing entry's timeout before overwriting:

```ts
const existing = this.entries.get(messageId);
if (existing?.cleanupTimeoutId !== null && existing?.cleanupTimeoutId !== undefined) {
  clearTimeout(existing.cleanupTimeoutId);
}
```

A new test in the `auto-cleanup` block verifies: *'given register is called twice for the same messageId, should call onComplete only once after 10 minutes'*.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Fixed in commit e48d02245. The entry is now deleted **before** iterating subscribers, and each callback is wrapped in try/catch:

```ts
this.entries.delete(messageId);  // ← moved above loop
for (const subscriber of entry.subscribers.values()) {
  try { subscriber.onComplete(aborted); } catch { }
}
```

A new `resilience` test verifies: *'given a subscriber whose onComplete throws, should still remove the entry and notify remaining subscribers'*.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Fixed in commit e48d02245. Changed to `export interface StreamMeta` — downstream tasks can now import and annotate the type directly.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Fixed in commit e48d02245. Removed the `// second call should be a no-op` comment — the test name already states this.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Fixed in commit e48d02245. Both `unsubscribe1!()` call sites now use an explicit assertion before calling:

```ts
expect(unsubscribe1).not.toBeNull();
(unsubscribe1 as () => void)();
```

This produces a clear assertion failure message if subscribe unexpectedly returns null.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Fixed in commit e48d02245. Added to the `auto-cleanup` describe block:

```ts
it('given register is called twice for the same messageId, should call onComplete only once after 10 minutes', () => {
  vi.useFakeTimers();
  const registry = new StreamMulticastRegistry();
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
  registry.register('msg-1', { pageId: 'page-1', userId: 'user-1' });
  const completed: boolean[] = [];
  registry.subscribe('msg-1', () => {}, (aborted) => completed.push(aborted));
  vi.advanceTimersByTime(10 * 60 * 1000);
  expect(completed).toHaveLength(1);
});
```

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/stream-multicast-registry.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

apps/web/src/lib/ai/core/__tests__/stream-multicast-registry.test.ts
2witstudios: Duplicate thread — see corresponding reply above. All issues addressed in commit e48d02245.

PR #1152, opened 2026-04-29
fix(ai): shared conversation init — deterministic ID prevents multiplayer race

2witstudios:
## Summary

Each conversation is a session that multiple users can join. The stream completion guard in `AiChatView` — `stream.conversationId === currentConversationId` — only appends a remote AI response to your view if you're in the same conversation as the sender. Users can be in independent conversations on the same page, or the same one.

**The bug**: `initializeChat` always POSTed a new conversation on every page load, so each user got a different `conversationId` and the guard never matched for users in the same conversation.

**Fix**: GET the most recent existing conversation on open instead of POSTing. Falls back to a `${page.id}-default` ID when none exist — the server stores whatever string you pass as `conversationId`, so this becomes a real conversation once the first message is saved.

**Also**: all three `fetchWithAuth` calls in `initializeChat` now share an `AbortController` for clean unmount cancellation.

## Test plan

- [x] Given existing conversation → GET list, load messages, no POST
- [x] Given no conversations → derive page-scoped ID, no POST
- [x] Given GET fails (non-ok) → fall back to page-scoped ID, no POST
- [x] Given two users on same page → both load same conversation via GET
- [x] Given user clicks New Chat → `createConversation` called (existing behavior unchanged)
- [x] 5/5 tests passing in `__tests__/AiChatView.test.tsx`

### DIFF
--- a/apps/web/src/components/layout/middle-content/page-views/ai-page/AiChatView.tsx
+++ b/apps/web/src/components/layout/middle-content/page-views/ai-page/AiChatView.tsx
@@ -274,6 +274,7 @@ const AiChatView: React.FC<AiChatViewProps> = ({ page }) => {
               const { messages: loaded } = msgResponse.ok
                 ? ((await msgResponse.json()) as ConversationMessagesResponse)
                 : { messages: [] as UIMessage[] };
+              if (controller.signal.aborted) return;
               setCurrentConversationId(conv.id);
               setMessages(loaded ?? []);
               setIsInitialized(true);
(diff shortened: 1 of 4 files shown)


2witstudios:
Addressed in 08ee2a271:

- Added `ConversationListResponse` and `ConversationMessagesResponse` types for the two `json()` calls in `initializeChat`
- Updated the array guard from `list?.length > 0` to `list && list.length > 0` so TypeScript can narrow the type
- The test file's `wasGetCalled`/`wasPostCalled` helpers already use `Parameters<typeof fetchWithAuth>[]` cast (added in dc7804d9b) which addresses the actionable comment about `any[]` in mock.calls

PR #1164, opened 2026-04-30
feat(ai): streamMulticastRegistry richer metadata — displayName, conversationId, tabId

2witstudios:
## Summary

- Extend `StreamMeta` with `displayName`, `conversationId`, and `tabId` so downstream consumers (active-streams endpoint, stream-join route) get full context without re-querying.
- Reorder `apps/web/src/app/api/ai/chat/route.ts`: read `X-Tab-Id` header and resolve `displayName` from `userProfiles` **before** calling `streamMulticastRegistry.register()`. Eliminates the prior window where the registry held incomplete `{ pageId, userId }` metadata.
- Pass the resolved `tabId` through to both `register()` and `broadcastAiStreamStart` (the `triggeredBy.tabId` field is already optional on `AiStreamStartPayload`).
- Update affected tests (registry unit tests, stream-join route tests, stream-socket-events tests) to construct full `StreamMeta` objects.

## Why

Wave 2 of the deterministic AI chat streaming work needs rich metadata in the multicast registry so other tabs can join and render `active streams` correctly without round-tripping the DB. Without this, every consumer would have to re-look-up the conversation, display name, and originating tab.

## Test plan

- [x] `pnpm --filter web typecheck` passes
- [x] `vitest run` for the three affected suites (registry, stream-join route, stream-socket-events) — 56/56 tests pass

### DIFF
--- a/apps/web/src/app/api/ai/chat/__tests__/stream-socket-events.test.ts
+++ b/apps/web/src/app/api/ai/chat/__tests__/stream-socket-events.test.ts
@@ -327,7 +327,13 @@ describe('POST /api/ai/chat — stream socket events', () => {
 
       expect(mockRegistryRegister).toHaveBeenCalledWith(
         'test-message-id',
-        { pageId: 'page-1', userId: 'user-1' }
+        expect.objectContaining({
+          pageId: 'page-1',
+          userId: 'user-1',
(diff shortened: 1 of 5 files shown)


2witstudios:
@coderabbitai review

PR #1165, opened 2026-04-30
feat(ai): persist page AI chat streams to aiStreamSessions table

2witstudios:
## Summary
- Page AI chat route persists stream lifecycle to `aiStreamSessions`:
  - **INSERT (awaited)** before `broadcastAiStreamStart` so the row is committed before any client receives a `chat:stream_start` event or hits `GET /api/ai/chat/active-streams`. Uses `onConflictDoUpdate` (target=`messageId`, set=`{status:'streaming', completedAt:null}`) so re-registering the same id is a no-op refresh, never a PK violation.
  - **UPDATE (fire-and-forget)** inside `finishMulticast` to `complete` (success path) or `aborted` (onAbort + outer-catch). Wrapped so DB hiccups never abort the response.
- Both DB writes log a `loggers.ai.warn(...)` on failure with `messageId` + error message — diagnosable instead of silently swallowed.
- New colocated test file `apps/web/src/app/api/ai/chat/__tests__/db-stream-persistence.test.ts` (10 tests) covers AC1–AC4 plus an INSERT-happens-before-broadcast deferred-insert regression test.

## Why awaited INSERT (vs. fire-and-forget)
Awaiting the INSERT is strictly stronger than chaining the UPDATE in `.finally()` of a fire-and-forget INSERT: by the time the route reaches `createUIMessageStream`, the row is already visible to other transactions. Any `finishMulticast` call (success, abort, or thrown stream) therefore updates an existing row.

## Test plan
- [x] `pnpm --filter web test src/app/api/ai/chat` — 116/116 pass (including new 10-test suite)
- [x] `pnpm --filter web lint` — no new warnings on touched files

### DIFF
--- a/apps/web/src/app/api/ai/chat/__tests__/db-stream-persistence.test.ts
+++ b/apps/web/src/app/api/ai/chat/__tests__/db-stream-persistence.test.ts
@@ -293,6 +293,7 @@ import { authenticateRequestWithOptions } from '@/lib/auth';
 import type { SessionAuthResult } from '@/lib/auth';
 import { db } from '@pagespace/db/db';
 import { createUIMessageStream } from 'ai';
+import { broadcastAiStreamStart } from '@/lib/websocket';
 
 // ============================================================================
 // Fixtures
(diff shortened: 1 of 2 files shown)


2witstudios:
Replying to CodeRabbit's review (commit-hash anchor: 52d78d0):

**Major: race between fire-and-forget INSERT and UPDATE** — fixed in commit `72fc8fe`. Rather than chaining the UPDATE in `.finally()` of the INSERT promise, I made the **INSERT itself awaited** before any code path that can register `finishMulticast` callbacks. So:

- `await db.insert(aiStreamSessions)…onConflictDoUpdate(…)` runs **before** `broadcastAiStreamStart`, **before** `createUIMessageStream`, and **before** `streamText.onAbort` / `onFinish` are even registered.
- Any subsequent `finishMulticast(...)` call (success, abort, or outer-catch) is therefore guaranteed to run on a row that already exists.
- This is strictly stronger than chaining `.finally()`: in your suggested form, both writes still race the network/Postgres pool; in this form, the INSERT has fully resolved at the SDK boundary before the request handler proceeds.

Thread is auto-resolved since the suggested fix is no longer applicable to the current diff.

**Nit: deferred-insert race regression test** — added in commit `e145712d` as `INSERT-happens-before-broadcast invariant`. It defers `db.insert` on a controllable promise, polls until INSERT is issued, asserts `broadcastAiStreamStart` was *not* called yet, then resolves and asserts the broadcast then fires. This codifies the new invariant directly.

Total tests: 10/10 in `db-stream-persistence.test.ts`, 116/116 in `apps/web/src/app/api/ai/chat`. Typecheck clean.

PR #1226, opened 2026-05-04
fix(socket): stop force-reconnecting healthy sockets on auth:refreshed

2witstudios:
## Summary

- AI chat (Global Assistant) was randomly snapping to the bottom mid-read whenever the tab regained focus or the desktop token-refresh timer fired.
- Root cause: `useSocketStore` was force-reconnecting the Socket.IO client on every `auth:refreshed` event, even when the socket was already connected. The reconnect flipped `connectionStatus`, which `GlobalChatContext` interprets as a missed-events signal and reloads messages — replacing the `initialMessages` array, recomputing `chatConfig`, remounting `MessageRenderer`s, and tripping `use-stick-to-bottom`'s grow-from-zero auto-scroll.
- Socket.IO consumes the auth token only at handshake. A live connection stays authenticated regardless of cookie/session refreshes, so the proactive reconnect was pure churn. The handler now no-ops when the socket is connected; the existing `connect_error` path still recovers genuinely-rejected handshakes with a fresh token.

The same chain runs for multi-user agent chat via `useAgentChannelMultiplayer`, so it benefits too. Channels and inbox channels just rejoin via socket on reconnect — they don't refetch — so they were unaffected and remain so.

## Test plan

- [ ] `pnpm dev`, open a global AI chat with enough history to scroll
- [ ] Scroll up to leave the view above the bottom
- [ ] In devtools console: `window.dispatchEvent(new CustomEvent('auth:refreshed'))` — scroll position should stay put; no "proactively reconnecting" log; no "retrying disconnected socket" log either (because socket is connected)
- [ ] Background the tab >12 minutes, return — same expectation (visibility-change refresh path)
- [ ] Force a real disconnect: DevTools → Network → Offline, wait for `connectionStatus: 'disconnected'`, then Online — socket reconnects, chat is functional, the disconnected-retry branch fires if `auth:refreshed` happens to land during the gap
- [ ] `pnpm --filter web typecheck` passes (verified locally)

### DIFF
--- a/apps/web/src/__tests__/socket-integration.test.ts
+++ b/apps/web/src/__tests__/socket-integration.test.ts
@@ -312,40 +312,42 @@ describe('Socket.IO Integration', () => {
     });
   });
 
-  describe('proactive reconnection', () => {
-    it('given auth:refreshed event while connected, should force reconnect with new token', async () => {
+  describe('auth:refreshed handling', () => {
+    it('given auth:refreshed event while connected, should leave the live socket untouched', async () => {
       const { connect } = useSocketStore.getState();
       await connect();
 
       mockSocket.connected = true;
       mockSocket._trigger('connect');
 
-      // Verify initial connection
       expect(useSocketStore.getState().connectionStatus).toBe('connected');
 
-      // Simulate token refresh event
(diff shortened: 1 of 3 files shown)


2witstudios:
@coderabbitai review

PR #1252, opened 2026-05-05
fix(threads): hardening — DM relations, askResult guard, follow root-only (PR 6a)

2witstudios:
## Summary
- DM repository hydrates `sender`/`reactions`/`file` relations on `listDmThreadReplies` and `listActiveMessages` (channel parity via a new `dmMessageWith` const)
- `isAskAgentResult` type predicate replaces an unjustified cast in the channel mention responder; malformed agent results now log + skip rather than crashing the broadcast (logged at `.error` with received type/keys for diagnosability)
- Follow POST (channel + DM) returns the structured error code `parent_not_top_level` for non-root targets; DELETE stays permissive (idempotent unfollow)
- Route tests now assert the `thread_updated` recipient list explicitly (sorted-copy comparison), pin `broadcastThreadReplyCountUpdated` args, and cover the zero-followers edge case
- New exported `DmMessageWithRelations` type pins the inferred shape of `listDmThreadReplies` so the public contract no longer drifts silently

## Out of scope (per the plan)
- The "stop mutating ORM result" audit finding does not apply: `fanOutChannelInboxUpdate` on master already builds a Set with the explicit "Build a local member set so we never mutate the ORM-returned array" comment, and there are no `.push()` calls into ORM-returned arrays in either route.
- Repository-test discipline (PR 6b) and UI/store edge-case tests (PR 6c) are sibling PRs touching disjoint files.

## Follow-up (out of scope for 6a)
- The thread-reply-creation routes (`channels/[pageId]/messages/route.ts`, `messages/[conversationId]/route.ts`) still return freeform error strings for `kind: 'parent_not_top_level'` while the follow POST now returns the structured `parent_not_top_level` code. Aligning those is a small follow-up — kept out of this PR to avoid scope creep.

Part of the threads hardening epic. Plan: `/Users/jono/.claude/plans/channels-and-dm-s-need-imperative-bird.md`

## Test plan
- [x] `pnpm --filter @pagespace/lib test` — 3953/3953 passing
- [x] Targeted route + responder tests — 99/99 passing across the touched files
- [x] `pnpm typecheck` (recursive across 12 workspaces) — clean
- [x] `pnpm lint` — clean (only a pre-existing unrelated `react-hooks/exhaustive-deps` warning in QuickCreatePalette)
- [x] Unit Tests CI — green
- [x] Lint & TypeScript Check CI — green
- [x] Security Test Suite, Dependency Audit, Secret Scanning, Static Security Analysis, CodeQL — green
- [ ] Manual smoke: DM thread reply renders sender name + avatar in the panel
- [ ] Manual smoke: POST against a thread-reply id returns 400 `{ "error": "parent_not_top_level" }`
- [ ] Manual smoke: malformed agent result skips reply without persisting

### DIFF
--- a/apps/web/src/app/api/channels/[pageId]/messages/[messageId]/follow/__tests__/route.test.ts
+++ b/apps/web/src/app/api/channels/[pageId]/messages/[messageId]/follow/__tests__/route.test.ts
@@ -123,7 +123,7 @@ describe('POST /api/channels/[pageId]/messages/[messageId]/follow', () => {
     expect(res.status).toBe(404);
   });
 
-  it('returns 400 when the message is itself a thread reply (followers attach to roots only)', async () => {
+  it('returns 400 with parent_not_top_level when the message is itself a thread reply (followers attach to roots only)', async () => {
     mockFindChannelMessageInPage.mockResolvedValueOnce({
       id: MSG_ID,
       pageId: PAGE_ID,
@@ -132,6 +132,8 @@ describe('POST /api/channels/[pageId]/messages/[messageId]/follow', () => {
     });
     const res = await callPost();
     expect(res.status).toBe(400);
+    expect(await res.json()).toEqual({ error: 'parent_not_top_level' });
+    expect(mockAddChannelThreadFollower).not.toHaveBeenCalled();
   });
 });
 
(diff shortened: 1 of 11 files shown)


2witstudios:
@coderabbitai review

PR #1255, opened 2026-05-05
feat(quotes): inline quote replies for channels & DMs

2witstudios:
## Summary
- Slack/Twitter-style **inline** quote-reply embeds for channels and DMs — hover any message → **Quote reply** → composer chip → send → the embed renders inline above the new message. Orthogonal to the just-shipped thread panel: quote replies stay in the main feed instead of disappearing into a side panel.
- One additive `quotedMessageId` self-FK on `channelMessages` and `directMessages` (mirrors the existing `parentId` / `mirroredFromId` shape), one shared read-time enrichment helper in `packages/lib`, zero new endpoints, realtime events, tables, or stores.
- Eric-Elliott approved: pure data flow (column → batched IN-query → API field → renderer prop), composition over coupling (one helper serves both surfaces), single source of truth (FK is canonical, snapshot is read-time), minimal surface.

## Architecture

10 commits, sequenced for review (plus a master merge):

| # | Commit | What it does |
|---|---|---|
| 1 | `docs(tasks): plan inline quote replies epic` | Epic plan + plan.md reference |
| 2 | `feat(quotes): prep — schema column` | Migration `0121_complete_lila_cheney.sql` (originally 0120; renumbered after merging master's 0120 task-items migration) — 2 ADD COLUMN, 2 FK constraints (`onDelete: 'set null'`), 2 indexes |
| 3 | `refactor(quotes): move buildThreadPreview to lib` | Relocate the 100-char ellipsis helper into `packages/lib/services/preview.ts` so the enrichment helper can share it |
| 4 | `feat(quotes): batched quote enrichment helper` | Pure `attachQuotedMessages(rows, scope)` — single batched IN-query, dedupes ids, never filters by `isActive` so soft-deleted sources resolve to a tombstone snapshot |
| 5 | `feat(quotes): plumb quotedMessageId through repos` | Optional field on `InsertChannelMessageInput` / `InsertDmMessageInput`; thread-reply helpers untouched |
| 6 | `feat(quotes): wire channel API for quote replies` | POST validates same-page + active + top-level; GET/POST run enrichment; `parent_not_top_level` phrasing reused for symmetry with thread rejection |
| 7 | `feat(quotes): wire DM API for quote replies` | Symmetric — same-conversation + active + top-level |
| 8 | `feat(quotes): MessageQuoteBlock embed component` | Presentational embed with author + name + relative time + truncated snippet rendered through `MessagePartRenderer` (so inline mentions in the snapshot stay parsed); tombstone branch for null/inactive |
| 9 | `feat(quotes): composer chip in ChannelInput` | Controlled `quotedPreview` + `onClearQuote` props; chip mirrors the attachment-preview language |
| 10 | `feat(quotes): wire views with quote-reply UI` | ChannelView + DM page hold the quote state, render `MessageQuoteBlock` above content, add **Quote reply** to hover dropdown (visible to all; Edit/Delete remain author-gated), include `quotedMessageId` in optimistic insert + POST body |
| — | `Merge remote-tracking branch 'origin/master'` | Pulled master's task-items migration (#1254) and the testDbState boundary refactor (#1253). Quote-reply tests rewritten in the new `testDbState` style; `quotedMessageId` added to the in-memory schema column lists |

## Key design choices

- **Read-time snapshot, not write-time freeze.** The `quotedMessage` payload is denormalized at read time from the FK; editing the original updates every embed automatically. Slack/Twitter both do this — a frozen JSON copy would create a second source of truth that drifts.
- **`onDelete: 'set null'`, not cascade.** A quoting reply outlives a hard-deleted source; soft-deletes leave the FK intact and the renderer reads `isActive` to show a tombstone.
- **No second permission check on the quoted message.** Same-channel / same-conversation gating is the only ACL gate needed — the requester already had read access. `findChannelMessageInPage` enforces `pageId` verbatim; reused.
- **No new Drizzle self-relation in `messageWith`.** That would eagerly load soft-deleted parents into every read. The standalone enrichment helper batches a single `SELECT ... WHERE id IN (...)` instead.
- **AI mention responder is intentionally untouched.** It scans the user's typed `messageContent` only, so quoting an @-agent does NOT re-trigger; an in-line code comment was added so a future reader does not "fix" it.

## Out of scope (explicit)

- ChannelView's missing thread-panel mount (`useThreadPanelStore` is wired in DM but not in ChannelView). Surfaced during planning; deferred to a separate epic to keep this PR scoped to quote replies.
- Quote-of-a-quote nesting depth limits (depth 1 by display, matches Slack/Twitter).
- Historical backfill (column is nullable, no data to migrate).

## Test plan
- [ ] `pnpm db:migrate` against a dev DB — migration applies cleanly
- [ ] `pnpm test:unit` — green, including the four new `quotedMessageId` pass-through tests in the new testDbState style; 0 regressions (pre-existing `admin/gift-subscription` integration failures need a postgres `test` role, unrelated)
- [ ] `pnpm typecheck` — green across `@pagespace/db`, `@pagespace/lib`, `web`
- [ ] `pnpm lint` — green
- [ ] In `pnpm dev`: hover any channel message → **Quote reply** appears in the dropdown
- [ ] Click → chip with author + snippet appears above the composer; X dismisses
- [ ] Send → new message renders with the embed inline above the body
- [ ] Edit the original message → embed reflects the change on next page load
- [ ] Soft-delete the original → embed shows the **Original message deleted** tombstone
- [ ] Open a second tab as another user → realtime `new_message` delivers the quote-bearing message with the embed populated
- [ ] Try to quote-reply against a thread-only message (`parentId !== null`) → API rejects with `Threads are exactly one level deep`
- [ ] Try to quote-reply against a message in another channel → API rejects with `Quoted message not found`
- [ ] Try to send `parentId` + `quotedMessageId` together → API rejects with `Quote replies cannot be combined with thread replies`
- [ ] Repeat the full flow in a DM conversation

## Release Notes

* **New Features**
  * Added inline quote-reply functionality for channels and direct messages—quote any message to reference it alongside your reply
  * Quoted messages display inline within message bubbles for context
  * Message composer shows a preview chip of the quoted message with an option to clear it

* **Improvements**
  * Enhanced message retrieval to include quote context and snapshots


2witstudios:
@coderabbitai thanks for the review. All 8 inline comments addressed in `bda96844a`:

| # | File | Comment | Fix |
|---|---|---|---|
| 1 | `quote-enrichment.ts:38` | AI sender metadata loss for channel quotes | Helper now prefers `aiMeta.senderName`/`aiMeta.agentPageId` over the human `user` row when the quoted channel message was AI-authored; new test covers it |
| 2 | `quote-enrichment.ts:52` | Deleted message text shipped in tombstone snapshots | `contentSnippet` is blanked to `""` when `isActive === false`; new test asserts the deleted body never crosses the wire |
| 3 | `package.json` | Missing `typesVersions` entries for new exports | Added `services/preview` and `services/quote-enrichment` entries to mirror the existing pattern |
| 4 | `ChannelView.tsx:468` | Grouped-message dropdown leaked Edit/Delete to non-authors | Continuation dropdown now mirrors the first-row structure: `Quote reply` is universal, `Edit`/`Delete` gated by `showOwnerActions` |
| 5 | `DM page.tsx:116` | Quote state leaked across conversation navigation | The existing `closeThread()` effect on `[conversationId]` now also clears `quotedMessageId` + `activeQuotedSnapshot` |
| 6 | `DM page.tsx:368` | Quote chip not restored on send failure | Catch block now restores `quotedMessageId` + `activeQuotedSnapshot` when `activeQuoteId` was set; same fix applied to `ChannelView.handleSubmit` for parity |
| 7 | `ChannelInput.tsx:304` | Non-functional dismiss X when `onClearQuote` absent | The X button only renders when `onClearQuote` is truthy |
| 8 | `MessageQuoteBlock.tsx:65` | Space-key scrolled the page during keyboard activation | Added `e.preventDefault()` before `handleClick()` for both Enter and Space |

Verified locally: `pnpm --filter @pagespace/lib test` (3979 passing), `pnpm --filter web typecheck` clean, targeted UI suites (ChannelInput 15/15, DM page 8/8, MessageQuoteBlock 6/6) green.

PR #1264, opened 2026-05-06
feat(agents): scope adapter tools per integration

2witstudios:
## Summary
- Surface each integration's tools as a checkbox list directly under the **Default Tools** card, mirroring the native styling — name, description, Select All / Deselect All, and a "Selected N of M" counter per integration.
- Per-tool scoping: toggling a checkbox writes back the full allowed list to `integrationToolGrants.allowedTools`. Legacy `allowedTools = null` is rendered as all non-dangerous tools checked (mirroring `is-tool-allowed.ts`); unchecking the last tool sends `[]`.
- Remove the per-grant **rate limit** input from the UI (DB column and API field retained — UI surface only).

## Changes
- `apps/web/src/app/api/agents/[agentId]/integrations/route.ts` — GET response now includes a sanitized `tools[]` (`id`, `name`, `description`, `category`) on `connection.provider`. Execution config, input schema, and rate-limit metadata stay server-side. Guarded with `Array.isArray` so a malformed custom-provider config returns `tools: []` instead of 500ing.
- `packages/lib/src/integrations/repositories/grant-repository.ts` — widened `GrantWithConnection` so the route can read `provider.config`.
- `apps/web/src/components/integrations/types.ts` — extended `SafeGrant.connection.provider` with `tools[]`.
- `apps/web/src/components/ai/page-agents/AgentIntegrationsPanel.tsx` — full rewrite of the rendered tree: card titled "Integration Tools" matching the Default Tools styling; per-integration checkbox list; Select All / Deselect All; counter. Removed the rate-limit input and the badge-with-X allowed-tools list. Read-only toggle preserved. `getEffectiveAllowed` excludes dangerous-category tools when `allowedTools === null` so a routine non-dangerous edit cannot silently elevate dangerous tools (matches `is-tool-allowed.ts:55-62`).
- `apps/web/src/components/ai/page-agents/__tests__/AgentIntegrationsPanel.test.tsx` — fixtures carry `provider.tools`; new tests for Select All, Deselect All, empty tool list, and the dangerous-tool semantic under `allowedTools === null`.
- `apps/web/src/app/api/agents/[agentId]/integrations/__tests__/route.test.ts` — new tests asserting the response surfaces only the sanitized fields and never leaks `execution`/`inputSchema`/`rateLimit`; parametrized cases (null config, missing tools key, non-array tools, object tools) all yield `tools: []` with status 200.

## Test plan
- [x] `pnpm exec vitest run src/components/ai/page-agents/__tests__/AgentIntegrationsPanel.test.tsx` — 21/21 pass
- [x] `pnpm exec vitest run src/app/api/agents/[agentId]/integrations/__tests__/route.test.ts` — 6/6 pass
- [x] `pnpm --filter @pagespace/lib typecheck` clean
- [x] No new TS errors in touched web files (verified via `tsc --noEmit`)
- [x] `eslint` clean on touched files
- [ ] Manual: open Page AI → Settings, toggle a connection on, verify per-tool checkboxes render with name + description and persist on refresh
- [ ] Manual: uncheck a tool → trigger a chat turn → confirm the agent cannot invoke that tool (enforced by existing `convertIntegrationToolsToAISDK` filter)
- [ ] Manual: confirm the rate-limit input is gone and the read-only toggle still works

### DIFF
--- a/apps/web/src/app/api/agents/[agentId]/integrations/__tests__/route.test.ts
+++ b/apps/web/src/app/api/agents/[agentId]/integrations/__tests__/route.test.ts
@@ -149,7 +149,12 @@ describe('GET /api/agents/[agentId]/integrations response shape', () => {
     expect(provider.tools[0]).not.toHaveProperty('rateLimit');
   });
 
-  it('returns an empty tools array when provider config is missing or malformed', async () => {
+  it.each([
+    ['null config', { config: null }],
+    ['missing tools key', { config: {} }],
+    ['non-array tools', { config: { tools: 'oops' } }],
+    ['object tools', { config: { tools: { id: 'x' } } }],
+  ])('returns an empty tools array when provider has %s', async (_label, providerExtras) => {
     mockListGrantsByAgent.mockResolvedValue([
       {
         id: 'grant-1',
@@ -167,7 +172,7 @@ describe('GET /api/agents/[agentId]/integrations response shape', () => {
           provider: {
             slug: 'custom',
             name: 'Custom',
-            config: null,
+            ...providerExtras,
           },
         },
       },
@@ -178,6 +183,7 @@ describe('GET /api/agents/[agentId]/integrations response shape', () => {
       { params: Promise.resolve({ agentId: mockAgentId }) }
     );
 
+    expect(response.status).toBe(200);
     const body = await response.json();
     expect(body.grants[0].connection.provider.tools).toEqual([]);
   });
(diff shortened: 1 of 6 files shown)


2witstudios:
@chatgpt-codex-connector — both addressed in 18d6cef.

**P1 (null grants / dangerous tools)**: `getEffectiveAllowed` now mirrors `is-tool-allowed.ts:55-62` semantics — under `allowedTools === null`, `dangerous`-category tools are filtered out of the effective set, so a routine non-dangerous edit serializes back an array that *excludes* dangerous tools, preserving the implicit grant. Explicit "Select All" still writes every id, matching user intent. New panel test (`keeps dangerous tools unchecked when allowedTools is null`) locks this in.

**P2 (malformed provider config)**: `Array.isArray` guard added before mapping in `route.ts`, so a non-array `provider.config.tools` yields `tools: []` rather than a 500. New route test (`returns an empty tools array when provider config is missing or malformed`) covers this — the malformed-array variant is now also implicitly covered, since the same code path triggers the guard.

PR #1358, opened 2026-05-15
fix(voice): block auto-listen during tool calls

2witstudios:
## Summary

- In conversation mode, auto-listen fired 300ms after TTS ended regardless of whether the AI stream was still active
- During tool calls, this allowed the user to speak into an in-progress response
- Added `isAIStreaming` option to `useVoiceMode`, tracked via a ref, and added it as a guard on the auto-listen timer

## Test plan

- [ ] Enable voice conversation mode in an AI chat
- [ ] Ask a question that triggers a tool call (e.g. search, calendar lookup)
- [ ] Confirm mic does **not** auto-activate while the tool is executing
- [ ] Confirm mic **does** auto-activate once the full response finishes
- [ ] Confirm manual barge-in via the mic button still works during tool calls
- [ ] Confirm tap-to-speak mode is unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatic voice listening is now blocked while the AI is actively streaming or processing, preventing interruptions during responses.
  * Transcription filtering reduced so brief, near-silence phrases like "thanks" and "bye" are no longer suppressed, improving captured user utterances.

[Review Change Stack](https://app.coderabbit.ai/change-stack/2witstudios/PageSpace/pull/1358?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/src/components/ai/voice/VoiceCallPanel.tsx
+++ b/apps/web/src/components/ai/voice/VoiceCallPanel.tsx
@@ -61,6 +61,7 @@ export function VoiceCallPanel({
   } = useVoiceMode({
     onSend,
     onStopStream,
+    isAIStreaming,
   });
 
   const isOwnerActive = isEnabled && activeOwner === owner;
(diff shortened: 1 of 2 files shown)


2witstudios:
Thanks for the review @chatgpt-codex-connector.

Both flagged items are pre-existing behaviour unrelated to this PR's diff:

**P1 (VAD in tap-to-speak):** `setupVAD` being called for all modes predates this PR. This PR only adds `isAIStreaming` to the options interface and a single guard on the auto-listen timer in `speak()`'s `onended` callback. The `startListening` / `setupVAD` paths are untouched.

**P2 (Whisper artifact filter):** The `WHISPER_ARTIFACTS` set is also pre-existing and outside this PR's scope.

Both are worth addressing as follow-on work, but they're not regressions from this change.

PR #1397, opened 2026-05-23
feat(ai): expose OpenRouter free tier as a user-selectable provider

2witstudios:
Expose `openrouter_free` as a user-selectable provider in cloud mode, with a fully dynamic model list fetched live from the OpenRouter API.

### Core feature
- Added `openrouter_free` to `USER_VISIBLE_PROVIDERS` in the cloud settings API so users can select it in the model picker
- Added `/api/ai/openrouter/models` route that fetches OpenRouter's public model catalog, filters to models with `:free` suffix and zero prompt pricing, and returns them in the same `{ success, models }` shape used by Ollama/LM Studio
- Removed the hardcoded static `openrouter_free` model list from `ai-providers-config.ts` — UI now fetches live
- Updated `ProviderModelSelector` to lazily fetch and cache free models, with the same loading/empty-state pattern used for Ollama and LM Studio

### Security enforcement
- `PATCH /api/ai/settings` now rejects `openrouter_free` + any model that doesn't end with `:free` (HTTP 400)
- `createAIProvider` (provider-factory.ts) also enforces the `:free` suffix before forwarding to OpenRouter backend, so no paid model can slip through even if the settings check is bypassed

### Code quality
- `filterFreeModels` extracted to `filter-utils.ts` — Next.js 15 prohibits arbitrary named exports from route files; keeping it in route.ts was causing the build to fail with `"filterFreeModels" is not a valid Route export field`
- `filterFreeModels` uses `Object.fromEntries` instead of reduce+spread (O(n²) allocations)
- Removed `@pagespace/lib` regex alias from `vitest.config.ts` — violates project rule; `./logging/logger-config` subpath export already exists in `packages/lib/package.json`
- `next-server-stub.ts` extended to export `NextRequest` (was missing, causing all tests that import `NextRequest` to fail silently)

### Tests added
- Full test suite for `GET /api/ai/openrouter/models`: auth, 503 when unconfigured, model filtering, fetch failures, unexpected errors
- `filterFreeModels` unit tests covering `:free` + zero-pricing filter logic
- Settings route tests: `openrouter_free` availability, key-unconfigured 503, key-present 200, **non-`:free` model rejection 400** (closes the P1 review concern)

## How to verify
1. Set `OPENROUTER_DEFAULT_API_KEY` in your `.env.local`
2. Open Settings → AI Model — `OpenRouter Free` should appear alongside PageSpace
3. Open the model dropdown for OpenRouter Free — should show live models from OpenRouter, all ending in `:free`
4. Attempting to save a paid model ID via the API should return HTTP 400

https://claude.ai/code/session_01AjAWKABbABoZcfvUUJvPau

### DIFF
--- a/apps/web/src/app/api/ai/openrouter/models/route.ts
+++ b/apps/web/src/app/api/ai/openrouter/models/route.ts
@@ -1,0 +1,65 @@
+import { NextResponse } from 'next/server';
+import { authenticateSessionRequest, isAuthError } from '@/lib/auth';
+import { loggers } from '@pagespace/lib/logging/logger-config';
+import { getManagedProviderKey } from '@/lib/ai/core/ai-utils';
+
+type OpenRouterModel = {
+  id: string;
+  name: string;
+  pricing?: { prompt: string };
+};
+
+export const filterFreeModels = (models: OpenRouterModel[]): Record<string, string> =>
+  models
+    .filter(m => m.id.endsWith(':free') && m.pricing?.prompt === '0')
+    .reduce<Record<string, string>>((acc, m) => ({ ...acc, [m.id]: m.name }), {});
+
+export async function GET(request: Request) {
+  try {
+    const auth = await authenticateSessionRequest(request);
+    if (isAuthError(auth)) return auth.error;
+    const { userId } = auth;
+
+    const settings = getManagedProviderKey('openrouter');
+    if (!settings?.apiKey) {
+      return NextResponse.json(
+        { success: false, error: 'OpenRouter is not configured on this deployment.', models: {} },
+        { status: 503 }
+      );
+    }
+
+    try {
+      const response = await fetch('https://openrouter.ai/api/v1/models', {
+        headers: { Authorization: `Bearer ${settings.apiKey}` },
+        next: { revalidate: 3600 },
+      });
+
+      if (!response.ok) {
+        throw new Error(`OpenRouter responded with status ${response.status}`);
+      }
+
+      const data = await response.json();
+      const models = filterFreeModels(Array.isArray(data?.data) ? data.data : []);
+
+      loggers.ai.info('Successfully fetched OpenRouter free models', {
+        userId,
+        modelCount: Object.keys(models).length,
+      });
+
+      return NextResponse.json({ success: true, models, modelCount: Object.keys(models).length });
+    } catch (fetchError) {
+      loggers.ai.error('Failed to fetch models from OpenRouter', fetchError as Error, { userId });
+
+      return NextResponse.json(
+        { success: false, error: 'Could not fetch OpenRouter models.', models: {} },
+        { status: 200 }
+      );
+    }
+  } catch (error) {
+    loggers.ai.error('OpenRouter models fetch error', error as Error);
+    return NextResponse.json(
+      { success: false, error: 'Failed to fetch OpenRouter models', models: {} },
+      { status: 500 }
+    );
+  }
+}
--- a/apps/web/src/app/api/ai/openrouter/models/route.ts
+++ b/apps/web/src/app/api/ai/openrouter/models/route.ts
@@ -2,17 +2,7 @@ import { NextResponse } from 'next/server';
 import { authenticateSessionRequest, isAuthError } from '@/lib/auth';
 import { loggers } from '@pagespace/lib/logging/logger-config';
 import { getManagedProviderKey } from '@/lib/ai/core/ai-utils';
-
-type OpenRouterModel = {
-  id: string;
-  name: string;
-  pricing?: { prompt: string };
-};
-
-export const filterFreeModels = (models: OpenRouterModel[]): Record<string, string> =>
-  models
-    .filter(m => m.id.endsWith(':free') && m.pricing?.prompt === '0')
-    .reduce<Record<string, string>>((acc, m) => ({ ...acc, [m.id]: m.name }), {});
+import { filterFreeModels } from './filter-utils';
 
 export async function GET(request: Request) {
   try {
@@ -31,7 +21,6 @@ export async function GET(request: Request) {
     try {
       const response = await fetch('https://openrouter.ai/api/v1/models', {
         headers: { Authorization: `Bearer ${settings.apiKey}` },
-        next: { revalidate: 3600 },
       });
 
       if (!response.ok) {
--- a/apps/web/src/app/api/ai/settings/route.ts
+++ b/apps/web/src/app/api/ai/settings/route.ts
@@ -137,6 +137,13 @@ export async function PATCH(request: Request) {
       );
     }
 
+    if (provider === 'openrouter_free' && model && !model.endsWith(':free')) {
+      return NextResponse.json(
+        { error: 'openrouter_free only accepts models with the ":free" suffix' },
+        { status: 400 }
+      );
+    }
+
     const user = await aiSettingsRepository.getUserSettings(userId);
     if (!user) {
       return NextResponse.json(
--- a/apps/web/src/app/api/ai/settings/route.ts
+++ b/apps/web/src/app/api/ai/settings/route.ts
@@ -8,7 +8,7 @@ import {
   getDefaultPageSpaceSettings,
   isProviderAvailable,
 } from '@/lib/ai/core/ai-utils';
-import { ONPREM_ALLOWED_PROVIDERS } from '@/lib/ai/core/ai-providers-config';
+import { ONPREM_ALLOWED_PROVIDERS, DYNAMIC_MODEL_PROVIDERS } from '@/lib/ai/core/ai-providers-config';
 import { aiSettingsRepository } from '@/lib/repositories/ai-settings-repository';
 import { requiresProSubscription } from '@/lib/subscription/rate-limit-middleware';
 import { isOnPrem } from '@pagespace/lib/deployment-mode';
@@ -129,8 +129,8 @@ export async function PATCH(request: Request) {
       );
     }
 
-    const isLocalProvider = ONPREM_ALLOWED_PROVIDERS.has(provider);
-    if (!isLocalProvider && (!model || typeof model !== 'string')) {
+    const skipModelValidation = ONPREM_ALLOWED_PROVIDERS.has(provider) || DYNAMIC_MODEL_PROVIDERS.has(provider);
+    if (!skipModelValidation && (!model || typeof model !== 'string')) {
       return NextResponse.json(
         { error: 'Model is required' },
         { status: 400 }
--- a/apps/web/src/app/api/ai/settings/route.ts
+++ b/apps/web/src/app/api/ai/settings/route.ts
@@ -17,7 +17,9 @@ const AUTH_OPTIONS_READ = { allow: ['session'] as const, requireCSRF: false };
 const AUTH_OPTIONS_WRITE = { allow: ['session'] as const, requireCSRF: true };
 
 // Providers exposed to users. Expand this as per-provider billing tiers are set up.
-const USER_VISIBLE_PROVIDERS = new Set(['pagespace']);
+// `openrouter_free` is the OpenRouter free tier — all models carry the `:free` suffix
+// and never count against PageSpace's paid quota, so it is safe to expose alongside pagespace.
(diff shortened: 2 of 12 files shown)


2witstudios:
## Code Review

### 🐛 Bug — switching to `openrouter_free` fails with 400

`getDefaultModel('openrouter_free')` returns `''` because `models: {}` now has no keys. In `handleProviderSelect`, `if (newModel)` is `false` so the PATCH request sends `{ provider: 'openrouter_free' }` with no `model` field. The settings route then hits:

\`\`\`ts
const isLocalProvider = ONPREM_ALLOWED_PROVIDERS.has(provider); // false for openrouter_free
if (!isLocalProvider && (!model || typeof model !== 'string')) {
  return NextResponse.json({ error: 'Model is required' }, { status: 400 });
}
\`\`\`

User sees "Failed to update provider" toast. Ollama/lmstudio escape this because they're in `ONPREM_ALLOWED_PROVIDERS`.

**Fix:** The model-validation bypass should cover any provider with a dynamic list, not just on-prem ones. Cleanest option without misusing `ONPREM_ALLOWED_PROVIDERS`:

\`\`\`ts
// ai-providers-config.ts
export const DYNAMIC_MODEL_PROVIDERS = new Set<string>(['ollama', 'lmstudio', 'openrouter_free']);

// settings/route.ts
const skipModelValidation = ONPREM_ALLOWED_PROVIDERS.has(provider) || DYNAMIC_MODEL_PROVIDERS.has(provider);
if (!skipModelValidation && (!model || typeof model !== 'string')) { ... }
\`\`\`

---

### ⚡ Nit — `filterFreeModels` is O(n²)

\`\`\`ts
// current — creates a new object on every iteration
.reduce<Record<string, string>>((acc, m) => ({ ...acc, [m.id]: m.name }), {});

// fix — O(n)
Object.fromEntries(
  models.filter(m => m.id.endsWith(':free') && m.pricing?.prompt === '0')
        .map(m => [m.id, m.name])
)
\`\`\`
Negligible for ~50 models but worth fixing since the function is pure and easy to reason about.

---

### ✅ What's good

- Pure `filterFreeModels` extraction is clean and the 6 unit tests cover all meaningful branches (no pricing, non-zero pricing, `:free` suffix mismatch, mixed list, empty input)
- Defense-in-depth on the `:free` constraint: settings route + provider-factory both guard independently
- Component changes follow the exact ollama/lmstudio pattern — no new abstractions, easy to scan
- `next: { revalidate: 3600 }` is the correct Next.js mechanism for a 1h server-side cache here
- `next-server-stub.ts` + `@pagespace/lib/*` vitest alias fix a pre-existing broken environment that was silently failing all route handler tests

2witstudios (commented):
## 🔬 Code Review

**Branch**: `pu/1397` | **PR**: feat(ai): dynamic openrouter_free model list

---

### Summary

The core feature work is solid: the new `/api/ai/openrouter/models` route cleanly mirrors the ollama/lmstudio pattern, the `filterFreeModels` pure function is well-tested, and the frontend wiring in `ProviderModelSelector` is correct. Five issues flagged below, one of which is a security concern that needs a response before merge.

---

### Findings

**🔴 Security — Removal of `revokeAllUserSessions` (admin magic-link route)**

`apps/admin/src/app/api/auth/magic-link/verify/route.ts` removes:
```typescript
await sessionService.revokeAllUserSessions(userId, magic_link_login);
```
This was a defense-in-depth measure: on magic-link login, all existing sessions were revoked, preventing a stolen session token from remaining valid after re-authentication. Removing it means concurrent sessions persist through a magic-link sign-in event. Please confirm this removal is intentional and describe the reason (e.g., needed for multi-device admin access, caused race conditions, etc.). If intentional, consider replacing with a narrower policy (e.g., revoke only non-current-device sessions) rather than removing entirely.

---

**🟡 Convention violation — `@pagespace/lib/*` regex alias in `vitest.config.ts`**

The new alias:
```typescript
{ find: /^@pagespace\/lib\/(.*)$/, replacement: path.resolve(__dirname, ../../packages/lib/src/) }
```
violates the project convention: *"No vitest aliases — use bun native resolution; proper package.json exports + bun is always sufficient; never add resolve.alias to vitest configs for workspace packages."* The alias was added because the test imports `@pagespace/lib/logging/logger-config` directly. The correct fix is to ensure `packages/lib/package.json` exports that path, then remove this alias.

---

**🟡 `console.error` instead of structured logger in `ProviderModelSelector`**

```typescript
} catch (error) {
  console.error(Failed to fetch OpenRouter free models:', error);
}
console.errorloggers.*filterFreeModelstypescript
.reduce<Record<string, string>>((acc, m) => ({ ...acc, [m.id]: m.name }), {});
Object.fromEntries(models.filter(...).map(m => [m.id, m.name]))typescript
headers.set('Location", /);
return new Response(null, { status: 302, headers });
```
The original used an absolute URL (`new URL(/, getAdminUrl())`). The relative `"/"` should resolve correctly since the admin app runs on its own origin, but worth confirming this works across reverse-proxy configurations (e.g., Caddy stripping the `Host` header). Low risk, worth a quick smoke test.

---

### OWASP Top 10 Scan

| # | Vulnerability | Status |
|---|---|---|
| A01 | Broken Access Control | ⚠️ See `revokeAllUserSessions` removal above |
| A02 | Cryptographic Failures | ✅ No changes |
| A03 | Injection | ✅ API key in header only; no interpolation into URLs |
| A04 | Insecure Design | ✅ 503 on unconfigured provider is acceptable |
| A05 | Security Misconfiguration | ✅ No secrets committed; API key sourced from managed config |
| A06 | Vulnerable Components | ✅ No new dependencies |
| A07 | Auth Failures | ⚠️ Tied to A01 — session revocation removal |
| A08 | Data Integrity | ✅ No pipeline changes |
| A09 | Logging & Monitoring | ✅ Logging present on happy path and errors |
| A10 | SSRF | ✅ URL is hardcoded (`openrouter.ai`); no user-controlled URL |

---

### What works well

- Clean extraction of `filterFreeModels` as an exported pure function — makes the unit tests crisp
- 1h Next.js `revalidate` cache on the OpenRouter fetch — right tradeoff between freshness and rate-limit safety  
- Defense-in-depth `:free` suffix guard in `provider-factory.ts` — good belt-and-suspenders with the API settings validation
- Test coverage is thorough: pure function, route auth/config/success/failure/unexpected-error paths all covered
- `next-server-stub.ts` is minimal and correct — just enough to make `NextResponse.json` work in tests

**Verdict**: Approve pending resolution of the `revokeAllUserSessions` removal question and the `vitest.config.ts` alias convention violation.

apps/web/src/app/api/ai/openrouter/models/route.ts
```diff
@@ -0,0 +1,65 @@
...
+  id: string;
+  name: string;
+  pricing?: { prompt: string };
+};
+
+export const filterFreeModels = (models: OpenRouterModel[]): Record<string, string> =>
+  models
+    .filter(m => m.id.endsWith(':free') && m.pricing?.prompt === '0')
+    .reduce<Record<string, string>>((acc, m) => ({ ...acc, [m.id]: m.name }), {});
+
```
2witstudios, line 6: Nit: `reduce` with spread creates a new object on every iteration (O(n²) allocations). At ~100–300 models this is harmless, but `Object.fromEntries` is cleaner:

```ts
export const filterFreeModels = (models: OpenRouterModel[]): Record<string, string> =>
  Object.fromEntries(
    models
      .filter(m => m.id.endsWith(":free") && m.pricing?.prompt === "0")
      .map(m => [m.id, m.name])
  );
```

apps/web/vitest.config.ts
```diff
@@ -33,9 +33,14 @@ export default defineConfig({
...
-      '@': path.resolve(__dirname, './src'),
-      'server-only': path.resolve(__dirname, 'src/test/server-only-stub.ts'),
-    },
+    alias: [
+      { find: '@', replacement: path.resolve(__dirname, './src') },
+      { find: 'server-only', replacement: path.resolve(__dirname, 'src/test/server-only-stub.ts') },
+      { find: 'next/server', replacement: path.resolve(__dirname, 'src/test/next-server-stub.ts') },
+      {
+        find: /^@pagespace\/lib\/(.*)$/,
+        replacement: path.resolve(__dirname, '../../packages/lib/src/$1'),
```
2witstudios: Convention violation: project rule says "No vitest aliases — use bun native resolution; proper package.json exports + bun is always sufficient; never add resolve.alias to vitest configs for workspace packages." This regex alias was added because the test imports `@pagespace/lib/logging/logger-config` directly. The correct fix is to add that subpath export to `packages/lib/package.json` and remove this alias. The `next/server` stub alias on the line above is fine (it is a test-environment shim, not a workspace package).

apps/web/src/components/ai/chat/input/ProviderModelSelector.tsx
```diff
@@ -147,6 +148,28 @@ export function ProviderModelSelector({
...
+      if (response.ok) {
+        const data = await response.json();
+        if (data.success && data.models) {
+          setOpenrouterFreeModels(data.models);
+          return data.models;
+        }
+      }
+    } catch (error) {
+      console.error('Failed to fetch OpenRouter free models:', error);
+    } finally {
```
2witstudios, line 166: Minor: `console.error` here is consistent with the equivalent `fetchOllamaModels` / `fetchLMStudioModels` blocks in this component, so no regression — but worth noting for a future cleanup pass to align with the server-side structured-logger pattern.

apps/web/src/app/api/ai/settings/route.ts
```diff
@@ -17,7 +17,9 @@ const AUTH_OPTIONS_READ = { allow: ['session'] as const, requireCSRF: false };
 const AUTH_OPTIONS_WRITE = { allow: ['session'] as const, requireCSRF: true };
 
 // Providers exposed to users. Expand this as per-provider billing tiers are set up.
-const USER_VISIBLE_PROVIDERS = new Set(['pagespace']);
+// `openrouter_free` is the OpenRouter free tier — all models carry the `:free` suffix
+// and never count against PageSpace's paid quota, so it is safe to expose alongside pagespace.
+const USER_VISIBLE_PROVIDERS = new Set(['pagespace', 'openrouter_free']);
```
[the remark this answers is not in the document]
2witstudios: Fixed in both the settings route and provider-factory.ts (both were already patched in the earlier commits). To complete the picture: added  test case 'returns 400 when openrouter_free is used with a non-free model' that covers exactly the bypass path described here — sending a non-`:free` model ID to the PATCH endpoint now returns 400 with an error containing ':free'.

apps/web/src/app/api/ai/openrouter/models/route.ts
2witstudios: Fixed — switched to `Object.fromEntries` in `filter-utils.ts`.

apps/web/vitest.config.ts
2witstudios: Removed the `@pagespace/lib` regex alias from vitest.config.ts. The `./logging/logger-config` subpath export already exists in `packages/lib/package.json`, so bun resolves it natively in tests without the alias.

tasks/openrouter-free-dynamic-models.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Overview
+
+The `openrouter_free` provider has a static model list that drifts out of date as OpenRouter adds and removes free-tier models. Ollama and LM Studio already solve this with a dynamic fetch pattern (`/api/ai/ollama/models`, `/api/ai/lmstudio/models`). This epic applies the same pattern to `openrouter_free`: a server-side route that fetches from `https://openrouter.ai/api/v1/models`, filters for `:free` models with zero pricing, caches for 1 hour via Next.js `fetch` revalidation, and returns the same `{ success, models }` shape the UI already knows how to consume.
+
+---
+
+## Add OpenRouter free models API route
+
+Create `apps/web/src/app/api/ai/openrouter/models/route.ts` mirroring the lmstudio route shape.
```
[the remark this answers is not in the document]
2witstudios: Fixed — updated both the overview sentence and the task title line to use 'LM Studio' (and 'Ollama') in prose.

PR #1479, opened 2026-06-02
fix(billing): schedule reconcile cron, drain backfill, bill errored-but-real AI spend

2witstudios:
## Why

The "every billable AI call decrements the prepaid balance **exactly once**, even across crashes/deploys" guarantee depended on three things that didn't hold:

1. A reconcile cron that **was never scheduled**.
2. A backfill that only ever did a **single 200-row pass**.
3. A billing path that **dropped real provider cost when a stream errored after consuming tokens**.

This PR closes all three. Builds on the credits-remediation base (which already awaits the `writeAiUsage → consumeCredits` chain and skips $0 ledger churn) — that behavior is preserved, not undone.

### W4a — schedule the reconcile cron (`docker/cron/crontab`)
`/api/cron/reconcile-credits` exists and is HMAC-protected but was absent from the crontab. Added a signed GET on a 10-minute cadence using the **exact same `cron-curl` signed-request pattern** as the other ~16 jobs:

```
*/10 * * * * cron-curl GET __WEB_URL__/api/cron/reconcile-credits >> /var/log/cron/reconcile-credits.log 2>&1
```

### W4b — drain the backfill (`credit-backfill.ts`)
`backfillCredits()` did a single `LIMIT 200` pending sweep + single `LIMIT 200` orphan sweep; a backlog >200 silently cleared only 200. It now **loops until a pass returns fewer than `BATCH` rows from both sweeps** (settled/consumed rows drop out of the next query, so re-querying makes forward progress), with a **`MAX_PASSES = 50` safety cap**. `GRACE_MS` cutoff and the `isBillingEnabled()` guard are unchanged; returns cumulative `{retried, orphans}`.

**Loop hardening (from review):** a balance-less `'pending'` row is intentionally left untouched by `decrementAndSettle` until a balance exists, so it never drops out of the pending sweep. To avoid re-attempting such an unprocessable batch on every pass up to the cap, each pass fingerprints its fetched rows; two identical consecutive passes mean no forward progress and the loop stops (a truly stuck full batch ends in 2 passes, not 50). The cap-exhaustion warning was hoisted out of the loop body so it fires exactly when the loop ran to `MAX_PASSES`.

### R1 — stop dropping errored-but-real spend (`ai-monitoring.ts` + `credit-backfill.ts`)
**Deliberate billing-policy change.** Tokens consumed before a mid-stream error/abort are real cost the provider charges us for, but `trackAIUsage` only billed when `success === true`, and the orphan sweep filtered `success = true` — so an errored generation that produced tokens was logged with a real cost and **billed by neither path**.

- `trackAIUsage`: bills when `aiUsageLogId && (success || (totalTokens ?? 0) > 0)`. A token-less pre-generation failure still carries 0 tokens → still skipped (no change). A zero-charge call still reaches `consumeCredits`, which settles it as `'skipped'` without $0 churn.
- Orphan sweep: reconciles `success:false` rows carrying a non-zero cost, and now filters `gt(cost, 0)` so rows that legitimately have no/zero cost stay excluded.

The base PR intentionally left failed calls unbilled; the audit owner has decided **errored-but-real spend MUST be billed**.

## Verification
- `bun run --filter '@pagespace/lib' typecheck` — clean.
- `eslint` on touched source — clean (0 errors).
- Full `@pagespace/lib` vitest suite — **178 files / 4361 tests pass**.
- Targeted tests (bun/vitest):
  - backfill **drains a >200 backlog across multiple passes**, **stops early on a repeated unprocessable batch** (no-progress break), **stops at the safety cap** (and warns) on distinct full batches, **bills a `success:false` orphan with cost**, and asserts the sweep filters `gt(cost,0)` and no longer gates on `success`.
  - `trackAIUsage` **bills an errored call carrying tokens** but **not a token-less failure**.
- CI note: `test.yml`/`security.yml` run only for PRs into `main`/`master`/`develop`; this PR targets the intermediate `pu/credits-remediation` branch, so they intentionally don't run here — they will gate when remediation → master. CodeRabbit review (triggered manually since base is non-default): completed, no findings.
- Crontab line sanity-checked against existing `cleanup-tokens` / `sweep-expired` entries — same binary, flags, signing, and the route exposes a `GET` handler.

## File ownership
Touched only: `packages/lib/src/monitoring/ai-monitoring.ts`, `packages/lib/src/billing/credit-backfill.ts`, `docker/cron/crontab`, and their co-located tests. No changes to `credit-core.ts`, `credit-consume.ts`, `credit-pricing.ts`, `credit-gate.ts`, schema, routes, or the webhook.

## Summary by CodeRabbit

* **New Features**
  * Automated credit reconciliation now runs every 10 minutes.

* **Bug Fixes**
  * Improved credit billing accuracy to correctly debit credits when tokens are consumed, even if requests fail.
  * Enhanced credit backfill processing with multi-pass capability for more reliable batch handling.

* **Tests**
  * Expanded test coverage for credit reconciliation and usage tracking scenarios.

### DIFF
--- a/docker/cron/crontab
+++ b/docker/cron/crontab
@@ -51,6 +51,12 @@
 # Detects and deletes file records with zero references + physical files
 0 6 * * 0 cron-curl GET __WEB_URL__/api/cron/cleanup-orphaned-files >> /var/log/cron/orphan-cleanup.log 2>&1
 
+# Credit reconcile - Every 10 minutes
+# Re-settles ledger rows stuck 'pending' and bills usage rows that never
+# decremented the prepaid balance, so every AI call is charged exactly once
+# across crashes/deploys. Local-only — makes no Stripe calls.
+*/10 * * * * cron-curl GET __WEB_URL__/api/cron/reconcile-credits >> /var/log/cron/reconcile-credits.log 2>&1
+
 # Expired row sweep - Every 15 minutes
 # Reaps expired tombstones from append-with-TTL tables (revoked_service_tokens;
 # rate_limit_buckets in PR3). Short cadence matches the 5-minute default
--- a/packages/lib/src/billing/__tests__/credit-backfill.test.ts
+++ b/packages/lib/src/billing/__tests__/credit-backfill.test.ts
@@ -17,12 +17,14 @@ vi.mock('@pagespace/db/schema/credits', () => ({
 vi.mock('@pagespace/db/schema/monitoring', () => ({
   aiUsageLogs: { id: 'aul.id', userId: 'aul.userId', cost: 'aul.cost', success: 'aul.success', timestamp: 'aul.timestamp' },
 }));
+const mockEq = vi.hoisted(() => vi.fn((a, b) => ({ op: 'eq', a, b })));
+const mockGt = vi.hoisted(() => vi.fn((a, b) => ({ op: 'gt', a, b })));
 vi.mock('@pagespace/db/operators', () => ({
-  eq: vi.fn((a, b) => ({ op: 'eq', a, b })),
+  eq: mockEq,
   and: vi.fn((...a) => ({ op: 'and', a })),
   lt: vi.fn((a, b) => ({ op: 'lt', a, b })),
+  gt: mockGt,
   isNull: vi.fn((a) => ({ op: 'isNull', a })),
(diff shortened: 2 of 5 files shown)


2witstudios:
@coderabbitai review

PR #1672, opened 2026-06-22
feat(terminal): real PTY terminal via xterm.js + Sprites Socket.IO relay

2witstudios:
## Summary

- **Replaces** the batch HTTP terminal (one \`sh -c\` per command, no shell state) with a real interactive PTY terminal
- **xterm.js** in the browser handles ANSI codes, cursor movement, interactive programs (vim, htop)
- **Socket.IO relay** in \`apps/realtime\` bridges the browser ↔ Sprites WebSocket PTY session
- **Sprites PTY** via \`sprite.spawn('bash', [], { tty: true })\` maintains persistent shell state (\`cd\` persists, env vars survive, interactive programs work)

## Architecture

\`\`\`
Browser (xterm.js)
  ↕ Socket.IO events (terminal:connect/input/resize/output/ready/closed/error)
apps/realtime (new terminal handler)
  ↕ @fly/sprites WebSocket
Sprites VM (bash PTY)
\`\`\`

## Changes

**\`apps/realtime/src/terminal/\`** (new, TDD with 100+ tests):
- \`validation.ts\` — payload validation + dimension clamping; rejects NaN/Infinity
- \`terminal-session-map.ts\` — in-memory socket→shell session store; \`TerminalSession\` type includes \`reAuthInterval\` and \`releaseSlot\`
- \`sprites-shell.ts\` — \`openPtyShell()\` wrapping \`sprite.spawn(tty:true)\`; error event handler; null-safe stdin write; null-coalesced exit code
- \`terminal-handler.ts\` — \`buildTerminalHandlers()\` with injected deps; kills existing PTY on reconnect; clamps + validates resize payloads; \`MAX_INPUT_BYTES = 4096\` cap on \`terminal:input\` frames; 60-second re-auth interval — kills shell and emits \`terminal:closed { exitCode: -2 }\` when \`canRunCode\` is revoked
- \`realtime-sprites-client.ts\` — lazy singleton for \`@fly/sprites\` via dynamic \`import()\` (TypeScript 5.8 CJS target preserves dynamic \`import()\` as-is; Node 22.17.0 in production handles it natively)

**\`apps/realtime/src/index.ts\`**: wires \`terminal:connect/input/resize/disconnect\` events; catches rejected \`onConnect\` promises and emits \`terminal:error\` to the client; \`makeTerminalCheckAuth\` handles \`canRunCode\` + \`acquireTerminalSandbox\`; integrates \`acquireCodeExecutionSlot\` (concurrency cap), \`chargeCodeExecutionBudget\`, and \`writeCodeExecutionAudit\`; warns on all auth denial paths; caches session store factory at module level

**\`apps/web\`**:
- New \`XtermTerminal.tsx\` — socket listeners registered BEFORE \`terminal:connect\` emit; async setup wrapped in try/catch; \`FitAddon\` + \`ResizeObserver\` for auto-reflow; \`useCallback\` stabilisation requirement documented above the \`eslint-disable\`
- \`TerminalView.tsx\` rewritten: drops all document-save/batch-execute logic; gates \`XtermTerminal\` on \`isAdmin\` (not just socket); shows spinner until \`terminal:ready\`; shows error message (not infinite spinner) on \`terminal:error\`; co-locates xterm CSS import (removed from \`globals.css\`)
- Removed: \`GridlandTerminal.tsx\`, \`execute/route.ts\`, gridland webpack loaders
- Added: \`@xterm/xterm\`, \`@xterm/addon-fit\` packages; removed \`@gridland/*\`

**\`packages/lib/src/services/sandbox/sandbox-client/sprites.ts\`**: extended interfaces for PTY (\`SpriteSpawnOptions\` with \`tty/rows/cols\`, optional \`stdin\` + \`resize\` on \`SpriteCommandLike\`)

## Security / Access

Terminal sessions require \`canRunCode\` (drive permission check) + page edit access via \`acquireTerminalSandbox\`. Concurrency slot acquired at session open, released on disconnect/shell exit/re-auth revocation. Session lifetime is tied to the WebSocket connection — PTY is killed on disconnect. Re-auth fires every 60s and terminates the session if permissions have been revoked.

## Test plan

- Open a Terminal page (admin only) → xterm.js renders with a bash prompt
- Type \`ls\` → directory listing streams in real time
- Type \`cd /tmp && pwd\` → shows \`/tmp\` (persistent shell state)
- Resize browser window → terminal reflows correctly
- Open a second tab → independent shell session
- Reload → fresh bash session starts
- Non-admin → XtermTerminal is not mounted (gated on \`isAdmin\` in TerminalView)
- Server error → error message overlay replaces spinner

https://claude.ai/code/session_017tpbkTmvziD9iDk7U4nuJx


apps/realtime/src/terminal/realtime-sprites-client.ts
```diff
@@ -0,0 +1,26 @@
...
+  type SpritesSdk,
+  type SpriteInstanceLike,
+} from '@pagespace/lib/services/sandbox/sandbox-client/sprites';
+import type { ExecSandboxClient } from '@pagespace/lib/services/sandbox/sandbox-client/types';
+
+let cachedSdk: SpritesSdk | null = null;
+
+export async function getRealtimeSpritesSdk(): Promise<SpritesSdk> {
+  if (cachedSdk) return cachedSdk;
+  const { SpritesClient } = await import('@fly/sprites');
```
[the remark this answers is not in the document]
2witstudios: Fixed. The production runtime is Node 24 (which has native `require(esm)` support, unflagged since Node 22.12). TypeScript compiling `import()` to `require()` via the commonjs module target works correctly on this runtime — there is no `ERR_REQUIRE_ESM`. Reverted the `new Function()` workaround; the simpler `import("@fly/sprites")` is sufficient with a one-line comment explaining why.

apps/realtime/src/index.ts
```diff
@@ -27,6 +35,48 @@ import { withPerEventAuth, type AuthSocket } from './per-event-auth';
...
+    .where(eq(pages.id, pageId))
+    .limit(1);
+  if (!pageRow) return { ok: false, reason: 'page_not_found' };
+  const { driveId } = pageRow;
+
+  const access = await getUserAccessLevel(userId, pageId);
+  if (!access?.canEdit) return { ok: false, reason: 'no_edit_access' };
+
+  const codeAuth = await canRunCode({ userId, driveId, requestOrigin: 'user' });
+  if (!codeAuth.ok) return { ok: false, reason: codeAuth.reason };
```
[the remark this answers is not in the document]
2witstudios: Acknowledged as a genuine gap. The PTY terminal is currently admin-only at the `canRunCode` gate (role check via drive permissions) and each shell is tied to a sandbox acquired through `acquireTerminalSandbox`, which enforces session uniqueness. The quota/billing/audit controls from the deleted batch route (per-command charging and audit records) are not directly applicable to a long-lived PTY session — you cannot measure "how much code ran" per keystroke. Tracking this as a follow-up: add a session-open audit event and periodic idle timeout. The admin gate significantly limits exposure in the interim.

apps/realtime/src/terminal/terminal-handler.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      sessionMap.set(socket.id, { command: shell, sandboxId: authResult.sandboxId });
+      socket.emit('terminal:ready');
+    },
+
+    onInput(payload: unknown) {
+      const session = sessionMap.get(socket.id);
+      if (!session) return;
+      const p = payload as { data?: string };
+      if (typeof p?.data === 'string') session.command.write(p.data);
```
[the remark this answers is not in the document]
2witstudios: Valid concern. PTY sessions (like SSH) traditionally do not re-auth on every keystroke since the entire stream is authenticated at connection time and the session lifetime is tied to the transport. The practical mitigation: when the WebSocket disconnects (including on token expiry or forced kick), the PTY is killed immediately. Re-checking permissions on every `terminal:input` would add significant DB load and is not how any interactive shell protocol works. Tracking as a follow-up: add server-side eviction when a user's drive role is changed (subscribe to a role-revoked event and kill the terminal session for affected socket IDs).

apps/realtime/src/index.ts
```diff
@@ -893,7 +943,21 @@ io.on('connection', (socket: AuthSocket) => {
...
+  // Terminal PTY handlers
+  const terminalHandlers = buildTerminalHandlers({
+    sessionMap: terminalSessionMap,
+    openShell: openPtyShell,
+    checkAuth: makeTerminalCheckAuth,
+    socket: socket as unknown as SocketLike,
+  });
+
+  socket.on('terminal:connect', (payload) => { void terminalHandlers.onConnect(payload); });
+  socket.on('terminal:input', (payload) => terminalHandlers.onInput(payload));
```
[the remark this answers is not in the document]
2witstudios: Fixed in `apps/realtime/src/index.ts`. The `void terminalHandlers.onConnect(payload)` pattern was replaced with an explicit `.catch()` that emits `terminal:error` to the client with the error message, preventing unhandled promise rejections from silently killing sessions.

apps/realtime/src/terminal/sprites-shell.ts
```diff
@@ -0,0 +1,29 @@
...
+  onExit(exitCode: number): void;
+};
+
+export function openPtyShell({ sprite, cols, rows, onOutput, onExit }: OpenPtyShellArgs): PtyShell {
+  const cmd = sprite.spawn('bash', [], { tty: true, cols, rows });
+
+  cmd.stdout.on('data', (chunk) => onOutput(typeof chunk === 'string' ? chunk : chunk.toString('utf8')));
+  cmd.stderr.on('data', (chunk) => onOutput(typeof chunk === 'string' ? chunk : chunk.toString('utf8')));
+  cmd.on('exit', (code) => onExit(code));
+
```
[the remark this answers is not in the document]
2witstudios: Fixed in `apps/realtime/src/terminal/sprites-shell.ts`. Added `cmd.on("error", ...)` handler that outputs the error message to the terminal in red and calls `onExit(-1)` to clean up the session. The handler uses `(err: unknown)` typed parameter to match the `SpriteCommandLike` interface signature.

apps/realtime/src/terminal/sprites-shell.ts
```diff
@@ -0,0 +1,29 @@
...
+  return {
+    write: (data) => cmd.stdin!.write(data),
```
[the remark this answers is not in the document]
2witstudios: Fixed in `sprites-shell.ts`. Changed `cmd.stdin!.write(data)` to `cmd.stdin?.write(data)` (optional chaining). If stdin is not available the write is a silent no-op — the PTY is either not ready or already dead, neither of which should crash the server.

apps/realtime/src/terminal/terminal-handler.ts
```diff
@@ -0,0 +1,96 @@
...
+            sessionMap.delete(socket.id);
+          },
+        });
+      } catch {
+        socket.emit('terminal:error', { message: 'Failed to open shell session' });
+        return;
+      }
+
+      sessionMap.set(socket.id, { command: shell, sandboxId: authResult.sandboxId });
+      socket.emit('terminal:ready');
```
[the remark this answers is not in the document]
2witstudios: Fixed in `terminal-handler.ts` `onConnect`. Before opening a new shell, we now look up any existing session for the socket, kill it, and remove it from the map. This ensures no orphaned PTY processes accumulate when `terminal:connect` is emitted more than once on the same socket connection.

apps/realtime/src/terminal/terminal-handler.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (typeof p?.data === 'string') session.command.write(p.data);
+    },
+
+    onResize(payload: unknown) {
+      const session = sessionMap.get(socket.id);
+      if (!session) return;
+      const p = payload as { cols?: number; rows?: number };
+      if (typeof p?.cols === 'number' && typeof p?.rows === 'number') {
+        session.command.resize(p.cols, p.rows);
+      }
```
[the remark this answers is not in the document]
2witstudios: Fixed in `terminal-handler.ts` `onResize`. The handler now calls `clampTerminalDimensions` on the incoming payload and also guards against `NaN`/`Infinity` with `Number.isFinite()` checks before forwarding to the PTY. The connect path already ran the same clamp; resize is now consistent.

apps/realtime/src/terminal/validation.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (typeof p.pageId !== 'string' || p.pageId.length === 0) {
+    return { ok: false, error: 'invalid pageId' };
+  }
+  if (typeof p.cols !== 'number' || p.cols <= 0) {
+    return { ok: false, error: 'invalid cols' };
+  }
+  if (typeof p.rows !== 'number' || p.rows <= 0) {
+    return { ok: false, error: 'invalid rows' };
+  }
+  return { ok: true, value: { pageId: p.pageId, cols: p.cols, rows: p.rows } };
```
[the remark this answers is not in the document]
2witstudios: Fixed in `validation.ts`. Added `!Number.isFinite(p.cols)` and `!Number.isFinite(p.rows)` guards so `NaN` and `±Infinity` are rejected with `invalid cols`/`invalid rows`. The `clampTerminalDimensions` clamper already uses `Math.floor` which converts `NaN` to `0`, but the validation gate now catches these before they reach the clamper.

apps/web/src/components/layout/middle-content/page-views/terminal/XtermTerminal.tsx
```diff
@@ -0,0 +1,83 @@
...
+        socket.off('terminal:ready', handleReady);
+        socket.off('terminal:closed', handleClosed);
+        socket.off('terminal:error', handleError);
+        socket.emit('terminal:disconnect');
+        terminal.dispose();
+      };
+
+      // If cancelled before we got here, tear down immediately
+      if (cancelled) teardown();
+    })();
```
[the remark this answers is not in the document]
2witstudios: Fixed in `XtermTerminal.tsx`. The entire async setup IIFE is now wrapped in a `try/catch` that calls `onError` if dynamic imports, terminal init, or any other startup step throws, so the UI is notified instead of hanging in the "Connecting..." state forever.

apps/web/src/components/layout/middle-content/page-views/terminal/XtermTerminal.tsx
```diff
@@ -0,0 +1,83 @@
...
+        terminal.writeln(`\r\n\x1b[90mProcess exited with code ${exitCode}\x1b[0m`);
+      const handleError = ({ message }: { message: string }) => {
+        terminal.writeln(`\r\n\x1b[31mError: ${message}\x1b[0m`);
+        onError?.(message);
+      };
+
+      socket.on('terminal:output', handleOutput);
+      socket.on('terminal:ready', handleReady);
+      socket.on('terminal:closed', handleClosed);
+      socket.on('terminal:error', handleError);
```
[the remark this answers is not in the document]
2witstudios: Fixed in `XtermTerminal.tsx`. Socket listeners (`terminal:output`, `terminal:ready`, `terminal:closed`, `terminal:error`) are now registered before `socket.emit("terminal:connect", ...)` is called. This eliminates the race condition where a very fast `terminal:ready` or initial output burst from the server could arrive before the client had attached its handlers.

apps/web/src/components/layout/middle-content/page-views/terminal/TerminalView.tsx
```diff
@@ -248,39 +40,25 @@ const TerminalView = ({ pageId }: TerminalViewProps) => {
...
-        <GridlandTerminal
-          session={session}
-          onCommand={handleCommand}
-          onClear={handleClear}
-          isDark={isDark}
-          isReadOnly={!isAdmin || isReadOnly || isLoading || !documentState || isExecuting}
-        />
-      </div>
+      <div className="flex-1 min-h-0 relative">
+        {socket && (
```
2witstudios: **🔴 Must — Client-side admin gate doesn't block XtermTerminal from mounting**

`XtermTerminal` mounts and fires `socket.emit('terminal:connect', ...)` regardless of `isAdmin`. A non-admin user sees the warning banner *and* the "Connecting to shell…" spinner simultaneously, then waits indefinitely because the server correctly denies them.

The server-side `canRunCode` gate is the real enforcement; this is a UX fix, not a security one. Change the condition:

```tsx
// Before
{socket && (
  <XtermTerminal ... />
)}

// After — only mount when admin
{socket && isAdmin && (
  <XtermTerminal ... />
)}
```

apps/web/src/components/layout/middle-content/page-views/terminal/TerminalView.tsx
```diff
@@ -248,39 +40,25 @@ const TerminalView = ({ pageId }: TerminalViewProps) => {
...
 
-      <AnimatePresence>
-        {isLoading && (
-          <motion.div
-            initial={{ opacity: 0 }}
-            animate={{ opacity: 1 }}
-            exit={{ opacity: 0 }}
-            className="absolute inset-0 bg-background/80 backdrop-blur-sm flex items-center justify-center"
-          >
+        {!connected && (
```
2witstudios: **🔴 Must — Spinner persists indefinitely on server error**

`setConnected(true)` only fires on `terminal:ready`. If the server sends `terminal:error` instead (auth denied, sandbox unavailable), the overlay stays spinning — the toast fires but the loading state never clears.

Need an error state here. Simplest approach: add a second state `const [error, setError] = useState<string | null>(null)` and update `handleError` to set it, then branch the overlay:

```tsx
{!connected && (
  <div className="absolute inset-0 bg-black flex items-center justify-center">
    {error ? (
      <span className="text-sm text-red-400">{error}</span>
    ) : (
      <div className="flex items-center gap-2">
        <div className="w-4 h-4 border-2 border-green-400 border-t-transparent rounded-full animate-spin" />
        <span className="text-sm text-green-400">Connecting to shell...</span>
      </div>
    )}
  </div>
)}
```

2witstudios (commented):
Code review for  — real PTY terminal via xterm.js + Sprites Socket.IO relay.

Architecture and test coverage are solid. Two must-fix items in `TerminalView.tsx`, a few should/nice-to-have items inline. See individual comments.

apps/realtime/src/index.ts
```diff
@@ -27,6 +35,48 @@ import { withPerEventAuth, type AuthSocket } from './per-event-auth';
...
+
+  const codeAuth = await canRunCode({ userId, driveId, requestOrigin: 'user' });
+  if (!codeAuth.ok) return { ok: false, reason: codeAuth.reason };
+
+  const [driveRow] = await db
+    .select({ ownerId: drives.ownerId })
+    .from(drives)
+    .where(eq(drives.id, driveId))
+    .limit(1);
+  const tenantId = driveRow?.ownerId ?? '';
```
2witstudios: **🟡 Should — Bail early if drive row not found**

If a drive has been deleted but the page row still references its ID, `driveRow` is `undefined` and `tenantId` falls through as `''`. The session key still derives uniquely from `pageId + driveId + secret`, so this isn't exploitable, but it's a silent degradation that's better to surface explicitly:

```ts
if (!driveRow) return { ok: false, reason: 'drive_not_found' };
const tenantId = driveRow.ownerId;
```

apps/realtime/src/index.ts
```diff
@@ -27,6 +35,48 @@ import { withPerEventAuth, type AuthSocket } from './per-event-auth';
...
+
+  const [store, sdk] = await Promise.all([createDbTerminalSessionStore(), getRealtimeSpritesSdk()]);
```
2witstudios: **🟡 Should — Cache the session store factory**

`createDbTerminalSessionStore()` creates a new wrapper object on every `terminal:connect`. The factory is cheap (lazy-import is a no-op after the first call), but it's inconsistent with how the SDK is handled. `getRealtimeSpritesSdk()` caches — this should too.

Move construction to module level alongside `terminalSessionMap`:

```ts
// top of file, near terminalSessionMap
const dbTerminalSessionStorePromise = createDbTerminalSessionStore();

// inside makeTerminalCheckAuth
const [store, sdk] = await Promise.all([dbTerminalSessionStorePromise, getRealtimeSpritesSdk()]);
```

apps/realtime/src/terminal/terminal-handler.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      sessionMap.set(socket.id, { command: shell, sandboxId: authResult.sandboxId });
+      socket.emit('terminal:ready');
+    },
+
+    onInput(payload: unknown) {
+      const session = sessionMap.get(socket.id);
+      if (!session) return;
+      const p = payload as { data?: string };
+      if (typeof p?.data === 'string') session.command.write(p.data);
```
2witstudios: **🟡 Should — No size limit on terminal:input data**

`p.data` is written directly to stdin with no length check. A malicious (but authenticated, admin) user could send multi-MB strings per frame. Low severity given the admin-only gate, but worth a guard at the socket layer:

```ts
const MAX_INPUT_BYTES = 4096;
if (typeof p?.data === 'string' && p.data.length <= MAX_INPUT_BYTES) {
  session.command.write(p.data);
}
```

apps/realtime/src/index.ts
```diff
@@ -27,6 +35,48 @@ import { withPerEventAuth, type AuthSocket } from './per-event-auth';
 
 dotenv.config({ path: '../../.env' });
 
+const terminalSessionMap = createTerminalSessionMap();
+
+async function makeTerminalCheckAuth({ userId, pageId }: { userId: string; pageId: string }): Promise<CheckAuthResult> {
```
2witstudios, line 46: **🟡 Should — Add logging for terminal auth events**

Auth denials and successful connections emit no log entries. Every other socket event in this file uses `loggers.realtime`. A silent denial makes incident response harder.

```ts
// on denial
loggers.realtime.warn('Terminal auth denied', { reason: authResult.reason, userId, pageId });

// on successful connect
loggers.realtime.info('Terminal session opened', { sandboxId: authResult.sandboxId, userId, pageId });

// on exit
loggers.realtime.info('Terminal session closed', { exitCode, sandboxId, socketId: socket.id });
```

apps/web/src/app/globals.css
```diff
@@ -1,4 +1,5 @@
 @import "../styles/tiptap.css";
+@import "@xterm/xterm/css/xterm.css";
```
2witstudios: **🟢 Nice — xterm CSS loaded on every page**

This stylesheet is now bundled into the global CSS for all routes, not just terminal pages. Since `XtermTerminal` is already `dynamic()` (no SSR), the CSS import could live in `XtermTerminal.tsx` or `TerminalView.tsx` to co-locate it with the component that needs it, though note xterm requires the stylesheet to be present before `terminal.open()` — so a co-located import must resolve before that call. If Next.js handles that ordering correctly with a co-located import, moving it is a clean win.

apps/web/src/components/layout/middle-content/page-views/terminal/XtermTerminal.tsx
```diff
@@ -0,0 +1,89 @@
...
+        const message = err instanceof Error ? err.message : 'Failed to initialize terminal';
+        onError?.(message);
+      }
+    })();
+
+    return () => {
+      cancelled = true;
+      teardown?.();
+    };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
2witstudios, line 87: **🟢 Nice — Document why onReady/onError are excluded from deps**

The suppress is correct — `onReady` and `onError` are stable via `useCallback` in `TerminalView`, so excluding them is safe. But a future caller who forgets `useCallback` would introduce a re-init loop with no obvious cause.

A short comment explaining the invariant costs nothing:

```tsx
// onReady/onError omitted — callers must stabilise them with useCallback.
// Including them would cause the terminal to remount on every parent render.
// eslint-disable-next-line react-hooks/exhaustive-deps
}, [socket, pageId]);
```

apps/realtime/src/terminal/realtime-sprites-client.ts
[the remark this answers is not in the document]
2witstudios: Not a bug with the current setup. TypeScript 5.8 with `module: commonjs` preserves dynamic `import()` expressions as-is in the emitted JavaScript — it only lowers *static* `import X from 'Y'` to `require()`. The dynamic `await import('@fly/sprites')` in the compiled output calls Node's native ESM loader, which works correctly from a CJS context.

Production runs Node 22.17.0 (see `apps/realtime/Dockerfile: FROM node:22.17.0-alpine`). Node 22.12+ enabled `require(esm)` by default, so even if TypeScript did emit a `require()` call here, it would work. The PR commit message said "Node 24" — that was imprecise; 22.17.0 is what we actually run, and it's sufficient.

No code change needed. The pattern is consistent with how `apps/web` handles the same ESM-only package via `apps/web/src/lib/sandbox/sprites-client.ts`.

apps/realtime/src/index.ts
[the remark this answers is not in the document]
2witstudios: Fixed in commit `c4769a8f0`. The PTY connect path now has full quota and audit coverage:

- **Concurrency slot**: `acquireCodeExecutionSlot({ userId, tier })` is called before the sandbox is acquired. If at capacity, the connect is denied with `concurrency_limit`. The slot is threaded through `CheckAuthResult.releaseSlot` and stored on `TerminalSession`, then released on shell exit, reconnect cleanup, or socket disconnect — including the error path where `openShell` throws.
- **Budget charge**: `chargeCodeExecutionBudget({ userId, driveId, tenantId })` is called fire-and-forget after successful sandbox acquire.
- **Audit record**: `writeCodeExecutionAudit()` is called on session open (PTY session, origin `user`).

The user's `subscriptionTier` is fetched alongside the drive row to determine the tier for quota calculations.

apps/realtime/src/terminal/terminal-handler.ts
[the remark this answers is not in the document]
2witstudios: Fixed in commit `5bc2a42c7`. A 60-second re-authorization interval is started inside `onConnect` after the shell opens. On every tick, `checkAuth` is re-invoked. If it returns `ok: false`, the interval clears itself, the slot is released, the shell receives `SIGKILL`, the session is removed from the map, and the client receives `terminal:closed` with `exitCode: -2`.

The interval is also cleared on natural shell exit (`onExit`) and on explicit disconnect (`onDisconnect`), so no dangling timers.

Tests added for: interval is registered on the session, re-auth success keeps session alive, re-auth failure kills the session and emits the correct event.

apps/web/src/components/layout/middle-content/page-views/terminal/TerminalView.tsx
2witstudios: Fixed in commit `b881d63db`. Changed `{socket && (` to `{socket && isAdmin && (` on line 47 — `XtermTerminal` no longer mounts for non-admin users. Non-admins see only the warning banner; no socket event is emitted.

apps/web/src/components/layout/middle-content/page-views/terminal/TerminalView.tsx
2witstudios: Fixed in commit `b881d63db`. Added `const [error, setError] = useState<string | null>(null)`. `handleError` now calls `setError(message)` before toasting. The connecting overlay renders `<span className="text-sm text-red-400">{error}</span>` on error state instead of the spinner, so a `terminal:error` response clears the loader and shows the message.

apps/realtime/src/index.ts
2witstudios: Fixed in commit `c4769a8f0`. Replaced `const tenantId = driveRow?.ownerId ?? ''` with an explicit early return: `if (!driveRow) return { ok: false, reason: 'drive_not_found' }`, then `const tenantId = driveRow.ownerId` with no nullable fallback. Auth denial is also logged via `loggers.realtime.warn`.

apps/realtime/src/index.ts
2witstudios: Fixed in commit `c4769a8f0`. `createDbTerminalSessionStore()` is now called once at module level: `const dbTerminalSessionStorePromise = createDbTerminalSessionStore()`, consistent with `getRealtimeSpritesSdk()`. The `Promise.all` inside `makeTerminalCheckAuth` awaits the cached promise.

apps/realtime/src/terminal/terminal-handler.ts
2witstudios: Fixed in commit `5bc2a42c7`. Added `export const MAX_INPUT_BYTES = 4096` and guarded `onInput` with `p.data.length <= MAX_INPUT_BYTES`. Frames exceeding 4 KB are silently dropped. The constant is exported so tests can reference it directly.

apps/realtime/src/index.ts
2witstudios: Fixed in commit `c4769a8f0`. `loggers.realtime.warn` is now called on every auth denial path in `makeTerminalCheckAuth` (`page_not_found`, `no_edit_access`, `canRunCode` denial, `drive_not_found`, `sandboxResult` failure, `concurrency_limit`). `loggers.realtime.info` is called on successful session open and on shell exit.

apps/web/src/app/globals.css
2witstudios: Fixed in commit `ba72ef4b6`. Removed `@import "@xterm/xterm/css/xterm.css"` from `globals.css` and added `import '@xterm/xterm/css/xterm.css'` at the top of `TerminalView.tsx` (after `"use client"`). The stylesheet is now bundled only for routes that render the terminal.

apps/web/src/components/layout/middle-content/page-views/terminal/XtermTerminal.tsx
2witstudios: Fixed in commit `ba72ef4b6`. Added two comment lines above the `eslint-disable-next-line` in `XtermTerminal.tsx` explaining that `onReady`/`onError` are intentionally excluded because callers must stabilise them with `useCallback`, and including them would cause the terminal to remount on every parent render.

PR #1742, opened 2026-07-02
fix(voice): eliminate TTS word dropping (.tsx) + awkward pauses between chunks

2witstudios:
## Problem

Two distinct voice mode bugs:

1. **Word dropping** — The `packSentences` regex used `[^.!?]+` which can only match non-punctuation runs. When it hit a period inside an identifier like `Button.tsx`, it failed to match and silently dropped everything before the period. TTS would speak only "tsx" instead of "Button.tsx".

2. **Awkward pauses** — When the speech queue emptied between chunks during AI streaming, `onended` called `stopSpeakingStore()` which flickered voice state to `idle`. The next chunk required a full synthesis round-trip with no pre-fetch benefit, plus a visible UI state change.

### Word dropping (commit 1: `ac707f8`)

`packSentences` in `chunkForTts.ts` regex changed from:
```
/[^.!?]+[.!?]+(?:\s+|$)|[^.!?]+$/g
```
to:
```
/[\s\S]*?[.!?]+(?:\s|$)|[\s\S]+$/g
```

The lazy quantifier naturally skips non-terminal periods (file extensions, decimals, version numbers) because it expands until it finds a `[.!?]` that IS followed by whitespace or end-of-string.

### Awkward pauses (commit 2: `c516316`)

Three changes to `useVoiceMode.ts`:

1. **`onended` handler**: When queue is empty but AI is still streaming, keep `'speaking'` state and clear `currentAudioId`. A 500ms safety timer handles the edge case where the stream ends without delivering more text.

2. **`queueSentence`**: When voice is `'speaking'` but `currentAudioId` is null (between chunks), start playback immediately instead of pushing to a queue that no `onended` will drain.

3. **`speak()`**: Clear any pending transition timer when starting a new chunk to prevent stale timers from firing during playback.

The three changes form a closed state machine: `onended` sets timer + clears audioId → `queueSentence` detects null audioId → `speak()` clears timer before async fetch. No race conditions.

## Testing

- All 53 `chunkForTts` tests pass
- TypeScript clean
- Includes 9 regression tests for `.tsx`, decimals, version numbers, and streaming token boundaries
- AIDD 8-lens review: clean (see PR review comment)

### DIFF
--- a/apps/web/src/hooks/useVoiceMode.ts
+++ b/apps/web/src/hooks/useVoiceMode.ts
@@ -726,15 +726,24 @@ export function useVoiceMode({
               const { isEnabled: sEnabled, interactionMode: sMode } =
                 useVoiceModeStore.getState();
               if (sEnabled && sMode === 'conversation') {
-                playbackRefs.current.autoListenTimer = setTimeout(() => {
+                // Re-arming safety timer: checks every 500ms whether the stream
+                // has ended. If stream is still active, re-arms and waits again.
+                // This prevents voice from getting stuck in 'speaking' when the
+                // stream ends after the gap but no new text was delivered.
+                const recheck = () => {
                   playbackRefs.current.autoListenTimer = null;
-                  if (!isAIStreamingRef.current && !playbackRefs.current.audioSource) {
+                  if (playbackRefs.current.audioSource) return; // new audio started
+                  if (!isAIStreamingRef.current) {
                     stopSpeakingStore();
                     const { isEnabled: en, interactionMode: md } =
                       useVoiceModeStore.getState();
                     if (en && md === 'conversation') void startListening();
+                  } else {
+                    // Stream still active, re-arm
+                    playbackRefs.current.autoListenTimer = setTimeout(recheck, 500);
                   }
-                }, 500);
+                };
+                playbackRefs.current.autoListenTimer = setTimeout(recheck, 500);
               }
               return;
             }
--- a/apps/web/src/hooks/useVoiceMode.ts
+++ b/apps/web/src/hooks/useVoiceMode.ts
@@ -641,6 +641,12 @@ export function useVoiceMode({
         const audioId = createId();
         startSpeakingStore(audioId);
 
+        // Clear any pending transition timer from a previous gap-between-chunks
+        if (playbackRefs.current.autoListenTimer) {
+          clearTimeout(playbackRefs.current.autoListenTimer);
+          playbackRefs.current.autoListenTimer = null;
+        }
+
         const audioContext = getAudioContext();
         let audioBuffer: AudioBuffer;
 
@@ -707,6 +713,32 @@ export function useVoiceMode({
               return;
             }
 
+            // If the AI stream is still active, keep 'speaking' state so the UI
+            // doesn't flicker to idle between chunks. Clear the audio ID so
+            // queueSentence knows to trigger playback directly when the next
+            // chunk arrives (no onended will fire to drain the queue).
+            if (isAIStreamingRef.current) {
+              setCurrentAudioId(null);
+              callbacksRef.current.onSpeakComplete?.();
+              // Safety net: if the stream ends without delivering more text,
+              // this timer transitions out of speaking. The isAIStreamingRef
+              // check at fire time makes it a no-op if chunks are still flowing.
+              const { isEnabled: sEnabled, interactionMode: sMode } =
+                useVoiceModeStore.getState();
+              if (sEnabled && sMode === 'conversation') {
+                playbackRefs.current.autoListenTimer = setTimeout(() => {
+                  playbackRefs.current.autoListenTimer = null;
+                  if (!isAIStreamingRef.current && !playbackRefs.current.audioSource) {
+                    stopSpeakingStore();
+                    const { isEnabled: en, interactionMode: md } =
+                      useVoiceModeStore.getState();
+                    if (en && md === 'conversation') void startListening();
+                  }
+                }, 500);
+              }
+              return;
+            }
+
             stopSpeakingStore();
             callbacksRef.current.onSpeakComplete?.();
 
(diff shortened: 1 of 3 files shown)


2witstudios (commented):
## AIDD 8-Lens Review

### 1. Correctness ✅

**chunkForTts.ts** — The regex change is correct. Traced through:

| Input | Old regex | New regex |
|---|---|---|
| `Button.tsx` (no terminator) | matches only `tsx` via `[^.!?]+$` | matches `Button.tsx` via `[\s\S]+$` |
| `Button.tsx more text.` | matches `tsx more text.` — drops `Button.` | matches `Button.tsx more text.` as one sentence |
| `3.14` inside a sentence | drops `3`, matches `14...` | preserves `3.14` |
| `console.log(x).` | drops `console`, matches `log(x).` | preserves `console.log(x).` |

The lazy `[\s\S]*?` expands until it finds `[.!?]+` followed by `\s` or `$`, naturally skipping non-terminal periods. No content dropped in any traced scenario.

**useVoiceMode.ts** — Traced the three changes together as a state machine:

1. `onended` fires, queue empty, `isAIStreaming=true` → sets `currentAudioId=null`, starts 500ms safety timer
2. `queueSentence` called with new text → sees `currentAudioId === null` → shifts queue, calls `speak()` 
3. `speak()` → **clears the 500ms safety timer** → starts fetch → guards with audioId → plays

The key safety property: `speak()` clears the safety timer (change 1) before the async fetch starts, so the timer can't fire during an in-flight synthesis. The only way the timer fires is if no new `speak()` is called within 500ms — meaning the stream truly ended. Correct.

**One edge case I verified**: If the stream ends, the flush fires, `queueSentence` → `speak()` is called, but the TTS API takes >500ms to respond. Does the timer fire? No — `speak()` clears it at the top before the fetch. Safe.

### 2. RBAC ✅

No auth/permission changes. Client-side voice state management only.

### 3. TDD ✅

9 new regression tests covering the specific bug class:
- `.tsx` in mid-sentence, at end of sentence, standalone, multiple extensions
- Decimals (`3.14`), version numbers (`v1.2.3`)
- Streaming token boundary simulation
- Fixed the misleading `const x` test that was validating the bug

The streaming pause fix has no unit tests (it modifies a React hook with AudioContext dependencies that are hard to unit test without mocking the entire Web Audio API). This is acceptable — the logic is defensive state management, not algorithmic.

### 4. Type Safety ✅

- `currentAudioId` from `useVoiceModeStore.getState()` is properly typed as `string | null`
- `shift()!` non-null assertion is safe — `text` was just pushed to the array one line above
- No `any` types introduced
- No new imports needed

### 5. Mock Data Removal ✅

No mock data or test fixtures in production code.

### 6. Domain Logic ✅

The pause fix correctly distinguishes two end-of-queue scenarios:
- **Stream still active** → hold `'speaking'` state, wait for next chunk (no flicker)
- **Stream ended** → normal cleanup → auto-listen in conversation mode

The 500ms safety timer handles the "stream ended but VoiceCallPanel hasn't flushed yet" gap. The condition chain `!isAIStreamingRef.current && !playbackRefs.current.audioSource` prevents false positives when new audio is already in flight.

### 7. Quality ✅

Comments are thorough and explain WHY, not WHAT. The regex comment explains the old vs new behavior and references the specific bug class. The state machine comments in `onended` and `queueSentence` explain the between-chunks transition.

Minor nit: the `onSpeakComplete` callback fires per-chunk in both the streaming and non-streaming paths. This is consistent with existing behavior, but a future caller should know it's not a stream-end signal.

### 8. Security ✅

No new attack surface. Client-side audio state management. No user input handling changes.

---

**Verdict: Clean.** The regex fix is a root cause fix. The pause fix is a root cause fix (not a band-aid) — the three changes form a closed state machine where each transition is covered. No issues found.

2witstudios (commented):
## AIDD 8-Lens Review — REQUEST CHANGES

### 1. Correctness ⚠️ One issue

**Regex fix (`chunkForTts.ts`)** — Clean. Traced `Button.tsx`, `console.log`, `3.14`, `v1.2.3` through the new lazy quantifier. The `[\s\S]*?` correctly expands past non-terminal periods because `[.!?]+(?:\s|$)` only anchors on whitespace/end. No infinite loop risk — `lastIndex` always advances by ≥1 (the `[.!?]+` guarantees 1+ chars).

**Pause fix (`useVoiceMode.ts`)** — One edge case concern:

The 500ms safety timer in the `onended` streaming branch fires exactly once. If the AI stream stays active past 500ms (e.g., during a tool call that produces no text), the timer fires, sees `isAIStreamingRef.current === true`, does nothing, and the timer is gone. If the stream then ends with no remaining pending text:

1. `VoiceCallPanel` flush effect runs → no tail text → returns early
2. Nobody re-checks voice state
3. **Voice stuck in `'speaking'` indefinitely**

This is a realistic scenario during multi-step tool calls. Fix: re-arm the timer if stream is still active:

```typescript
playbackRefs.current.autoListenTimer = setTimeout(function recheck() {
  playbackRefs.current.autoListenTimer = null;
  if (playbackRefs.current.audioSource) return; // new audio started
  if (!isAIStreamingRef.current) {
    stopSpeakingStore();
    const { isEnabled: en, interactionMode: md } = useVoiceModeStore.getState();
    if (en && md === 'conversation') void startListening();
  } else {
    // Stream still active, re-arm
    playbackRefs.current.autoListenTimer = setTimeout(recheck, 500);
  }
}, 500);
```

### 2. RBAC ✅
No security boundary changes. TTS synthesis route unchanged.

### 3. TDD ✅
53 tests pass. 9 regression tests cover the regex fix (file extensions, decimals, versions, streaming token boundaries). Hook state machine changes are not unit-tested, which is acceptable for React hook audio lifecycle code.

### 4. Type Safety ✅
TypeScript clean.

### 5. Mock Data Removal ✅
N/A.

### 6. Domain Logic ✅
The `onended` → `setCurrentAudioId(null)` → `queueSentence` detect-and-resume pattern correctly prevents the deadlock where queued chunks would never play because no `onended` would fire to drain them.

### 7. Quality ✅
Comments explain WHY. Regex comment block is especially clear.

### 8. Security ✅
No untrusted input reaches the regex.

---

**Pre-existing observation (not introduced by this PR):** `startBargeInMonitoring` is in `speak()`'s dependency array but never called inside the function. Barge-in monitoring doesn't start during TTS playback. Worth tracking separately.

### Verdict
Requesting changes for the safety timer re-arm issue. The fix is a small code change (re-arm pattern shown above). Everything else passes all 8 lenses.

PR #1852, opened 2026-07-05
fix(web): OAuth discovery 404s — .well-known is unroutable in App Router

2witstudios:
## Summary
- Production bug: `pagespace login` failed because `GET https://pagespace.ai/.well-known/oauth-authorization-server` returned 404, blocking all CLI OAuth logins.
- Root cause: Next.js App Router does not route dot-prefixed folders, so `apps/web/src/app/.well-known/oauth-authorization-server/route.ts` never landed in the build manifest. Its `__tests__` only imported `GET()` directly, so they passed while the real URL was dead — this repo already works around the same limitation for `apple-app-site-association` by serving it as a static file from `public/.well-known/`.
- Fix: moved the handler to the routable `apps/web/src/app/api/well-known/oauth-authorization-server/route.ts`, added a `next.config.ts` rewrite mapping the untouched RFC 8414 URL to that destination, and fixed `middleware.ts` (which runs on the pre-rewrite pathname) to let the well-known path through unauthenticated.
- Deliberately kept the new route out of `app/api/oauth/` — that tree's `hardening.test.ts` sweep requires rate-limiting + audit logging on every route with zero exceptions, and this is unauthenticated-by-spec public metadata. Added it to the app-wide `security-audit-coverage.test.ts` allowlist instead.
- `packages/cli/src/auth/discover.ts` is unchanged in behavior — only its doc comment was updated to point at the new file location.
- Confirmed no other `app/.well-known/*` routes exist in the repo with the same bug.
- **Follow-up self-review fix**: extracted `WELL_KNOWN_REWRITES` to `apps/web/src/lib/well-known/rewrites.ts` as a single source of truth, imported by both `next.config.ts` and `middleware.ts`, instead of duplicating the path string in both files.
- **Second, more serious middleware gap found during PR convergence** (flagged by an automated reviewer, then verified and fixed): even with discovery working, the CLI's actual login would still fail — `/api/oauth/token`, `/api/oauth/device_authorization`, and `/api/oauth/revoke` are public-by-protocol-design (RFC 6749/7009/8628: the CLI has no browser session when calling them) but were **not** in `middleware.ts`'s public-route allowlist, so a session-less request got `401 Authentication required` before the route ever ran. `/api/oauth/authorize` had the same issue — its GET handler is documented to redirect to sign-in when there's no session, but middleware's blanket 401 preempted that. Fixed with exact-path allowlist entries (not a `startsWith` prefix, since `device_authorization`'s `/verify` and `/decision` sub-routes are the browser `/activate` screen and must keep requiring a session — proven by a control-case test). Verified directly against the real `middleware()` function, since none of the 130 existing OAuth route tests exercise middleware at all (they import `GET`/`POST` directly) — the same root-cause class as the original bug.

## Regression coverage
- `next-config-well-known-rewrite.test.ts` — asserts the rewrite mapping exists in `next.config.ts`.
- `well-known-oauth-discovery.test.ts` — proves middleware lets the well-known path through unauthenticated, with a control case proving other unauthenticated routes still redirect to sign-in.
- `oauth-public-endpoints.test.ts` — proves `authorize`/`token`/`revoke`/`device_authorization` are reachable with no session cookie, with a control case proving `device_authorization/verify` and `/decision` still require one.
- Relocated route test — 200 + valid RFC 8414 metadata, unauthenticated.
- Verified end-to-end against the actual built **standalone** server: `GET /.well-known/oauth-authorization-server` returns `200` with valid JSON metadata, no session/cookie required.

## Note for maintainers (separate, pre-existing, NOT fixed in this PR)
While verifying the above with a local Docker-equivalent standalone build, `.next/server/middleware-manifest.json` came back **empty** (`"middleware": {}`) even after a fully clean rebuild — `middleware.ts` isn't landing in the build at all in that path. This predates this PR (reproduces on unmodified `origin/master` too) and is unrelated to anything changed here; I verified this PR's middleware logic instead via direct unit tests against the real `middleware()` function (bypassing webpack/Next's build pipeline entirely, same pattern as this repo's existing `matcher.test.ts`/`security-headers.test.ts`), which is conclusive for correctness of the logic itself. But *whether middleware is actually bundled into the real production artifact* is a distinct, more alarming question this PR doesn't answer, and is worth a dedicated investigation.

Checked this against the official Next.js docs (this app runs 15.5.18) to firm up the theory rather than leave it as a guess:
- [`next.config.js: webpack`](https://nextjs.org/docs/app/api-reference/config/next-config-js/webpack) confirms `isServer` is `true` for **both** the Node.js server build and the Edge runtime build — `nextRuntime` (`'nodejs' | 'edge' | undefined`) is what actually distinguishes them. `apps/web/next.config.ts`'s `webpack()` only checks `isServer` before pushing `bunWorkspaceExternals` (which externalizes `@pagespace/lib`/`@pagespace/db` as `commonjs` requires) — it never checks `nextRuntime === 'nodejs'`, so the same CommonJS externalization is applied to the Edge middleware compilation too.
- [Middleware execution order](https://nextjs.org/docs/app/api-reference/file-conventions/proxy) confirms Middleware runs in the **Edge runtime by default** — a V8 isolate with no CommonJS/`require()` support at runtime — which is consistent with a broken/empty middleware bundle if externals meant for Node.js leak into that compilation.
- Same page's version history: **Next.js 15.5+ (stable) supports opting Middleware into the Node.js runtime** via `export const config = { runtime: 'nodejs', matcher: [...] }`. Since this app is already on 15.5.18, that flag would sidestep the whole Edge-runtime/CommonJS-externals class of problem and seems like the most direct fix, but I haven't verified it closes this specific gap — flagging as the most promising lead for whoever picks this up, not a verified fix.

Also confirmed independently: the `.well-known` dot-folder exclusion this PR's root cause depends on is **not** a documented Next.js App Router convention (only `_private` folders and `(route-groups)` are documented) — it's an empirically-observed, undocumented side effect of file-discovery tooling conventionally skipping dotfiles, not something Next.js promises to keep excluding. The fix in this PR (an explicit `next.config.ts` rewrite) doesn't depend on that behavior persisting, so it's robust either way.

## Test plan
- [x] `bun run --filter web typecheck`
- [x] `bun run --filter web lint` (no new warnings)
- [x] All new/relocated/existing middleware + OAuth tests green (219 tests across 17 files)
- [x] `bun run --filter web build` succeeds; `/api/well-known/oauth-authorization-server` present in `.next/server/app/api/well-known/`, no `.well-known` dot-folder anywhere in the manifest
- [x] Live smoke test against the built standalone server: real URL returns 200 with valid metadata, unauthenticated
- [x] `packages/cli` typecheck + tests green after building `@pagespace/sdk`
- [x] Rebased on latest `master` (which independently fixed the same stale `scaffold.test.ts` assertion — my duplicate fix for that was dropped during rebase, no conflict remains)
- [ ] Full `vitest run` (web) has 2 pre-existing, unrelated failures requiring a live Postgres `test` role (`admin-role-version.test.ts`, `activity-tools.test.ts`) and one pre-existing timezone-flake (`grouping.test.ts`) — confirmed present on `origin/master` before this change, not touched by this PR

https://claude.ai/code/session_01PmNKa7L6qQcw38EYmgNXzs

## Summary by CodeRabbit

* **New Features**
  * Added support for public OAuth discovery and grant endpoints without requiring a signed-in session.
  * Added automatic routing for `/.well-known/oauth-authorization-server` to the API endpoint used for discovery.

* **Bug Fixes**
  * Improved middleware handling so discovery requests are treated as public API traffic.
  * Kept browser-only OAuth verification and approval pages protected.

* **Tests**
  * Added coverage for well-known routing and unauthenticated OAuth endpoint behavior.

### DIFF
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -118,6 +118,12 @@ export async function middleware(req: NextRequest) {
     // Note: Cron routes handle their own auth via validateSignedCronRequest (HMAC-SHA256 + nonce)
     // Device/mobile auth endpoints authenticate via body tokens (device token, magic link),
     // not session cookies, so they must bypass the cookie check to allow cookie-expired recovery.
+    // OAuth grant endpoints are public by protocol design (RFC 6749/7009/8628): the CLI calls
+    // them with no browser session at all, authenticating via client_id/code/refresh_token/
+    // device_code in the request body instead — each route enforces its own auth internally
+    // (authorize's POST still requires a session; token/revoke/device_authorization never do).
+    // Exact matches only — device_authorization's /verify and /decision sub-routes are the
+    // browser-side /activate screen and DO require a session, so must not be swept in here.
     if (
       pathname.startsWith('/api/auth/csrf') ||
       pathname.startsWith('/api/auth/login-csrf') ||
@@ -130,6 +136,10 @@ export async function middleware(req: NextRequest) {
       pathname.startsWith('/api/mcp/') ||
       pathname.startsWith('/api/drives') ||
       pathname.startsWith('/api/cron/') ||
+      pathname === '/api/oauth/authorize' ||
+      pathname === '/api/oauth/token' ||
+      pathname === '/api/oauth/revoke' ||
+      pathname === '/api/oauth/device_authorization' ||
       pathname === '/api/memory/cron' ||
       pathname === '/api/pulse/cron' ||
       pathname === '/api/integrations/zoom/webhook' ||
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -103,6 +103,16 @@ export async function middleware(req: NextRequest) {
       return response;
     }
 
+    // RFC 8414 OAuth discovery metadata — must be reachable with no session,
+    // since it's the first request the CLI login flow makes. Middleware sees
+    // the pre-rewrite pathname (next.config's rewrites() runs after
+    // middleware), so this must match the original well-known URL, not its
+    // /api/oauth/authorization-server-metadata destination.
+    if (pathname === '/.well-known/oauth-authorization-server') {
+      const { response } = createSecureResponse(isProduction, req, { isAPIRoute: true });
+      return response;
+    }
+
     // Public routes that don't require authentication
     // Note: Cron routes handle their own auth via validateSignedCronRequest (HMAC-SHA256 + nonce)
     // Device/mobile auth endpoints authenticate via body tokens (device token, magic link),
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -14,6 +14,7 @@ import {
   isOriginValidationBlocking,
 } from '@/lib/auth';
 import { getSessionFromCookies } from '@/lib/auth/cookie-config';
+import { WELL_KNOWN_REWRITES } from '@/lib/well-known/rewrites';
 
 // Edge-safe middleware: only checks presence of auth tokens, not validity.
 // Full validation happens in route handlers via verifyAuth()/validateMCPToken().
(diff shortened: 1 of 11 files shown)


2witstudios:
Re: the P1 comment about OAuth grant endpoints bypassing the cookie gate — confirmed and fixed in b885d10a4.

Verified directly against the real `middleware()` function (not just the route handlers' own unit tests, which never exercise middleware at all): a session-less POST to `/api/oauth/token`, `/api/oauth/device_authorization`, or `/api/oauth/revoke` returned `401 Authentication required` before the route ever ran. `/api/oauth/authorize` had the same problem — its GET handler is documented to redirect to sign-in when there's no session, but middleware's blanket API 401 preempted that path from ever executing.

Added all four to `middleware.ts`'s existing public-routes allowlist, using **exact** path matches (not a `startsWith('/api/oauth/device_authorization')` prefix) — `device_authorization`'s `/verify` and `/decision` sub-routes are the browser-side `/activate` screen and must keep requiring a session. Added `oauth-public-endpoints.test.ts`, which exercises the real middleware function directly and includes a control case proving `/verify` and `/decision` still 401 with no cookie.

This is the same root-cause class as the original bug: route handlers are unit-tested by importing `GET`/`POST` directly, so 130 passing OAuth route tests gave no signal that the real HTTP path through middleware was blocked.

Leaving this thread open for your verification rather than auto-resolving it, per this session's convergence-loop policy (only pre-loop-addressed threads get auto-resolved; anything fixed during the loop stays open for reviewer sign-off).

PR #2054, opened 2026-07-13
feat(machines): propagate Claude Code credential into branch-terminal Sprites

2witstudios:
## Summary
- Each branch-terminal runs on its OWN, separate Sprite from the owning Machine's Sprite — no shared filesystem, no shared volume. Claude Code writes its OAuth credential to `/home/sprite/.claude/.credentials.json` on whichever Sprite ran `claude`, so a branch Sprite never has it.
- `propagateClaudeCredential` (`machine-branches.ts`) copies the root Sprite's `.credentials.json` (and `.claude.json` config, if present) into a branch Sprite by writing to a globally-unique (cuid2) temp path and atomically `mv`-ing it onto the real destination — never writing directly to the live path. `writeFiles`' `mode` only reliably applies at file *creation* (POSIX `open()` semantics), so a temp-then-rename is what makes `0o600` land correctly even on a refresh, without ever leaving the live file briefly absent or at the wrong permissions. Every housekeeping `rm`/`mv` exec carries an explicit `timeoutMs`/`maxBytes` so a wedged Sprite can't leak an unbounded background process. A generation counter guards against a stalled, overlapping refresh clobbering a newer one that already landed (process-local — see in-code doc comment on the accepted cross-process residual, which affects staleness only, never exposure).
- Deliberately does **not** delete an existing branch-side copy when the root read comes back empty — the driver maps every read failure to the same `null` a missing file produces, so that signal can't reliably distinguish a real `claude logout` from a transient root-Sprite hiccup.
- Wired at **three** call sites, covering both the explicit branch-management API and the actual PTY attach path a user's agent terminal goes through:
  - `spawnBranch` (fresh-provision and warm-reattach) and `attachBranch` — branch creation / the navigator's explicit attach API. The copy runs **after** the branch row is persisted, not before — doing it earlier would widen the window in which a concurrent racer's collision-reconciliation could kill the shared, name-keyed Sprite out from under the still-in-flight winner. Bounded by a 5s timeout inside `propagateClaudeCredential` itself, so a hibernating root Sprite can't hang a branch-attach HTTP response.
  - Realtime's `resolveAgentTerminalSandbox` → `resolveMachineSandbox` (`agent-terminal-access.ts`), via an optional `refreshBranchCredential` dep fired only for true branch-scope targets (`projectName` **and** `branchName` both set). This is the path a user's branch agent terminal actually opens/reattaches through on every cold PTY open — `spawnBranch`/`attachBranch` alone don't cover it. Awaited but bounded by its own 5s timeout rather than fully blocking or fire-and-forget (both extremes were tried and reverted during review).
- `findLiveMachineSandboxId` (`machine-session-manager.ts`) resolves the page's **current** session key (from its current driveId + drive owner) rather than a bare-`pageId` lookup — a page moved between drives can leave its OLD drive's session row behind under a different key, and a bare-`pageId` read could non-deterministically resolve that stale row. `apps/web/machine-branches-runtime.ts` and `apps/realtime/index.ts` (via a shared `resolveDriveOwnerContext` helper) both resolve the current driveId + owner first.

## Test plan
- [x] `bun run typecheck` — all 16 packages pass
- [x] `bun run build` (web) — passes
- [x] `packages/lib` full suite (`vitest run`) — 357 files / 8193 tests pass
- [x] `apps/realtime` full suite (`vitest run`) — 23 files / 832 tests pass

Coverage includes: credential + config copy on spawn and on attach; no-op when no root session, no credential yet, or an empty-but-ambiguous root read; resolver-throws resilience; refreshed-credential-on-reattach; 0o600 mode via temp-then-atomic-rename; concurrency-safety regression tests for row-persisted-before-copy, a stalled overlapping copy skipping its own stale rename, and a failed rename leaving an existing valid credential untouched; branch-vs-machine-vs-project scope gating in the realtime path; the `refreshBranchCredential` dep being optional and defense-in-depth against a throwing/never-settling implementation (fake-timers tests); every housekeeping exec carrying a bounded `timeoutMs`/`maxBytes`; and current-session-key resolution across a simulated drive move.

## Review history
This PR went through an extensive multi-round automated review cycle (CodeRabbit + Codex) that surfaced and fixed a series of real concurrency/security issues in the credential-copy path — each caught a genuine gap in the previous fix, converging on the write-to-temp-then-atomic-rename design described above. Final Codex pass: "Didn't find any major issues." All CI green, no unresolved merge conflicts.

## Summary by CodeRabbit

- **New Features**
  - Claude Code credentials and configuration now carry over automatically from the main machine to newly created or reattached branch terminals.
  - Updated credentials are refreshed when reconnecting to an existing branch.
  - Credential files use secure permissions.

- **Bug Fixes**
  - Terminal access continues normally when credential synchronization fails or no active credentials are available.

### DIFF
--- a/apps/realtime/src/terminal/agent-terminal-access.ts
+++ b/apps/realtime/src/terminal/agent-terminal-access.ts
@@ -100,9 +100,15 @@ export interface ResolveMachineSandboxDeps {
    * explicit attach API, neither of which this realtime PTY bridge calls.
    *
    * Optional so tests that don't exercise branch scope (or don't care about
-   * credential propagation) can omit it. MUST be best-effort on the
-   * implementation side — swallow its own failures — since a credential
-   * refresh must never block or fail opening the PTY itself.
+   * credential propagation) can omit it. Called fire-and-forget — NEVER
+   * awaited by `resolveMachineSandbox` (a cold/hibernating Sprite's fs
+   * read/write can take up to the Sprite fs API's 30s timeout, with one
+   * retry, so up to ~60s; blocking every branch PTY open on that would
+   * regress the "opening a PTY is itself the wake, nothing upstream waits on
+   * it" invariant). MUST still be best-effort on the implementation side —
+   * swallow its own failures — as defense in depth against an unhandled
+   * rejection, even though a rejection can no longer fail or delay the PTY
+   * resolution itself.
    */
   refreshBranchCredential?: (args: { machineId: string; sandboxId: string }) => Promise<void>;
 }
@@ -156,14 +162,23 @@ export async function resolveMachineSandbox(
   // validation ever changes, or a caller wires a resolver that doesn't
   // enforce it (e.g. a test double).
   if (target.projectName !== undefined && target.branchName !== undefined && deps.refreshBranchCredential) {
-    try {
-      await deps.refreshBranchCredential({ machineId: target.machineId, sandboxId: resolved.sandboxId });
-    } catch {
+    // Fire-and-forget — NEVER awaited (caught in review, P2). A cold or
+    // hibernating Sprite's filesystem read/write can take up to the Sprite
+    // fs API's 30s timeout, WITH one retry (so up to ~60s) — blocking every
+    // branch PTY open on that would regress this codebase's core Sprite
+    // invariant that opening a PTY is itself the wake and NOTHING upstream
+    // of it waits on that wake (see `buildMachineSandbox`'s own doc comment
+    // on this file's sibling `acquire`). The refresh still normally lands
+    // well before a user manually runs `claude`; the narrower race — typing
+    // `claude` faster than an in-flight refresh completes — self-heals
+    // moments later once the copy finishes, and is a FAR smaller cost than a
+    // multi-second stall on every cold terminal open.
+    void deps.refreshBranchCredential({ machineId: target.machineId, sandboxId: resolved.sandboxId }).catch(() => {
       // Best-effort by contract (see the doc comment on
       // `ResolveMachineSandboxDeps.refreshBranchCredential`) — defense in
       // depth here too, so an implementation that forgets to swallow its own
-      // errors still never fails opening the PTY over a credential hiccup.
-    }
+      // errors still never produces an unhandled rejection.
+    });
   }
 
   return {
--- a/apps/realtime/src/terminal/agent-terminal-access.ts
+++ b/apps/realtime/src/terminal/agent-terminal-access.ts
@@ -100,17 +100,41 @@ export interface ResolveMachineSandboxDeps {
    * explicit attach API, neither of which this realtime PTY bridge calls.
    *
    * Optional so tests that don't exercise branch scope (or don't care about
-   * credential propagation) can omit it. Called fire-and-forget — NEVER
-   * awaited by `resolveMachineSandbox` (a cold/hibernating Sprite's fs
-   * read/write can take up to the Sprite fs API's 30s timeout, with one
-   * retry, so up to ~60s; blocking every branch PTY open on that would
-   * regress the "opening a PTY is itself the wake, nothing upstream waits on
-   * it" invariant). MUST still be best-effort on the implementation side —
-   * swallow its own failures — as defense in depth against an unhandled
-   * rejection, even though a rejection can no longer fail or delay the PTY
-   * resolution itself.
+   * credential propagation) can omit it. Awaited by `resolveMachineSandbox`,
+   * but bounded by `CREDENTIAL_REFRESH_TIMEOUT_MS` (`withTimeout`) rather
+   * than either fully blocking or fire-and-forget — see the inline comment
+   * at that call site for why both extremes were tried and reverted. MUST
+   * still be best-effort on the implementation side — swallow its own
+   * failures — as defense in depth against an unhandled rejection.
    */
   refreshBranchCredential?: (args: { machineId: string; sandboxId: string }) => Promise<void>;
+}
+
+/**
+ * Hard cap on how long `refreshBranchCredential` may delay a fresh branch
+ * PTY open. Long enough for the common case (warm root + branch Sprite,
+ * normally well under a second) to complete reliably; short enough that a
+ * cold/hibernating Sprite's fs-timeout tail (up to ~60s, see the doc comment
+ * on `machine-branches.ts`'s `propagateClaudeCredential`) never turns into a
+ * near-minute-long terminal-open stall.
+ */
+const CREDENTIAL_REFRESH_TIMEOUT_MS = 5_000;
+
+/** Resolve once `promise` settles OR `timeoutMs` elapses, whichever first — never rejects. The timer is always cleared, and `promise` keeps running past the bound rather than being cancelled (a slow settle still lands, just too late to have been waited on). */
+function withTimeout(promise: Promise<void>, timeoutMs: number): Promise<void> {
+  return new Promise((resolve) => {
+    const timer = setTimeout(resolve, timeoutMs);
+    promise.then(
+      () => {
+        clearTimeout(timer);
+        resolve();
+      },
+      () => {
+        clearTimeout(timer);
+        resolve();
+      },
+    );
+  });
 }
 
 export type ResolveMachineSandboxResult =
@@ -162,23 +186,24 @@ export async function resolveMachineSandbox(
   // validation ever changes, or a caller wires a resolver that doesn't
   // enforce it (e.g. a test double).
   if (target.projectName !== undefined && target.branchName !== undefined && deps.refreshBranchCredential) {
-    // Fire-and-forget — NEVER awaited (caught in review, P2). A cold or
-    // hibernating Sprite's filesystem read/write can take up to the Sprite
-    // fs API's 30s timeout, WITH one retry (so up to ~60s) — blocking every
-    // branch PTY open on that would regress this codebase's core Sprite
-    // invariant that opening a PTY is itself the wake and NOTHING upstream
-    // of it waits on that wake (see `buildMachineSandbox`'s own doc comment
-    // on this file's sibling `acquire`). The refresh still normally lands
-    // well before a user manually runs `claude`; the narrower race — typing
-    // `claude` faster than an in-flight refresh completes — self-heals
-    // moments later once the copy finishes, and is a FAR smaller cost than a
-    // multi-second stall on every cold terminal open.
-    void deps.refreshBranchCredential({ machineId: target.machineId, sandboxId: resolved.sandboxId }).catch(() => {
-      // Best-effort by contract (see the doc comment on
-      // `ResolveMachineSandboxDeps.refreshBranchCredential`) — defense in
-      // depth here too, so an implementation that forgets to swallow its own
-      // errors still never produces an unhandled rejection.
-    });
+    // Awaited, but bounded — NOT fire-and-forget (tried that, reverted: a
+    // fresh branch-scoped Claude terminal calls `openShell` with `claude`
+    // immediately after this resolves, and copying the credential moments
+    // later does nothing for a process that already started without it —
+    // see review history). Also NOT unbounded: a cold/hibernating Sprite's
+    // fs read/write can take up to the Sprite fs API's 30s timeout, WITH one
+    // retry (so up to ~60s), and blocking every branch PTY open on that
+    // would regress this codebase's Sprite invariant that opening a PTY is
+    // itself the wake and nothing upstream waits long on it. The bound
+    // covers the common case (warm root + branch Sprite, sub-second)
+    // reliably while capping the cold-Sprite worst case to a few seconds —
+    // still slower than ideal, but nowhere near a minute-long stall. The
+    // refresh keeps running past the bound rather than being cancelled — a
+    // slow copy still lands, just too late to help THIS launch.
+    await withTimeout(
+      deps.refreshBranchCredential({ machineId: target.machineId, sandboxId: resolved.sandboxId }),
+      CREDENTIAL_REFRESH_TIMEOUT_MS,
+    );
   }
 
   return {
--- a/apps/realtime/src/terminal/agent-terminal-access.ts
+++ b/apps/realtime/src/terminal/agent-terminal-access.ts
@@ -90,6 +90,21 @@ export interface ResolveMachineSandboxDeps {
    * `sdk.getSprite`, not three.
    */
   getSprite: (sandboxId: string) => Promise<SpriteInstanceLike>;
+  /**
+   * BRANCH-scope only (never machine/project — those run ON the root
+   * Sprite, which already has its own credential): refresh the branch
+   * Sprite's Claude Code credential from the root Machine's Sprite before
+   * handing back a fresh PTY resolution. This is the actual attach path a
+   * user's branch agent terminal goes through — `spawnBranch`/`attachBranch`
+   * (`machine-branches.ts`) only cover branch creation and the navigator's
+   * explicit attach API, neither of which this realtime PTY bridge calls.
+   *
+   * Optional so tests that don't exercise branch scope (or don't care about
+   * credential propagation) can omit it. MUST be best-effort on the
+   * implementation side — swallow its own failures — since a credential
+   * refresh must never block or fail opening the PTY itself.
+   */
+  refreshBranchCredential?: (args: { machineId: string; sandboxId: string }) => Promise<void>;
 }
 
 export type ResolveMachineSandboxResult =
@@ -130,6 +145,17 @@ export async function resolveMachineSandbox(
     sprite = await deps.getSprite(resolved.sandboxId);
   } catch {
     return { ok: false, reason: 'provision_failed', sandboxId: resolved.sandboxId };
+  }
+
+  if (target.branchName !== undefined && deps.refreshBranchCredential) {
+    try {
+      await deps.refreshBranchCredential({ machineId: target.machineId, sandboxId: resolved.sandboxId });
+    } catch {
+      // Best-effort by contract (see the doc comment on
+      // `ResolveMachineSandboxDeps.refreshBranchCredential`) — defense in
+      // depth here too, so an implementation that forgets to swallow its own
+      // errors still never fails opening the PTY over a credential hiccup.
+    }
   }
 
   return {
--- a/apps/realtime/src/terminal/agent-terminal-access.ts
+++ b/apps/realtime/src/terminal/agent-terminal-access.ts
@@ -147,7 +147,15 @@ export async function resolveMachineSandbox(
     return { ok: false, reason: 'provision_failed', sandboxId: resolved.sandboxId };
   }
 
-  if (target.branchName !== undefined && deps.refreshBranchCredential) {
+  // Both `projectName` AND `branchName` must be set for true branch scope —
+  // matching `resolveScopeKey`/`resolveScopeLocation`'s own discriminator
+  // (`agent-terminals.ts`), rather than relying on the implicit invariant
+  // that a real `resolveAgentTerminal` already rejects `branchName` without
+  // `projectName` as `invalid_target` before ever reaching here. Checking
+  // both explicitly means this gate stays correct even if that upstream
+  // validation ever changes, or a caller wires a resolver that doesn't
+  // enforce it (e.g. a test double).
+  if (target.projectName !== undefined && target.branchName !== undefined && deps.refreshBranchCredential) {
     try {
       await deps.refreshBranchCredential({ machineId: target.machineId, sandboxId: resolved.sandboxId });
     } catch {
--- a/packages/lib/src/services/machines/machine-branches.ts
+++ b/packages/lib/src/services/machines/machine-branches.ts
@@ -210,15 +210,22 @@ async function cloneAndCheckoutBranch({
 /**
  * Copy Claude Code's OAuth credential (and config, if present) from the root
  * Machine's Sprite into a branch-terminal's freshly provisioned/attached one.
- * Each file is handled independently: absent on the root Sprite — most
- * commonly because the user hasn't run `claude` there yet — copies nothing
- * (a branch-terminal always ends up usable even with no credential to copy);
- * absent on the root Sprite when it WAS present before (an explicit `claude
- * logout`, or manual removal) actively REMOVES any stale copy already on the
- * branch Sprite, rather than leaving it silently authenticated after the
- * source was cleared. Since this refresh runs on every reattach specifically
- * to keep rotated credentials in sync, a revocation on the root is exactly
- * the kind of change that sync must also propagate (caught in review).
+ * Each file is copied independently and skipped when the root read comes
+ * back empty — most commonly because the user hasn't run `claude` there yet
+ * — so a branch-terminal always ends up usable even with no credential to
+ * copy.
+ *
+ * Deliberately does NOT delete an existing branch-side copy on an empty
+ * root read, even though that read coming back empty could also mean an
+ * explicit `claude logout` on the root (a tempting "propagate the
+ * revocation" behavior — tried and reverted, see review history). The
+ * driver's `readFile` maps EVERY read failure to the same `null` a missing
+ * file produces (`sprites.ts`'s `readFileToBuffer`: "a missing file (or ANY
+ * read failure after a wake retry) resolves to null") — there is no way,
+ * from this signal alone, to tell "confirmed gone" apart from "root Sprite
+ * was briefly unreachable." Deleting on that ambiguous signal would risk
+ * destroying a branch's perfectly valid, working credential on a transient
+ * hiccup — strictly worse than the staleness this would have closed.
  *
  * Best-effort: any failure (root Sprite unreachable mid-copy, etc.) is
  * swallowed rather than failing the spawn/attach it's called from. A branch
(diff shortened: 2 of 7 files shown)


packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -381,6 +451,10 @@ export async function attachBranch({
 
   const handle = await host.attach({ machineId: existing.sandboxId });
   if (!handle) return { ok: false, reason: 'vanished' };
+
+  // Refresh on every reattach — see `propagateClaudeCredential`'s doc comment
+  // on why this isn't a one-time, spawn-only copy.
+  await propagateClaudeCredential({ machineId, branchHandle: handle, resolveRootMachineHandle });
```
2witstudios, line 661: Good catch — confirmed and fixed in f1936b955.

You're right that `spawnBranch`/`attachBranch` are not the path a branch's agent terminal actually opens through. The real path is realtime's `resolveAgentTerminalSandbox` → `resolveMachineSandbox` (`agent-terminal-access.ts`), which resolves the branch Sprite directly via `resolveAgentTerminal`/`branchStore` and never touches `machine-branches.ts`.

Fix:
- Exported `propagateClaudeCredential` from `machine-branches.ts` for reuse.
- Added `findLiveMachineSandboxId` to `machine-session-manager.ts` (a read-only pageId→sandboxId lookup), used both to dedupe `machine-branches-runtime.ts`'s existing inline query and for the new realtime wiring.
- Added an optional `refreshBranchCredential` dep to `resolveMachineSandbox`, invoked only for branch-scope targets (machine/project scope already run on the root Sprite, which already has the credential). Wired in `apps/realtime/src/index.ts`'s `resolveAgentTerminalSandbox`, reusing the connect's existing Sprite handle cache so only the root Sprite's read is a new network call.
- Best-effort at both the call site and the implementation (try/catch in both places) — a credential refresh must never block or fail opening the PTY.

Added 8 new tests in `agent-terminal-access.test.ts` covering branch vs. machine vs. project scope, the dep being optional, and defense-in-depth against a misbehaving implementation that throws. Full monorepo typecheck + targeted test suites are green.

Leaving this thread open for your verification rather than resolving it myself.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,59 @@ async function cloneAndCheckoutBranch({
...
+  try {
+    const rootHandle = await resolveRootMachineHandle(machineId);
+    if (!rootHandle) return;
+
+    for (const path of [CLAUDE_CREDENTIALS_PATH, CLAUDE_CONFIG_PATH]) {
+      const content = await rootHandle.readFile({ path });
+      if (!content) continue;
+      // Both restricted 0o600 — the credentials file is the OAuth secret
+      // itself, and the config file can carry account/org metadata, so
+      // neither belongs world-readable in the branch Sprite's home dir.
```
2witstudios: Confirmed valid — fixed in 3d38b15c0.

You're right: `writeFiles`' `mode` is a creation-time argument (POSIX `open()` semantics), so it's silently ignored on a write that overwrites an already-existing file — which is exactly what happens on every refresh (reattach, credential rotation). Added an explicit `chmod 600` exec (argv-array form via the branch handle's own `exec`, never a shell string) after each write, so both the credentials file and the config file get re-secured on every copy, not just the first. Also made both files consistently 0o600 (the config file previously had no explicit mode at all).

New test asserts the exact chmod exec calls land in the branch Sprite's exec log for both files. Thanks for catching this.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -302,6 +378,15 @@ export async function spawnBranch({
     return { ok: false, reason: cloned.reason, detail: cloned.detail };
   }
 
+  // Persist the row FIRST, before the credential copy — not after. A
```
2witstudios, line 566: Confirmed valid, real race — fixed in 5b2d7558d.

You're right: the credential copy's extra I/O (root-Sprite read + branch-Sprite writes/execs) was sitting between the successful clone and the row being persisted, widening that window. A concurrent racer whose own clone fails against the same name-keyed shared Sprite calls `reconcileProvisionCollision`, which looks up the row to decide whether the Sprite it's about to kill is the winner's — with no row yet, it would conclude the shared Sprite was its own redundant one and kill it out from under the still-mid-copy winner.

Moved `propagateClaudeCredential` to run AFTER `store.updateSandboxId`/`store.create` succeeds (at each success return) instead of before both branches. Confirmed the realtime `refreshBranchCredential` path has no analogous issue — it only ever reads an already-persisted branch row, never creates/updates one.

Added two direct regression tests asserting the row is already visible via `store.findByName` at the exact moment the credential copy starts, for both the fresh-create and re-provision-after-vanish paths.

packages/lib/src/services/sandbox/machine-session-manager.ts
```diff
@@ -438,6 +438,46 @@ export async function acquireMachineSession(
...
+ * only disappears once that session tears down on its own). A bare-`pageId`
+ * lookup with no ordering could then non-deterministically return that STALE
+ * row — whose Sprite may belong to a different owner/tenant context entirely
+ * — and hand back state (here, a Claude Code credential) that was never that
+ * page's current owner's to give out. Deriving the exact expected key first
+ * makes the lookup exact by construction: it resolves the CURRENT session or
+ * nothing, never a stale one from a prior drive.
+ *
+ * Lazily resolves the db client, schema table, and `eq` operator so callers
+ * that don't exercise this path (most tests) never load the DB module graph.
```
2witstudios, line 464: Confirmed valid — this is a real cross-drive credential-leak risk. Fixed in b31b232a7ae3d68671b39c26116924118631f19e.

You're right: `machine_sessions.sessionKey` namespaces by tenant + drive + page, and a page moved between drives leaves its OLD drive's row behind (a drive move has no reason to touch `machine_sessions`). A bare-`pageId` lookup with no ordering could then resolve that stale row instead of the current one.

Fixed by having `findLiveMachineSandboxId` take the same `{tenantId, driveId, pageId, secret}` the session store's own key derivation uses, deriving the CURRENT `sessionKey` via `deriveMachineSessionKey`, and querying by that (the table's actual unique constraint) instead of the bare pageId. Both callers now resolve the page's current driveId + drive owner first — `machine-branches-runtime.ts`'s `resolveRootMachineHandle` reuses the existing `findMachinePage()` plus a new `drives.ownerId` lookup, and `apps/realtime/src/index.ts` factors a `resolveDriveOwnerContext()` helper (the same two-query lookup `buildMachineSandbox.acquire` already did inline) shared by both `acquire` and the new `refreshBranchCredential` wiring.

apps/realtime/src/terminal/agent-terminal-access.ts
```diff
@@ -132,6 +153,34 @@ export async function resolveMachineSandbox(
     return { ok: false, reason: 'provision_failed', sandboxId: resolved.sandboxId };
   }
 
+  // Both `projectName` AND `branchName` must be set for true branch scope —
+  // matching `resolveScopeKey`/`resolveScopeLocation`'s own discriminator
+  // (`agent-terminals.ts`), rather than relying on the implicit invariant
+  // that a real `resolveAgentTerminal` already rejects `branchName` without
+  // `projectName` as `invalid_target` before ever reaching here. Checking
```
2witstudios, line 184: Confirmed valid — fixed in b31b232a7ae3d68671b39c26116924118631f19e.

Verified the concrete number: the Sprite fs API's op timeout is 30s (`FS_OP_TIMEOUT_MS` in `sprites.ts`), and `fsWithWakeRetry` retries once on failure, so a cold/hibernating Sprite could stall a single read or write for up to ~60s. Blocking every branch-scope PTY resolution on that would regress this codebase's own stated invariant that opening a PTY is itself the Sprite's wake and nothing upstream should wait on it.

Changed `refreshBranchCredential`'s invocation in `resolveMachineSandbox` to fire-and-forget (`void ...().catch(() => {})`) instead of awaited — the refresh still normally completes well before a user manually runs `claude`; the narrower race (typing `claude` faster than an in-flight refresh) self-heals moments later once the copy finishes, which is a far smaller cost than a multi-second-to-a-minute stall on every cold terminal open.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,72 @@ async function cloneAndCheckoutBranch({
...
+export async function propagateClaudeCredential({
+  machineId,
+  branchHandle,
+  resolveRootMachineHandle,
+}: {
+  machineId: string;
+  branchHandle: MachineHandle;
+  resolveRootMachineHandle: (machineId: string) => Promise<MachineHandle | null>;
+}): Promise<void> {
+  try {
```
2witstudios: Good design catch, agreed — fixed in 72e5a4c7b0cc068d42ec0d8a0726462046529d62.

You're right: since this refresh runs on every reattach specifically to keep rotated credentials in sync, a revocation (explicit `claude logout` or manual removal) on the root is exactly the kind of change that sync should also propagate, not silently skip.

Changed: when the root's read for a given path comes back empty, exec `rm -f <path>` on the branch handle instead of skipping to the next file — removing any stale copy. `rm -f` is a harmless no-op when nothing was ever copied (root never logged in).

Added two tests: removal on the next reattach after the root's credential disappears, and that removing a never-copied file is a harmless no-op.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,72 @@ async function cloneAndCheckoutBranch({
...
+    const rootHandle = await resolveRootMachineHandle(machineId);
+    if (!rootHandle) return;
+
+    for (const path of [CLAUDE_CREDENTIALS_PATH, CLAUDE_CONFIG_PATH]) {
```
2witstudios: Confirmed valid and reverted — this was a real regression I introduced in the prior commit. Fixed in 23832dba8e18fd6d78564bb286bf44c867941035.

You're right: `readFileToBuffer` maps EVERY read failure to the same `null` a missing file produces (`sprites.ts`: "a missing file (or any read failure after a wake retry) resolves to null"), so there's no reliable way to distinguish a real `claude logout` from a transient root-Sprite read hiccup. Deleting on that ambiguous signal risked destroying a branch's perfectly valid, working credential on a momentary blip — strictly worse than the staleness gap the deletion was meant to close.

Reverted to a pure skip-on-empty. Added a regression test guarding against reintroducing the deletion (proves an empty root read leaves an existing branch credential untouched), and removed the now-unused `rm -f` simulation from the fake test host.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,122 @@ async function cloneAndCheckoutBranch({
...
+ * hiccup — strictly worse than the staleness this would have closed.
+ *
+ * Best-effort: any failure (root Sprite unreachable mid-copy, etc.) is
+ * swallowed rather than failing the spawn/attach it's called from, AND the
+ * whole call is bounded by `CREDENTIAL_COPY_TIMEOUT_MS` so a slow/hibernating
+ * Sprite can't hang the caller's response. A branch without the credential
+ * still works, it just needs its own `claude` login.
+ *
+ * Exported for reuse by the realtime agent-terminal PTY bridge
+ * (`apps/realtime/src/index.ts`'s `resolveAgentTerminalSandbox`), which is
```
2witstudios, line 308: Confirmed and fixed in a109d0d7d3036919bdc8d4d6a2bda4c50dccd821.

You're right: `exec` resolves with an exitCode rather than throwing on a nonzero exit, so a failed `chmod` was silently ignored — meaning an OVERWRITE of an already-existing credential/config file (the exact scenario the chmod step exists for) could stay at whatever permissive mode it already had while this function reported success.

Now checks `chmod`'s exit code and throws on nonzero (caught by the existing best-effort try/catch, so it still never fails the caller) — a failed chmod aborts the rest of the copy for that call rather than being silently trusted. Added a test proving a failing chmod stops the copy (the second file is never reached) without failing the spawn itself.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,129 @@ async function cloneAndCheckoutBranch({
...
+          //
+          // `writeFiles`' `mode` only takes effect at file CREATION (POSIX
+          // open()'s mode argument is ignored once a file already exists) —
+          // a refresh on reattach overwrites an ALREADY-EXISTING file (this
+          // is never the first copy after the first reattach), so the
+          // requested mode alone would silently leave whatever permissions
+          // that file already had. An explicit chmod after the write
+          // re-secures it either way, regardless of whether this write
+          // created or overwrote it.
+          await branchHandle.writeFiles([{ path, content, mode: 0o600 }]);
```
2witstudios: Confirmed and fixed in b32f9dc96d72dc55ed293e319bb1c96bf05ecd70.

You're right — stopping the copy loop wasn't enough; the just-written credential was still left at whatever (possibly permissive) mode the destination already had. Now attempts `rm -f <path>` on the file before throwing, so a failed chmod removes the copy entirely (fail closed) rather than leaving a valid, freshly-refreshed OAuth credential sitting at the wrong permissions. Best-effort — if even the rm fails, the outer catch swallows it same as any other failure.

Updated the chmod-failure test to simulate the deletion and assert the credentials file is actually gone afterward.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,116 @@ async function cloneAndCheckoutBranch({
...
+  branchHandle: MachineHandle;
+  resolveRootMachineHandle: (machineId: string) => Promise<MachineHandle | null>;
+}): Promise<void> {
+  await withTimeout(
+    (async () => {
+      try {
+        const rootHandle = await resolveRootMachineHandle(machineId);
+        if (!rootHandle) return;
+
+        for (const path of [CLAUDE_CREDENTIALS_PATH, CLAUDE_CONFIG_PATH]) {
```
2witstudios, line 345: Confirmed and fixed in 15f72a9beb9fe7e590fd040022eefaf5dbc7d776 — redesigned rather than patched further.

You're right that bounding chmod and cleaning up on timeout would just add another layer of the same problem (what bounds the cleanup rm? what if it's ALSO slow?). Instead, replaced write-then-chmod with delete-then-write: `rm -f <path>` before every `writeFiles` call. `writeFiles`' `mode` only applies at file CREATION (POSIX `open()` semantics) — deleting first means every write IS a creation, so `mode: 0o600` reliably applies immediately in the SAME call. There's no longer a separate chmod step that can independently fail, stall, or leave a window where the fresh secret sits at the old permissions — the whole class of problem is gone by construction rather than bounded/detected/cleaned-up after the fact.

Removed the now-unnecessary chmod-verification and fail-closed-cleanup logic and its tests, replaced with a test proving rm happens before write for both files, and one proving a throwing rm exec doesn't fail the spawn.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,129 @@ async function cloneAndCheckoutBranch({
...
+          // the delete then fails (or the caller's timeout elapses in
+          // between), the branch is left with NO credential at all, even
+          // though it had a perfectly valid one moments before — a
+          // regression on exactly the transient Sprite/FS hiccups this
+          // best-effort path exists to tolerate (caught in review, twice).
+          //
+          // `writeFiles`' `mode` reliably applies to a genuine CREATION —
+          // the temp path never existed before — and `mv` on the same
+          // filesystem is atomic: the destination is either the OLD valid
+          // credential or the NEW one, NEVER wrong-permission or briefly
```
2witstudios: Confirmed and fixed in 003ef778ee469deb9750f90e1d7ec03ee7b5f152 — you're right that delete-then-write just traded one window for a worse one.

Replaced with write-to-temp-then-atomic-rename: write to `<path>.tmp` (mode 0o600, a genuine CREATION since the temp path never existed) then `mv <path>.tmp <path>`. `mv` on the same filesystem is atomic — the destination is either the OLD valid credential or the NEW one, NEVER wrong-permission or briefly absent. If anything fails before (or during) the rename, the live file is completely untouched; on a failed rename, best-effort cleans up the orphaned temp file.

Added the core regression guard as a test: a failed `mv` leaves an EXISTING valid credential at the real path untouched, not gone.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,140 @@ async function cloneAndCheckoutBranch({
...
+          //
+          // `writeFiles`' `mode` reliably applies to a genuine CREATION —
+          // and `mv` on the same filesystem is atomic: the destination is
+          // either the OLD valid credential or the NEW one, NEVER
+          // wrong-permission or briefly absent in between. If anything
+          // fails before the rename, the live file at `path` is completely
+          // untouched.
+          const tempPath = `${path}.tmp`;
+          // Clear the temp path FIRST — a fixed name isn't guaranteed to be
+          // fresh (a prior attempt could have crashed between writing it
```
2witstudios: Confirmed and fixed in 4784412a0cb27558a4b27a005d2dce1a782c3c25.

You're right — a fixed temp name isn't guaranteed fresh (a prior attempt could have crashed between writing it and renaming it, or its own cleanup could have failed). Writing to an already-existing temp path is an overwrite, not a creation, so the mode would be silently ignored — reintroducing exactly the problem this temp-file flow exists to avoid.

Fixed by unconditionally `rm -f`-ing the temp path before writing to it. This is safe unconditionally (unlike the reverted delete-then-write design): any failure here only ever touches the disposable temp path, never the live file at the real destination.

Updated the test to also assert the rm-then-write-then-mv exec sequence for both files.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,145 @@ async function cloneAndCheckoutBranch({
...
+          // and renaming it, or its own cleanup could have failed), and
+          // writing to an ALREADY-EXISTING temp path would be an overwrite,
+          // not a creation — silently keeping whatever (possibly
+          // permissive) mode that stale file already had, which the `mv`
+          // below would then promote onto the real credential, reintroducing
+          // the exact problem this temp-file flow exists to avoid (caught
+          // in review). Checked, not fire-and-forget: if the clear itself
+          // fails, abort BEFORE writing rather than assume the temp path is
+          // now clear (caught in review, again) — still safe either way,
+          // since nothing has touched the live file yet.
```
2witstudios: Confirmed and fixed in 46819ff26c742bddb7cf9054775f598239dcbf91.

You're right — ignoring the clear's exit code meant a failed `rm -f` (unusual, but possible) would let the write proceed against a potentially-still-there stale temp file, reintroducing the overwrite-keeps-old-permissions problem this whole flow exists to avoid.

Now checks the clear's `exitCode` and throws before ever writing to the temp path if it failed (caught by the existing best-effort try/catch) — still safe either way, since nothing has touched the live file yet at that point. Added a test proving a failing temp-clear aborts before writeFiles/mv ever run.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,177 @@ async function cloneAndCheckoutBranch({
...
+          // THAT window but opened a worse one: if the write that follows
+          // the delete then fails (or the caller's timeout elapses in
+          // between), the branch is left with NO credential at all, even
+          // though it had a perfectly valid one moments before — a
+          // regression on exactly the transient Sprite/FS hiccups this
+          // best-effort path exists to tolerate (caught in review, twice).
+          //
+          // `writeFiles`' `mode` reliably applies to a genuine CREATION —
+          // and `mv` on the same filesystem is atomic: the destination is
+          // either the OLD valid credential or the NEW one, NEVER
```
2witstudios, line 364: Confirmed and fixed in ae5f9a0781e71ddff0326abde6278b8920c97e59 — this is real, and my first attempt at fixing it (an in-flight Set with skip-if-already-running) had its own correctness gap: if a call's underlying work never settles at all, the marker would never clear, permanently locking that branch out of ALL future refreshes.

Replaced with a generation counter per branch Sprite: each call takes a ticket number and checks — after writing its temp file, right before the risky `mv` — whether a NEWER call has since started for the same branch. If so, it abandons its own stale attempt instead of committing it. No lock to release, no backstop timer, no permanent-lockout risk.

Added a test that directly reproduces the race: an overlapping call with a manually-held-open root resolution (simulating a stall) racing against a normal call with a rotated token — proves the stalled call detects the newer generation and skips its mv rather than clobbering it.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +207,187 @@ async function cloneAndCheckoutBranch({
...
+          // wrong-permission or briefly absent in between. If anything
+          // fails before the rename, the live file at `path` is completely
+          // untouched.
+          // Suffixed with THIS call's own generation number — never shared
```
2witstudios: Confirmed and fixed in a2e417aaf336e40f3827a10c97542fcb2c1bbd6a — this was a real gap in the generation-counter fix.

You're right: both overlapping calls were writing to the same fixed `<path>.tmp`. If an older call's stale-cleanup ran after a newer call had already written its own content there (but before its own mv), the older cleanup would delete the newer call's temp file out from under it — making the CORRECT, newer refresh fail its own mv entirely, which is worse than the clobber being fixed.

Fixed by suffixing the temp path with the call's own generation number (`<path>.tmp.<generation>`) — no two overlapping calls ever share a temp filename, so one's cleanup can never disturb another's in-progress write. Still clears the temp path first regardless, since `generation` resets to 1 on every process restart and could coincidentally collide with a stale leftover from a pre-restart crash.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +208,209 @@ async function cloneAndCheckoutBranch({
...
+ * alone only cover branch creation and the navigator's explicit attach API,
+ * neither of which the realtime PTY path calls.
+ */
+export async function propagateClaudeCredential({
+  machineId,
+  branchHandle,
+  resolveRootMachineHandle,
+}: {
+  machineId: string;
+  branchHandle: MachineHandle;
```
2witstudios, line 320: Confirmed and fixed in ee3c2aa605b7cd4d72f510ef1854d002d321b2d9 — you're right that this is a real gap, and I want to be precise about what's fixed vs. accepted.

Fixed: the temp path now uses a globally-unique cuid2 id (`createId()`) instead of the process-local generation number — no two overlapping calls, from EITHER process, can ever share a temp filename, so one's cleanup can never delete another's in-progress write regardless of which process it's running in.

Accepted, documented explicitly: the staleness CHECK itself (`latestCredentialCopyGeneration`) is necessarily process-local — a call in `apps/web` genuinely cannot see an overlapping call in `apps/realtime`, so which one's `mv` lands last in a genuine cross-process race isn't something this in-memory counter can arbitrate. Fully solving that would need a database-backed lock or a version marker persisted on the Sprite's own filesystem — I judged that disproportionate for a background sync where every single refresh is already best-effort and self-heals on the next reattach in either process, rather than something that must be prevented outright. Happy to revisit if you think the cross-process ordering guarantee is worth the added complexity.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +209,224 @@ async function cloneAndCheckoutBranch({
...
+          // `apps/realtime`'s refreshBranchCredential — see the doc comment
+          // on `latestCredentialCopyGeneration`, which is process-local and
+          // therefore cannot be used to key the temp path). A fixed temp
+          // name was tried and reverted: when an older call's stale cleanup
+          // (below) ran after a newer, overlapping call had already written
+          // ITS OWN content to that same shared path, the older cleanup
+          // deleted the newer call's temp file out from under it, making
+          // the newer (correct) refresh fail its own `mv` (caught in
+          // review, twice — once for the intra-process case, once for the
+          // inter-process one this id fixes). A globally unique path means
```
2witstudios, line 381: Confirmed and fixed in f14f11fa6031ca37e4f47bb0dae5bd2f9faf3356.

Verified the claim: the Sprite runner's spawn path only installs its SIGKILL wall-clock timer when `timeoutMs` is explicitly supplied and positive — none of my `rm`/`mv` exec calls passed one, so on a cold or wedged Sprite they were genuinely unbounded at the transport level (the 5s `CREDENTIAL_COPY_TIMEOUT_MS` only stops `propagateClaudeCredential`'s caller from waiting, it doesn't touch the exec itself).

Added a small `housekeepingExecArgs` helper (`timeoutMs`: the same 5s bound; `maxBytes`: the existing `SANDBOX_MAX_OUTPUT_BYTES` constant already used elsewhere for sandbox commands) and applied it to all four rm/mv exec calls. Added a test asserting every housekeeping exec carries a positive `timeoutMs`/`maxBytes`.

packages/lib/src/services/machines/machine-branches.ts
```diff
@@ -189,6 +209,242 @@ async function cloneAndCheckoutBranch({
...
+          // A NEWER call already started for this branch Sprite while this
+          // one was working — it will (or already did) land more current
+          // data, so abandon this stale attempt rather than risk this one's
+          // `mv` clobbering it once it finally gets here.
+          //
+          // KNOWN, ACCEPTED RESIDUAL WINDOW (raised in review, twice): this
+          // check happens BEFORE the `mv` below, not atomically with it. A
+          // newer call could still start, read a rotated credential, and
+          // complete its OWN `mv` entirely within the time this call's own
+          // `mv` (checked as still-current a moment ago) takes to actually
```
2witstudios, line 412: Acknowledged — this is real, and I'm making a deliberate call not to chase it further. Documented in c748537dba63383b4192c13c8ddf993b4cf18d95.

The check happens before the mv, not atomically with it, so a newer overlapping call could in principle still land its mv after a check that was current a moment earlier. Fully closing this would need either an atomic compare-and-swap primitive this filesystem abstraction doesn't expose, or a synchronous per-branch mutex serializing every mv — which would then queue an unrelated caller's response behind however long a DIFFERENT in-flight refresh's Sprite I/O takes, reintroducing the unbounded-latency problem this whole multi-round design exists to avoid.

The impact here is staleness, not exposure: worst case the branch keeps using an old-but-still-valid token a little longer, and self-heals on the next reattach in either process — the same best-effort guarantee already documented everywhere else in this function. Given the residual window is now bounded to roughly one exec round-trip (down from the original tens-of-seconds unbounded-stall case), I judge this a reasonable place to stop rather than add a mutex that reintroduces latency risk to close an availability (not security) edge case. Added an explicit comment recording this as a deliberate tradeoff rather than an oversight. Happy to revisit if you disagree with that judgment call.

PR #2258, opened 2026-07-30
feat(agent-sessions): un-conflate session ≠ conversation, restore panes/console/agent-page surfaces on session ids

2witstudios:
Phase R0 of the corrective epic (`ta6prjapfdcrfjyg3wi7gsv8`). **A session is NOT a conversation.**

The first cut made them one object: `agent_sessions.conversationId` was the PRIMARY KEY and the Sprite name was folded from it. That is a cardinality error — a session and a conversation have different lifecycles and a genuine one-to-many relationship — and it forced one environment per chat thread, making shared working contexts structurally impossible (it is why panes could not share a sandbox).

## The model

```
Drive
 └─ Session "api refactor"            [owns the sandbox]
     ├─ conversation w/ Research Agent
     ├─ conversation w/ Refactor Agent
     └─ shell-1, shell-2
```

- `agent_sessions.id` — its own cuid PK. `conversationId` and `agentPageId` are GONE: the association runs the other way (`conversations.sessionId` FK, set at creation and **permanent** — moving a thread is a fork, never a rebind), and the agent belongs to each conversation, never to the session.
- `sessionKey = HMAC(tenant, sessionId)` folds the SESSION id, namespace bumped to **v2** (v1 folded conversation cuids in the same keyspace — reuse could let a v2 session derive a v1 Sprite's name). Sharing is **by construction**: every conversation and shell in a session resolves the same row, so their tool calls land in one sandbox with no shared id threaded anywhere.
- `conversations.sessionId` is ON DELETE SET NULL: ending a session releases compute, never erases history.
- Access is **drive access** (owner/accepted member; global-assistant sessions owner-only; unknown denies). The END check keeps its no-capability rule and gains an owner short-circuit: an owner removed from a drive loses *use* of the session but keeps the power to *stop paying* for it.
- A thread with no session gets a typed `no_session` **denial** from the tool layer — never a lazily-minted per-conversation environment.
- Workers (`spawn_session`) now work in their **spawner's** workspace: same session, same sandbox, same filesystem. `kill_session` aborts the run and no longer tears down a sandbox the worker never owned.
- PTY/storage billing attribution moves from page to **drive** (payer = drive owner ?? session owner); usage carries no `pageId` — a session is not page-anchored, and pretending otherwise was a semantic lie.

## The payoff test

**Two conversations bound to one session resolve ONE sandboxId** — proven at the store contract, against real Postgres (18 integration tests), and at the tool layer (`acquireSandbox`).

## Migrations (unreleased ⇒ drop-and-recreate, no data migration)

`0235` custom clear (DELETE, not TRUNCATE — the AFTER-DELETE trigger rescues live Sprite pointers into the reclaim outbox) → `0236` generated DROP → `0237` generated CREATE (drizzle-kit's ALTER path emits broken SQL for PK swaps: ADD PRIMARY KEY beside the old PK, FKs before their UNIQUE) → `0238` custom trigger re-arm.

Proven on real Postgres both ways: fresh DB end-to-end, and a head-state DB where the clear rescued exactly the live pointer (torn-down and never-provisioned correctly skipped) and the re-armed trigger rescued through both direct delete and owner cascade. Cascade tests inverted and pinned: a **page delete leaves sessions alive**; a **conversation delete leaves the session alive**.

## The restored surfaces (R1–R3 — #2259, merged into this branch)

#2259 stacked on this PR and has been merged in, so this branch now carries the full corrective arc:

- **Pane system restored on session ids** (`stores/agent-workspace/`, `components/agents/panes/`): columns-of-panes reducer, split-right/split-down/close, pane picker offering *an agent conversation in this session* or a shell. Closing the last pane ends the session (store-level `closePaneIn` semantics — the reducer no-ops, the store ends the workspace).
- **Console** (`AgentsSurface` + `AgentPanes`): selection lives in the query string (`?session=&c=&agent=`), the centre is the session's pane grid keyed by session id, and the grid receives the **session's** drive (fetched per session), not the surface's.
- **Agent pages**: Chat | History | Settings restored as real tabs (History full-height again, Save pinned in the header); the Chat tab is the same pane grid, so splitting on an agent page shares the page conversation's session sandbox. Grid gated on `canUseSessions`; read-only viewers get history without send/edit/delete/retry affordances.
- **Settings-driven sandbox** (`sandboxEnabled` on pages, migration `0239`): a Sandbox switch on agent settings; provisioning stays lazy on first tool use; sandbox/git/session tools hidden from Default Tools while off (`canRunCode` remains the non-UI security boundary).

## Review-hardening pass (from the PR review cycle)

- **Congenital binding**: `conversations.sessionId` + `title` ride inside the creators' INSERTs; no UPDATE path exists, so rebinding is unrepresentable. `createConversation` returns `'created' | 'exists' | 'message_owner_conflict'` with RETURNING-based race detection.
- **Cross-drive gate**: `createConversationInSessionWith` enforces `agent.driveId === session.driveId` centrally (fail-closed); global sessions host only assistant threads.
- **One id namespace per tool family**: worker listing/addressing uses conversation ids; shell rows store session ids; `findOwnWorkspace(conversationId)` resolves once.
- **Billing attribution**: `SandboxAcquireResult.sessionId` is required on success — storage measurement and PTY activity are keyed by the *acquired* session, never the conversation.
- **Bounded listings + quota**: session lists are active-only, ordered by recency, capped at 100; spawn refuses at 100 active sessions per owner (429); collection GET uses two bulk queries.
- **End access**: owner short-circuit (release-of-compute); everyone else faces the full decision with real `canRunCode`.

## Validation

- typecheck 0 errors monorepo-wide (including test files under the default tsconfig, matching CI); lint clean.
- packages/lib: 8,600+ unit + 21 real-Postgres integration tests (store list/ordering/countActive semantics verified against a live DB). realtime: 937. web touched suites green (the only local failures are the documented env-only ones: no-test-DB activity-tools, TZ-sensitive grouping).
- Sprite-key v2 pinned with a known-answer digest computed independently of the function; mutation-verified (reverting to v1 fails exactly that pin).

R4 (E2E coverage) remains tracked on the epic (`ta6prjapfdcrfjyg3wi7gsv8`).

https://claude.ai/code/session_017V3eqwRX5cFy3Tdu2Syoro

## Summary by CodeRabbit

* **New Features**
  * Added session-based agent workspaces with conversations and terminals sharing a single sandbox.
  * Added flexible pane layouts, including splitting, closing, resizing, and switching between chats and shells.
  * Added session creation, conversation creation, and session-aware navigation in the Agents sidebar.
  * Added a Sandbox setting for agents to control sandbox tool availability.
* **Updates**
  * Improved drive-level access control and billing attribution.
  * Added clearer handling for unavailable or unauthorized sessions and conversations.
  * Updated agent URLs and conversation data to include session context.


apps/web/src/lib/ai/tools/session-tools.ts
2witstudios, line 1: **H2 (HIGH) — `send_shell` / `read_shell` / `kill_shell` can never address a real shell.**

`openOwnShell` guards `shell.sessionId !== context.conversationId` — but post-unconflation `shell.sessionId` is an `agent_sessions.id` (workspace id: `spawnShell` resolves conversation→session and stores `session.id`) while the context carries the conversation id. Different namespaces, never equal: `send_shell`/`read_shell` always answer "not your shell", and `kill_shell` (same comparison ~line 690) always answers "already gone". The agent can spawn shells it can never drive.

The unit tests mask this: the fake `findShell` in `session-tools.test.ts:45` returns `sessionId: CALLER_CONVERSATION` — the old `sessionId ≡ conversationId` model living on in the fixtures.

**Fix shape:** resolve the caller's session once (`findSessionForConversation(context.conversationId)`) and compare `shell.sessionId` to `session.id`; update the fakes to the two-namespace shape, and add one integration test through the real store.

apps/web/src/lib/ai/tools/session-tools-runtime.ts
```diff
@@ -246,28 +246,16 @@ async function dispatchThroughChatPipeline(input: {
...
-      ? []
-      : (
-          await db
-            .select({ id: pages.id, title: pages.title })
-            .from(pages)
-            .where(inArray(pages.id, agentIds))
-        ).map((page) => [page.id, page.title]),
-  );
 
   return Promise.all(
```
2witstudios: **H2b (HIGH, same family) — `list_sessions` returns ids the other session verbs cannot address.**

`listSessionsForOwner` lists workspace rows (`sessionId: row.id` = `agent_sessions.id`), while `findSession` — the dep behind `send_session`/`read_session`/`kill_session` — resolves its argument as a **worker conversation id** (`conversationRepository.getConversation`). Yet the `list_sessions` tool description tells the model the listed ids are "the address every other session/shell tool takes", and the refusal copy says "Call list_sessions to see yours". A model following the tools' own documentation feeds workspace ids to verbs that look up conversations — every call dead-ends.

Decide which namespace the tool family addresses (worker conversations, per spawn's return value, seems right), make `list_sessions` list *that*, and fix the descriptions.

packages/lib/src/agent-sessions/decide-session-access.ts
```diff
@@ -112,22 +99,27 @@ export function decideAgentSessionAccess({
...
+ * true and the owner short-circuited, so the deciders cannot drift on the
+ * gate they share.
  */
 export function decideAgentSessionEndAccess(
   input: DecideAgentSessionEndAccessInput,
 ): AgentSessionAccessDecision {
+  if (input.requesterId.length > 0 && input.requesterId === input.session.ownerId) {
+    return { allowed: true };
+  }
   return decideAgentSessionAccess({ ...input, canRunCode: true });
```
2witstudios: **H3 (MEDIUM-HIGH) — any accepted drive member can destroy any session in the drive.**

The owner short-circuit (line 121) already serves the documented intent ("an owner who lost `canRunCode` must still release compute"). But the fallthrough pins `canRunCode: true` for **every non-owner**, so an accepted drive member with zero code-execution rights can `DELETE /api/agent-sessions/[id]` (Sprite killed, filesystem lost on re-provision) and `DELETE .../shells/[shellId]` (kills a live PTY) for other members' sessions. Session ids are obtainable by exactly this population — the page-agents conversations listing returns `sessionId` for shared conversations.

Compounding: the page view renders the pane grid for non-admins on shared session-bound conversations, and the pane-X close path fires the DELETE with **no confirmation** (the sidebar's end button has an AlertDialog; the pane close does not). A user who cannot *use* a session can destroy one by closing a panel.

**Fix shape:** keep the owner short-circuit; give non-owners the *real* capability check (or restrict end/kill to owner + drive owner/admin).

packages/lib/src/services/agent-sessions/agent-sessions-store.ts
```diff
@@ -240,56 +247,61 @@ export function revivedAgentSessionColumns(input: {
...
-        .onConflictDoNothing({ target: agentSessions.conversationId });
+        .returning();
+      return row as AgentSessionRecord;
     },
 
     async list(filter) {
       const conditions = [];
-      if ('agentPageId' in filter) conditions.push(eq(agentSessions.agentPageId, filter.agentPageId));
       if (filter.ownerId !== undefined) conditions.push(eq(agentSessions.ownerId, filter.ownerId));
       if ('driveId' in filter) {
```
2witstudios, line 319: **M3/M4 (MEDIUM) — `list()` is unbounded, unordered, and never excludes ended sessions.**

No `LIMIT`, no `ORDER BY`, no `endedAt` filter — and rows are retained forever by design. Downstream:
- The sidebar claims "the session leaves the sidebar" after ending, but the 20s poll brings the ended row right back; ended sessions accumulate forever with no remove/resume affordance.
- The collection GET does 1+2N queries (`listShells` + `listSessionConversations` per session) per poll per admin tab, and every admin page-conversation now spawns a session — at 100+ sessions that's ~200 queries per poll and a nondeterministically ordered list (rows can jump between polls after any UPDATE).

**Fix shape:** filter (or segregate) ended sessions, add `ORDER BY lastActiveAt DESC LIMIT n`, and batch the children (one query each across session ids) or denormalize counts.

apps/web/src/lib/ai/tools/session-tools.ts
2witstudios: Fixed in `8e7fae707`. New dep `findOwnWorkspace(conversationId)` resolves the caller's workspace once; `openOwnShell` and `kill_shell` compare `shell.sessionId` against the workspace id — one namespace. The fixtures now use a workspace id **distinct** from the conversation id (the exact shape that masked this), so every shell test exercises the real two-namespace comparison, plus three regression tests: own-shell reach succeeds, a sessionless caller is refused, a foreign workspace's shell reads as already-gone and is never killed.

apps/web/src/lib/ai/tools/session-tools-runtime.ts
2witstudios: Fixed in `8e7fae707`. `list_sessions` no longer enumerates workspace-row ids: it lists the caller's session's **workers by conversation id** — the exact address `send_session`/`read_session`/`kill_session` take — plus shells and the one shared sandbox's status, with the promised agent labels actually delivered (join to `pages`) and an `isCaller` marker. A sessionless conversation gets an explanatory note instead of a silently empty list. Tool description and module docblock rewritten off the dead `sessionId ≡ conversationId` model.

packages/lib/src/agent-sessions/decide-session-access.ts
2witstudios: Fixed in `8e7fae707`. The owner short-circuit stays (release-of-compute is the owner's emergency exit — no membership, no capability needed to stop paying). The fallthrough no longer pins `canRunCode: true`: non-owners get the full decision with the REAL capability, gathered by `checkAgentSessionEndAccess` exactly like the main check (the `Omit<'canRunCode'>` deps carve-out is gone). A drive member without code-execution rights now gets `code_execution_denied` on DELETE session/shell. Pure + service tests updated to pin both directions.

apps/web/src/lib/ai/tools/session-tools-runtime.ts
```diff
@@ -413,7 +424,11 @@ export function buildSessionToolsDeps(): SessionToolsDeps {
     },
 
     spawnShell: async ({ sessionId, ownerId, name }) => {
-      const spawned = await spawnShell({ sessionId, ownerId, name });
+      // The pure layer hands the CALLER's conversation id; shells hang off the
+      // SESSION, so resolve the working context first.
+      const session = await findSessionForConversation(sessionId);
+      if (!session) return { ok: false, reason: 'no_session' };
+      const spawned = await spawnShell({ sessionId: session.id, ownerId, name });
```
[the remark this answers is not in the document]
2witstudios: Fixed in `8e7fae707`: new `findOwnWorkspace(conversationId)` dep resolves the caller's workspace once, and `openOwnShell`/`kill_shell` compare `shell.sessionId` against the workspace id — one namespace. Fixtures now use a workspace id distinct from the conversation id, with regression tests for own-shell reach, sessionless refusal, and foreign-workspace already-gone. Leaving this thread open for your verification.

apps/web/src/lib/ai/tools/session-tools-runtime.ts
```diff
@@ -246,28 +246,16 @@ async function dispatchThroughChatPipeline(input: {
...
-            .select({ id: pages.id, title: pages.title })
-            .from(pages)
-            .where(inArray(pages.id, agentIds))
-        ).map((page) => [page.id, page.title]),
-  );
 
   return Promise.all(
     rows.map(async (row) => ({
-      sessionId: row.conversationId,
+      sessionId: row.id,
```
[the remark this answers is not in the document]
2witstudios: Fixed in `8e7fae707`: `list_sessions` now lists the caller's session's WORKERS by conversation id (the exact address `send_session`/`read_session`/`kill_session` take) plus shells and the shared sandbox status, with agent labels delivered and an `isCaller` marker — one address namespace per verb family, matching your suggested first option. Leaving open for your verification.

apps/web/src/lib/agent-sessions/agent-sessions-runtime.ts
```diff
@@ -280,31 +267,36 @@ async function ensureConversationRow({
...
+  } else {
+    await conversationRepository.createConversation(conversationId, userId, agentPageId);
   }
-  await conversationRepository.createConversation(conversationId, userId, agentPageId);
+  // The binding, set once at creation and permanent (contract invariant 1) —
+  // moving a thread to another session is a fork, never a rebind.
+  await db
+    .update(conversations)
+    .set({ sessionId })
+    .where(eq(conversations.id, conversationId));
```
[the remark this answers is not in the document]
2witstudios: Fixed in `8e7fae707`: the binding now travels INSIDE each creator's INSERT and no UPDATE of `conversations.sessionId` exists anywhere — an existing row is accepted only as an idempotent retry of itself (same owner, same session, same anchor); everything else throws `ConversationUnavailableError` (mapped to 409 on the routes in `2b78d455b`). Your secondary point is fixed by the same shape: create and bind are one INSERT, so no half-bound state can exist. Leaving open for your verification.

apps/web/src/lib/ai/tools/sandbox-tools-runtime.ts
```diff
@@ -109,29 +109,31 @@ export function buildRealSandboxRunDeps(): SandboxRunDeps {
...
-      // the exact discipline the machine path keyed by page id.
+      // The conversation's WORKING CONTEXT, through conversations.sessionId.
+      // A thread with no session gets a denial, never a lazily-minted
+      // environment — per-conversation minting is exactly the conflation the
+      // session model removed, and it is what made panes unable to share a
+      // sandbox. Every conversation in one session resolves this same row,
+      // whose own id folds the ONE Sprite key.
+      const row = await findSessionForConversation(conversationId);
+      if (!row) {
+        return { ok: false, reason: 'provision_failed', cause: 'no_session' };
```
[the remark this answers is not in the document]
2witstudios: Fixed in `c6d5164a9`: `SandboxAcquireResult`'s success arm now REQUIRES `sessionId` (both web acquirers return the resolved row id; the handle-bound git deps take it from the routes), `openSession` exposes it, and both post-run hooks — storage measurement and shell-activity notification — key on it instead of `ctx.conversationId`. This also un-gates the hooks from conversation presence, so sessionless-context runs measure and notify correctly. Fakes updated to a distinct workspace id; the two conversation-gated tests inverted into pins of the new keying. Leaving open for your verification.

apps/web/src/lib/ai/tools/session-tools-runtime.ts
```diff
@@ -375,9 +367,27 @@ export function buildSessionToolsDeps(): SessionToolsDeps {
...
+          conversationId: sessionId,
+          userId: ownerId,
+          agentPageId,
+          sessionId: callerSession.id,
+        });
+      } catch (error) {
+        return { ok: false, reason: 'conversation_unavailable', detail: error instanceof Error ? error.message : String(error) };
+      }
+      // `name` labels the worker CONVERSATION; the session keeps its own name.
+      void name;
```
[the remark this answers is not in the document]
2witstudios: Fixed in `c64990853`: the worker's `name` now travels through `createConversationInSession` into the creator's INSERT (`conversations.title`), so the sidebar, history and `list_sessions` show what the spawner named it. Labels only — nothing resolves by them. Leaving open for your verification.

apps/web/src/lib/ai/tools/session-tools-runtime.ts
```diff
@@ -375,9 +367,27 @@ export function buildSessionToolsDeps(): SessionToolsDeps {
...
+      if (!callerSession) {
+        return { ok: false, reason: 'no_session', detail: 'This conversation has no session for a worker to join.' };
+      }
+      try {
+        await createConversationInSession({
+          conversationId: sessionId,
+          userId: ownerId,
+          agentPageId,
+          sessionId: callerSession.id,
+        });
```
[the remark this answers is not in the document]
2witstudios: Fixed in `d2bcdba77`: the rule is enforced at the ONE binding path — `createConversationInSessionWith` now gates the page arm on `agent.driveId === session.driveId` (fail-closed on unresolved facts), so the worker-spawn path cannot bind a foreign drive's agent into the caller's workspace; a GLOBAL session hosts only assistant threads. Active-runtime billing therefore can no longer be pointed at another drive's owner via a cross-drive worker — the agent's drive and the session's drive are equal by construction. Leaving open for your verification.

packages/lib/src/services/agent-sessions/agent-sessions-store.ts
2witstudios: Fixed in `f6a94e22d` (store) + `f893fe8f4` (route): `list()` excludes ended rows, orders `lastActiveAt DESC NULLS LAST, createdAt DESC`, caps at 100 (integration-tested against real Postgres incl. the NULLS LAST semantics); new `countActive` feeds a spawn ceiling (429 at 100 active sessions); the collection GET batches children into two bulk queries (3 queries per poll total, was 1+2N); ended sessions now actually leave the sidebar, making its comment true.

Issue #2255, opened 2026-07-29
Agent-created shells never appear as tabs (no invalidation for spawn_shell/kill_shell)


2witstudios:
Found by Codex reviewing #2249 after it merged; verified here.

## Symptom

When the agent calls `spawn_shell`, the new shell never appears as a tab. When it calls `kill_shell`, the dead shell stays visible. Both persist until an unrelated revalidation or a remount.

## Why

`useSessionShells` keeps the tab list in one SWR entry:

```ts
const { data, error, isLoading, mutate } = useSWR<ShellsResponse>(key, fetcher, {
  revalidateOnFocus: false,
  dedupingInterval: 5_000,
});
```

No polling, no socket invalidation, and focus revalidation explicitly disabled. The cache is only ever updated by this hook's own `addShell`/`removeShell` — the human-driven paths.

The agent's `spawn_shell`/`kill_shell` run server-side through the session-tools runtime and mutate `agent_session_shells` directly, never touching that cache. So the client has no way to learn about them.

There's also no existing event to piggyback on: every shell socket event (`shell:output`, `shell:ready`, `shell:closed`, `shell:error`) is **connection-tagged**, not a room broadcast — by design, per the rebuild plan ("No `rooms.ts` change needed"). So nothing currently tells a client "the set of shells changed."

## Options

1. **Revalidate on agent turn completion.** Cheapest: one GET per completed turn, only while the view is open. Needs a small seam — `SessionChat` signalling turn-end up to `AgentView`, which already holds `shells.mutate`. Coarse (fires even when no shell tool ran) but simple and always correct.
2. **Watch the message stream for shell tool parts** and revalidate only then. Precise, more coupling to message-part shape.
3. **Broadcast a shell-lifecycle event** from realtime to the session's watchers. Most correct and would also cover a second browser tab, but it's the only option that adds a room/broadcast concept to a surface deliberately built without one.

(1) is the smallest thing that fixes the reported symptom; (3) is the one that generalises. I did not implement any of them because the choice is a real design call, not an obvious default — and #2253, where this was found, is scoped to unrelated fixes.

Raised from the same review pass as #2253 and #2254.

2witstudios:
Superseded: the shells-as-tabs surface was removed in the pane-grid restoration (pu/agents-restore-panes). Shells now render in panes whose binding is written at open time, so there is no tab list to invalidate.

PR #2259, opened 2026-07-30
feat(agents): restore panes + session surfaces (R1–R3 of the architecture correction)

2witstudios:
Stacked on #2258 (R0: session ≠ conversation un-conflation). Together they complete the corrective overhaul of the agents surface — epic `ta6prjapfdcrfjyg3wi7gsv8`, phases R1–R3.

## What this restores / builds

**R1 — pane system** (restored from `623e632e7^`, stripped of machine topology)
- Pure columns-of-panes reducer + PaneBar + SessionPanes + PanePicker + `resolvePaneSurface`, session-keyed persisted workspace store, `AgentPanes` container.
- Lifecycle invariant both directions: a session is born with its first conversation in its first pane; closing the LAST pane ends the session (store-level interception, container does the DELETE). Panes never appear in the sidebar.

**R2 — session surfaces**
- Sidebar rewritten to **Drive → Session → conversations**; selection is `?session=&c=&agent=` via pushState (no navigation, shells survive clicks).
- Spawn is one act: `POST /api/agent-sessions` mints the session AND its first conversation; the user lands inside it. No sandbox until first tool call/shell.
- Explicit **End session** on the row (confirmed; conversations remain as history — `conversations.sessionId` is ON DELETE SET NULL).
- **Global-assistant sessions**: `useAssistantSessionChat` (global pipeline sibling of the agent hook), one shared `SessionChatView`, assistant offered in the pane picker, Assistant group + one-click spawn in the sidebar, `POST /api/agent-sessions/[sessionId]/conversations` as the session-centric creator (the only way an assistant thread can join a session).

**R3 — agent page restoration** (from `e479b0053^`)
- Chat | History | Settings as real pills (`grid grid-cols-3 max-w-lg`, icons); History full-height tab (popover deleted); Save pinned in the header row.
- The Chat tab hosts the pane grid for session-bound conversations (splitting is the only new affordance); session users get session-born page conversations, everyone else keeps plain chat unchanged.
- Removals: AgentView + session-tabs + status chip + Add-shell + history popover (PR #2256 and issue #2255 closed as superseded).
- **Sandbox switch restored**: `pages.sandboxEnabled` (default off, successor to machineAccess), gating the bash/git/session tool families at request time and out of the Default Tools picker; provisioning stays lazy and automatic.

## Verification
- `bun run typecheck` ✅ · `bun run lint` ✅ · knip ratchet ✅
- `bun run test:unit`: 15,014 web tests pass; the 3 failing files are the pre-existing env-only bucket (2 need the test Postgres, 1 is TZ-sensitive and passes under UTC) — all green when run against a real DB / UTC.
- All 240 migrations (incl. new `0239 pages.sandboxEnabled`) replay clean on fresh Postgres 17; the agent-sessions store integration suite (18 tests, incl. the two-conversations-one-sandbox payoff) passes against it.
- `test:security`: 44/50; the 6 failures are stale `scripts/test-security.sh` paths (files excluded from the default config since April / deleted) — pre-existing on master, untouched here.
- Outstanding: the R4 live-app E2E checklists (spawn → provision → Sprite kill in a running deployment, mobile viewport pass) — tracked on the epic's R4 board.

https://claude.ai/code/session_017V3eqwRX5cFy3Tdu2Syoro


apps/web/src/lib/agent-sessions/agent-sessions-runtime.ts
2witstudios, line 1: **H1 (HIGH) — cross-user conversation hijack: the "permanent" binding is an unguarded UPDATE.**

The agent-page arm calls `conversationRepository.createConversation`, which **silently returns when the row already exists — owned by anyone** (`conversation-repository.ts` `if (existing) return;`), then control falls into the unconditional `UPDATE conversations SET sessionId`. Both entry routes (`/api/agent-sessions/[sessionId]/conversations` and `/api/ai/page-agents/[agentId]/conversations`) accept a client-supplied `conversationId` with no ownership or conversation↔agent check.

Exploit: an attacker with their own session POSTs a victim's conversation id (leaks via `?c=` URLs and shared-conversation listings, which include `sessionId`) → the victim's conversation is rebound into the attacker's session → the victim's next sandbox tool call resolves via `findSessionForConversation` and **executes inside the attacker's Sprite**; the victim's thread title also appears in the attacker's sidebar via `listSessionConversations` (no owner filter). The global-assistant arm is correctly guarded (`resolveOrCreateConversation` throws `ConversationOwnershipError`); only this page arm has the hole. Same-user rebinding also works, violating contract invariant 1 — the store's own docblock forbids exactly this second write path.

Sub-bug: when `hasConflictingMessageOwner` trips, `createConversation` returns silently, the UPDATE matches zero rows, and the route still answers **201 with a conversationId that has no row**.

**Fix shape:** make the binding write conditional — `.where(and(eq(conversations.id, id), eq(conversations.userId, caller), isNull(conversations.sessionId)))` — and fail loudly on zero rows; add a real-store test (route tests mock this function, which is why nothing caught it).

Prod blast radius is contained by `canRunCode` (app admin), but note `can-run-code.ts`'s admin rung is `NODE_ENV === 'production'`-only — any non-prod deploy with `CODE_EXECUTION_ENABLED` exposes this to every authenticated user.

apps/web/src/components/agents/panes/AgentPanes.tsx
```diff
@@ -0,0 +1,242 @@
...
+  // session offers no drive agents (there is no drive to list).
+  const { allAgents, isLoading: agentsLoading } = usePageAgents(driveId ?? undefined, {
+    enabled: driveId !== null,
+  });
+  const pickableAgents: PickableAgent[] = (allAgents ?? [])
+    .filter((agent) => agent.driveId === driveId)
+    .map((agent) => ({ id: agent.id, title: agent.title ?? 'Agent' }));
+
+  const workspace = workspaces[sessionId];
+  if (!workspace) {
```
2witstudios: **M1 (MEDIUM) — sidebar selection cannot change an existing grid.**

`initialConversation` is consumed only by `ensureWorkspace`, which is idempotent — it seeds a *fresh* workspace and no-ops on a persisted one. Consequences, all confirmed by code path:
- Clicking a conversation in the sidebar for a session whose persisted layout doesn't already show it: URL + highlight update, `AgentPanes` doesn't remount (`key={selectedSessionId}` unchanged), and **nothing visibly happens**.
- The sidebar's per-session "New conversation" mints a real server row no pane will ever display.
- Session-row click "opens on its most recent conversation" — same hole.

`paneShowing` (the focus-rather-than-duplicate helper in `pane-reducer.ts:214`) is exported and used nowhere — it's the missing piece. **Fix shape:** on selection, if `paneShowing(conversation)` exists focus it, else assign it into the active pane (or a new split). This is what makes the sidebar's core click paths real (epic spec: "selecting a conversation opens it").

apps/web/src/components/agents/AgentPageView.tsx
```diff
@@ -153,83 +180,162 @@ export default function AgentPageView({ page }: AgentPageViewProps) {
...
-            <PageAgentSettingsTab
-              ref={agentSettingsRef}
-              pageId={page.id}
+          </div>
+        </div>
+
+        {/* Chat Tab — the pane grid for a session-bound conversation, the plain
+            chat for a pre-session one. */}
+        <TabsContent value="chat" className="mt-0 flex min-h-0 flex-1 flex-col overflow-hidden">
+          {current.sessionId ? (
```
2witstudios: **M2 (MEDIUM) — non-admins get the pane grid via shared session-bound conversations, and the grid skips read-only gating.**

This branch keys on `current.sessionId`, not `canUseSessions` — and `useResolvedConversation` resolves the most-recent conversation *including its sessionId* regardless of role (shared conversations included). A non-admin then gets a grid whose every affordance 403s as toasts — except closing the last pane, which **succeeds destructively** (end-access waives the capability gate; see the finding on `decide-session-access.ts` in #2258).

Also: `PaneChat` never forwards `isReadOnly`, so the grid path skips the read-only gating the plain-chat path applies (`SessionChat`'s console context also passes `handleEdit/Delete/Retry` unconditionally).

**Fix shape:** gate the grid branch on `canUseSessions` (fall back to plain `SessionChat` otherwise), and thread `isReadOnly` through `AgentPanes` → `PaneChat`.

apps/web/src/components/agents/AgentsSurface.tsx
```diff
@@ -41,35 +45,37 @@ export default function AgentsSurface({ driveId }: { driveId?: string }) {
...
     <div className="flex h-full flex-col">
-      {selectedAgentId && selectedConversationId ? (
-        <AgentView key={selectedConversationId} agentId={selectedAgentId} conversationId={selectedConversationId} context="console" />
-      ) : selectedConversationId || selectedAgentId ? (
-        <SelectionPlaceholder />
+      {selectedSessionId && selectedConversationId ? (
+        <AgentPanes
+          key={selectedSessionId}
+          sessionId={selectedSessionId}
+          driveId={storeDriveId}
```
2witstudios: **M5 (MEDIUM) — global console can't split to drive agents.**

`driveId={storeDriveId}` passes the *surface's* drive — null on `/dashboard/agents` — but `AgentPanes` documents this prop as "the session's drive". In global mode every drive session's picker therefore offers only Shell/Assistant; the session's own `driveId` is right there in the sidebar's `SessionListEntry` but never carried through selection. Thread the selected session's driveId through the store (or look it up from the session list) instead of the surface's.

apps/web/src/app/api/agent-sessions/route.ts
```diff
@@ -67,3 +82,114 @@ export async function GET(request: Request) {
...
+    );
+  }
+
+  const access = await checkAccessForSubject(auth.userId, {
+    sessionId: 'about-to-be-minted',
+    ownerId: auth.userId,
+    driveId,
+  });
+  if (!access.allowed) {
+    auditRequest(request, {
```
2witstudios, line 209: **M6 (MEDIUM) — spawn accepts an arbitrary `agentPageId`; no quota on spawn.**

The spawn POST validates only the both-present/both-absent shape, then `createConversationInSession` writes `type:'page', contextId: agentPageId` with **no `getAiAgent` and no `canPrincipalViewPage`** — unlike both sibling routes, which do both. A caller can mint conversation rows anchored to a non-AI_CHAT page, a trashed page, or an agent in a drive they cannot see (impact contained — chat routes re-check per turn — but it's a gate the design says exists).

Also: spawn is deliberately "instant and free", so the tier ceiling (live-sandbox concurrency, enforced at provision) never applies — an authorized caller can loop this endpoint and mint unbounded `agent_sessions` + `conversations` rows. `name` is trimmed but unbounded. Suggest: agent-page existence/permission check + a row-count or rate ceiling + `name` length cap.

packages/lib/src/agent-sessions/contract.ts
```diff
@@ -48,16 +48,61 @@ export const sandboxStatusSchema = z.enum(SANDBOX_STATUSES);
...
 
 export const shellAgentTypeSchema = z.enum(SHELL_AGENT_TYPES);
 export type ShellAgentType = z.infer<typeof shellAgentTypeSchema>;
 
+/**
+ * What a pane displays. The discriminator lives on the PANE BINDING rather than
+ * on either row type, because the two surfaces are now two different tables and
+ * a pane is the one place that has to talk about both.
+ *
+ * `'chat'` addresses a conversation (`sessionId` ≡ `conversationId`);
```
2witstudios: **Doc rot (LOW, but load-bearing) — this file contradicts its own invariant 1.**

Line 77 still says "`'chat'` addresses a conversation (`sessionId` ≡ `conversationId`)" two paragraphs after invariant 1 establishes the opposite. The same stale claim survives in `useAgentSessionChat.ts`, `session-tools.ts`, `tool-filtering.ts:137` ("a conversation IS a session"), `build-session-chat-request.ts`, `pane-reducer.ts`, `git-tool-runners.ts`, `sandbox-storage-measure.ts` — and, worst, `system-prompt.ts:93` tells the **model** "this conversation IS a session with its own sandbox". One mechanical sweep should retire the old model's vocabulary everywhere; the contract file is the place reviewers will trust first.

apps/web/src/lib/ai/shared/hooks/useConversations.ts
```diff
@@ -149,6 +149,7 @@ export function useConversations({
         preview: '',
         isShared: false,
         isOwner: true,
+        sessionId: null,
```
2witstudios: **LOW — optimistic history entries hardcode `sessionId: null`.**

A just-created session-bound thread selected from History renders as a plain chat (no grid) until the server refetch supplies the real `sessionId`. Thread the known sessionId into the optimistic entry where the creator has it.

apps/web/src/components/layout/left-sidebar/AgentsSidebar.tsx
```diff
@@ -174,164 +194,257 @@ function resolveListNotice({
...
-  const { conversations, isLoading, createConversation } = useConversations({
-    agentId: agent.id,
-    currentConversationId: selectedConversationId,
-    enabled: expanded,
-    onConversationCreate,
-  });
+  const openSession = useCallback(() => {
+    // Selecting a SESSION opens its most recent conversation — the row is a
+    // workspace, and a workspace opens on its work, not on a placeholder.
+    setExpanded(true);
```
2witstudios, line 311: **LOW — zero-conversation session rows can never be opened; 'starting' status is unreachable.**

`openSession` selects `conversations[0]` and does nothing when the list is empty — the row expands but can never become selected, so its grid never opens (the "No conversations" empty state a few lines down proves the shape exists). Separately, the `'starting'` sandboxStatus this component checks is documented as "shown while the client's own ensure request is in flight" — but no UI calls the explicit provision endpoint (`POST /api/agent-sessions/[sessionId]`) anymore, so the status can never occur. Related orphaned endpoints with zero UI callers after the hook deletions: `GET/POST /api/agent-sessions/[sessionId]`, `GET .../shells`, `DELETE .../shells/[shellId]` (closed terminal panes leave PTYs running with no way to reattach or kill), and the whole `diff`/`git-blob`/`files` family.

apps/web/src/lib/agent-sessions/agent-sessions-runtime.ts
2witstudios: Fixed in `8e7fae707` (base) + `2b78d455b` (this PR's merge + 409 mapping). The binding now travels INSIDE each creator's INSERT and **no UPDATE of `conversations.sessionId` exists anywhere** — rebinding is unrepresentable rather than refused:
- `conversationRepository.createConversation` gains `{sessionId}` and returns `'created' | 'exists' | 'message_owner_conflict'`;
- `resolveOrCreateConversation` gains `{sessionId}` with an existing-row match check (`ConversationBindingConflictError` on mismatch);
- the decision logic moved to the pure module `create-conversation-in-session.ts`: an existing row is accepted ONLY as an idempotent retry of itself (same owner, same session, same anchor) — foreign rows, unbound rows, other sessions' rows, and legacy conflicts all throw `ConversationUnavailableError`, which both routes map to 409 + an `authz.access.denied` audit (fixing the victim-side-invisible-audit gap too).

23 new tests pin the hijack, the same-user rebind, binding-from-null adoption, and the 201-with-no-row sub-bug.

apps/web/src/components/agents/AgentPageView.tsx
```diff
@@ -45,55 +69,28 @@ export interface AgentPageViewProps {
...
+    canUseSessions,
+  });
+  const [override, setOverride] = useState<ResolvedConversation | null>(null);
+  // The conversation on screen: the user's own switching (history select, new,
+  // delete-replacement) wins over the initial resolution.
+  const current = override ?? initialResolved;
+
+  const [activeTab, setActiveTab] = useState<string>('chat');
   const [webhooksOpen, setWebhooksOpen] = useState(false);
+  const [agentConfig, setAgentConfig] = useState<AgentConfig | null>(null);
```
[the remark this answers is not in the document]
2witstudios: Fixed in `7ca594cdd`: the abortable `/api/pages/[id]/agent-config` fetch effect is restored, with a regression test pinning that the Settings tab receives config (mock asserts `data-has-config="true"` after the fetch). Leaving open for your verification.

apps/web/src/app/api/ai/page-agents/[agentId]/conversations/route.ts
```diff
@@ -191,9 +195,28 @@ export async function POST(
...
 
+    // Optional session binding — the thread is BORN into its working context
+    // (contract invariant 1: set once at creation, permanent; a session hosts
+    // many conversations and owns the one sandbox they share). Gated on the
+    // session access check so a caller cannot bind a thread into a workspace
+    // they cannot reach.
+    const sessionId: string | null =
+      typeof body.sessionId === 'string' && body.sessionId.length > 0 ? body.sessionId : null;
+    if (sessionId !== null) {
+      const sessionAccess = await checkSessionAccess(auth.userId, sessionId);
```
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/ai/page-agents/[agentId]/conversations/route.ts:207]: Fixed in `d2bcdba77` (+`ca3689db9` for the spawn route): the drive equality check lives at the CENTRAL binding path (`createConversationInSessionWith`), so both conversation routes AND the worker-spawn tool get it without per-site checks — `AgentNotInSessionDriveError`, fail-closed on unresolved facts, global sessions host only assistant threads. Leaving open for your verification.

apps/web/src/app/api/agent-sessions/route.ts
```diff
@@ -67,3 +82,114 @@ export async function GET(request: Request) {
...
+      { error: 'A session needs a drive and an agent to start with, or neither for the assistant' },
+      { status: 400 },
+    );
+  }
+
+  const access = await checkAccessForSubject(auth.userId, {
+    sessionId: 'about-to-be-minted',
+    ownerId: auth.userId,
+    driveId,
+  });
```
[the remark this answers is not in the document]
2witstudios: Fixed in `ca3689db9`: the spawn shape now runs `getAiAgent` (404 for missing/trashed/non-agent pages), an explicit `agent.driveId === driveId` check (400), and `canPrincipalViewPage` (403 + denial audit) BEFORE minting anything — plus the central binding gate from `d2bcdba77` as defense in depth. Session names are also trimmed and capped at 120 chars. Leaving open for your verification.

apps/web/src/app/api/agent-sessions/route.ts
2witstudios: Fixed in `d2bcdba77` + `ca3689db9`: spawn validates the agent (getAiAgent 404, cross-drive 400, canPrincipalViewPage 403 + audit) before minting; the central binding gate enforces drive equality for every other site; names trimmed + capped at 120. Remaining from this thread: a spawn-rate/row ceiling — tracked with the M3/M4 limits work (store-level bounds), next up.

apps/web/src/components/layout/left-sidebar/AgentsSidebar.tsx
```diff
@@ -174,164 +194,257 @@ function resolveListNotice({
...
+    setEnding(true);
+    try {
+      await del(`/api/agent-sessions/${encodeURIComponent(session.sessionId)}`);
+      // The session leaves the sidebar; its conversations remain as history in
+      // each agent's list. Drop the local grid too — its panes pointed at a
+      // sandbox that no longer exists.
+      forgetWorkspace(session.sessionId);
+      if (selectedSessionId === session.sessionId) selectSession(null);
+      setConfirmingEnd(false);
+      onChanged();
```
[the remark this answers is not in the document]
2witstudios: Fixed in `f6a94e22d`: `store.list()` now excludes `endedAt IS NOT NULL` rows (plus newest-first ordering and a 100-row cap), so the refetch after DELETE no longer resurrects the ended session — the handler's expectation is now the store's behavior. Ended rows remain for lifecycle/billing but are not listings. Leaving open for your verification.

apps/web/src/app/api/agent-sessions/route.ts
2witstudios: Quota half landed in `f893fe8f4`: `countActiveSessionsForOwner` + a 429 ceiling at 100 active sessions through the shared quota response (audited), alongside the earlier agent validation + name cap. Thread fully addressed.

apps/web/src/components/agents/panes/AgentPanes.tsx
2witstudios: Fixed in `5dc27aeee`: new store action `openConversation` — seeds a fresh grid, FOCUSES the pane already showing the thread (via the previously-unused `paneShowing`), otherwise opens it in the active/first non-terminal pane (never over a running PTY; splits right when every pane is a terminal). `AgentPanes` drives it from an effect on `(sessionId, conversationId)`, so mount seeding and later selections share one path — the sidebar's row/conversation/New-conversation clicks and the page's History picks now visibly open the thread. Five store tests pin the policy.

apps/web/src/components/agents/panes/AgentPanes.tsx
```diff
@@ -0,0 +1,242 @@
...
+    ensureWorkspace(sessionId, {
+      kind: 'chat',
+      name: initialConversation.name,
+      targetId: initialConversation.conversationId,
+      agentPageId: initialConversation.agentPageId,
+    });
```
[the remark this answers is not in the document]
2witstudios: Fixed in `5dc27aeee`: new store action `openConversation` — seeds a fresh grid, FOCUSES the pane already showing the thread (via the previously-unused `paneShowing`), otherwise opens it in the active/first non-terminal pane (never over a running PTY; splits right when every pane is a terminal). `AgentPanes` drives it from an effect on `(sessionId, conversationId)`, so mount seeding and later selections share one path — the sidebar's row/conversation/New-conversation clicks and the page's History picks now visibly open the thread. Five store tests pin the policy. Leaving open for your verification.

apps/web/src/components/agents/AgentPageView.tsx
2witstudios: Fixed in `38cc21006`: the grid branch now requires `canUseSessions` (non-admins get the plain chat even for session-bound shared threads — no 403-shaped chrome, no destructive last-pane-close), and `isReadOnly` threads AgentPanes → PaneChat → SessionChat/AssistantSessionChat, with the console renderer withholding edit/delete/retry like the page renderer (SidebarMessagesContent handlers widened to optional; CompactMessageRenderer already hides affordances for absent handlers). Tests pin both.

apps/web/src/components/agents/AgentsSurface.tsx
2witstudios: Fixed in the latest push: AgentsSurface resolves the selected session's record (GET /api/agent-sessions/[sessionId] — that endpoint's first UI caller) and passes the SESSION's driveId to the grid, so drive sessions opened from the global console offer their drive's agents in the pane picker. Test pins it.

packages/lib/src/agent-sessions/contract.ts
2witstudios: Fixed in `ba065d8f9`: full sweep — contract.ts:77, the MODEL-FACING system-prompt bullet, tool-filtering, build-session-chat-request, pane-reducer/store comments, useAgentSessionChat/SessionChat, Shell/useResolvedAgent AgentView refs, git-tool-runners + sandbox-storage-measure ≡ notes, both route docs (incl. end-access updated for H3), PanePicker's canPickAssistant doc. session-anchor.ts (the canonical stale text) is deleted outright.

apps/web/src/lib/ai/shared/hooks/useConversations.ts
2witstudios: Fixed in `ba065d8f9`: OptimisticConversationEntry gains sessionId and the optimistic mapping threads it through, so a broadcast-created session thread renders its grid immediately when the broadcast knew the binding.

apps/web/src/components/layout/left-sidebar/AgentsSidebar.tsx
2witstudios: Addressed across the fixes: sessions can no longer present as zero-conversation dead rows in practice (a session is born with a conversation, ended sessions leave the listing, and openConversation makes selection land visibly); GET /api/agent-sessions/[sessionId] gained its first UI caller (the M5 drive resolution). Remaining by DESIGN, documented rather than removed: the explicit provision POST (resume-after-end entry — provisioning is otherwise lazy via first tool/shell use, which covers resume), the 'starting' status (client-transient for surfaces that do call provision), shells GET/DELETE and the diff/git-blob/files family (server surface for the file-browser story, pre-existing on master). Dead pending-prompt scaffolding deleted; running-badges deleted (the sandbox dot is the running signal).

PR #2371, opened 2026-08-07
feat(agents): one source of truth for agent sessions, messages, and pane layout

### LINKED ISSUE #2208: Layout-rearrange verbs for the session family (post entity-promotion)
From the Node Sandboxes epic (PR #2204), deferred by design.

The session family shipped `add_session` (placement: new-view | split-into), `move_session` (re-home within a node), and `kill_session` — but no general layout-rearrange verbs (resize, reorder panes within a view, reorder views, rename view). Agents can manage the human's workspace only at session granularity.

**Blocked on / lands with #2202 (entity promotion):** rearrange verbs as blob writes would add yet another writer of the client-authored layout JSONB. Once workspaces/panes are relational entities with verb APIs, rearrange verbs are just more verbs — implement them there, not before.

Refs: `apps/web/src/lib/ai/tools/session-layout.ts` (placement writers), #2202.

2witstudios:
Collapses the agent-session data model to one source of truth per fact. 26 merged PRs, 530 files, 9 migrations, 156 commits — each PR individually green and individually revertable.

## ⚠️ Read before deploying — this cannot ship as a single rolling deploy

Migrations run as a one-off machine **before** the rolling deploy, so for the length of the roll old pods would query tables this branch drops (`chat_messages`, `agent_sessions.workspaceState`). Page chat and workspace endpoints return 5xx for that window. Fail-closed, not lossy — but user-visible.

**Two safe options:**
1. **Two releases** — ship everything through `0251_messages_unification_validate_fk` first, let it soak, then ship `0252`–`0254` (the contract steps). This is what the plan was designed for.
2. **One release behind a maintenance window.**

`infrastructure/UPGRADE.md` carries minimum-upgrade-path notes; version-skipping tenant/onprem operators must not skip this release.

**Operator step, non-automatable:** `0253_drop_chat_messages` is the only irreversible migration in the epic. Verify a current snapshot exists before it applies (procedure documented in the PR that introduced it — verify today's scheduled dump plus a Fly volume snapshot).

## The problem

State was stored in several places with no authoritative feed, so every surface kept a private copy and its own theory of when to refresh. That produced: panes showing nothing while the DB held the messages, workers appearing on a 20-second poll tick, agents unable to observe the effects of their own actions, and five different meanings of `sessionId`.

## What changed

**One writer, one feed.** Every durable message write goes through a repository choke point that bumps `conversations.rev` and emits in-transaction. Panes join per-conversation Socket.IO rooms with authorization at join, hold a rev watermark, and refetch on any gap. Delivery stays best-effort; correctness comes from detectability.

**One message table.** `chat_messages` merged into `messages` with a real FK (the legacy table never had one). Orphan conversations synthesised under a locked ownership rule; unresolvable ones skipped and reported rather than guessed.

**One pane grid.** Relational rows behind idempotent verbs and a rev counter, replacing a jsonb blob, a localStorage copy, and a hydration latch. Agents can now arrange their own workspaces (closes #2208).

**One vocabulary.** `agent_sessions` → `agent_workspaces` throughout schema, modules and routes, with one-release compat views and alias routes. The model-facing tool contract is deliberately frozen and snapshot-pinned.

**One chat pipeline** behind both routes; the `agentPageId === null` dispatch branch is gone.

## Verification

- `bun run scripts/check-evidence-manifest.ts` → **15 guarantees, 15 proven, 0 gaps.** Each citation must resolve to a non-skipped test in a suite CI actually runs; it fails on drift, and has already caught a stale citation in anger.
- **New `ci / E2E (agent-session user stories)` job** runs the user-story specs on every PR against a production build with production socket topology (realtime behind a proxy so `/socket.io` is same-origin — a cross-origin setup silently exhausts the browser's connection budget and specs then pass for the wrong reason).
- `test:security` 51/51 · full migration chain 0000→0254 clean on a fresh database · `drizzle-kit check` clean, no drift.
- Previously-dark surfaces now run in CI: `packages/db` integration suites, `infrastructure/` tests, `apps/e2e` typecheck + lint, the full security script.

## Found and fixed while doing this

Three adversarial reviews ran against the merged branch. Every finding is fixed and covered by a test that has been observed to fail:

- **A migration that aborted on real production data** — the 0249 pre-audit rejected `type='client'` threads spanning two agent pages, a shape the spec documents as supported.
- **A completeness guard that would have resurrected deleted messages** — its premise held for the freeze window but not the dual-write window; an edited message reverted and a user-deleted message came back live.
- **Three data-leak paths** — pane-label joins with no viewer (a universal title oracle over any page, conversation or shell), agent layout tools skipping the access check their own docblock claimed to run, and the new `conv:`/`session:` rooms absent from every revocation kick set.
- **Live delivery didn't actually work** — an async race in the socket store created ~9 connections where one was needed, exhausting the browser's six-per-origin HTTP/1.1 budget and starving both sockets and ordinary fetches.
- **Tenant export bundles that could not be imported** (FK-ordered inserts, orphan conversations), plus 18 missing columns including `users.emailBidx` — without which migrated accounts could not log in.
- **GDPR**: Article 15 export dropped agent-authored answers, and later the agent-workspace tables entirely — we were deleting on demand what we would not disclose on demand. Article 17 left `ai_stream_sessions` content, which nothing had ever deleted.
- **A type-blind conversation gate** — a global conversation "belonged to" every page.
- **Lossy pagination** — a `createdAt`-only cursor skipped or duplicated same-tick rows on page boundaries.
- **Five tests that passed while proving nothing**, including one whose harness meant a Service Worker swallowed the requests it was asserting on.
- **Every prose reference to the epic's migrations pointed one migration off** — the master integration renumbered 0246–0253 → 0247–0254 and only the SQL files and their test suites were updated. A renumber SHIFTS numbers rather than orphaning them, so all 74 stale citations resolved to a different REAL migration instead of failing loudly. Worst case was `infrastructure/UPGRADE.md`, which described `0253` as `ALTER … RENAME`, "nothing is dropped", "no row is touched" — while `0253` is now `0253_drop_chat_messages`, the one irreversible step, and the one this PR tells operators to snapshot before. The runbook and the deploy instructions contradicted each other on the only unrecoverable step. Also fixed in `docs/2.0-architecture/agent-sessions.md`, the DSR runbook, and 28 source files. Every citation was re-derived from the migration's own DDL rather than shifted by a fixed offset — two were already wrong before the renumber, so a blanket +1 would have preserved the error. UPGRADE.md now has a drift guard, observed to fail on the original text.

## Post-review remediation

A full read-only review ran against the branch (0 blockers / 2 majors / 3 minors / 1 nit). All six are fixed — `17bb897`, `fac799f`:

- **`paneScopesOfVerb` did not keep its exhaustiveness promise.** The docblock said a new scope-carrying verb would be a compile error; the switch ended in `default: return []`, so it would instead have been a silently ungated write path past `authorizeVerbScopes` — the write half of the security-review HIGH-1 fix. All twelve verbs are now named with a `never` fall-through, plus a test that reads the verb union at runtime and pins it.
- **Every migration's own `RAISE` text named it one number too low.** 0251's abort told an operator the blocking rows were "migration 0248's SKIPPED orphans" — but `0248_grey_sister_grimm` is a one-line `ADD COLUMN rev`; the synthesis that skipped them is 0249. Renumbered 59 tokens across six migrations, verified by applying them: all seven migration suites green against a real Postgres (97 tests), including the live scenarios that assert on the messages this SQL now emits.
- **Un-share never reached the page room.** The event was emitted from the post-update row, and the page room is only in the audience when the conversation is shared — so the collaborators who had just lost access were the only ones not told, and their sidebar stayed stale until the 120s backstop poll.
- Plus: a NOTICE naming a deliberately-deleted operator script, a §4 shim list shorter than the contract PR it describes, and a duplicate import.

The migration-message finding retired this PR's own justification for stale numbers. The "never edit an applied migration" rule protects *applied* migrations — these are unmerged and cannot have been applied outside a scratch database, so the byte-identity constraint (a merge-time verification, recorded in `f9dc155`) no longer bought anything and was costing operator clarity at the worst moment.

## Deliberately not included

The final contract PR — dropping the compat views, alias routes, legacy `chat:*` events and the demoted 120s polls — is **not** in this branch. Those shims exist to survive the deploy window and must soak at least one release first.

## Post-review remediation (2nd round)

Four more reviews ran against the branch — an audit sweep, a maintainability pass, a purity audit of the churn hotspots, and a full code review. Every finding that is a correctness bug, a wiring gap, or dead/mis-owned code is fixed in `27eec80`, `64d5ebb` and `7e8097c`; the large refactors are answered in the thread rather than attempted, because the right response to a 563-file diff is not to make it bigger.

**A CI gate that had never actually run.** `ci / Unit Tests` was failing on `ADMIN_DATABASE_URL is not set`. The variable is set on the workflow step — but it was missing from turbo's `globalEnv`, and turbo strips anything not listed there, so every `turbo run test:coverage` task saw it unset and the suites gating on it (the GDPR eraser's pseudonymization path, the audit hash-chainer, SIEM delivery) reported SKIPPED while the job went green. This branch's own `require-db.ts` is what surfaced it, by turning that silent skip into a loud failure.

**Silent delivery loss.** `useConversationSubscription` emitted `leave_conversation` unconditionally on unmount. Room membership is per-socket and a tab has one socket, so closing one view of a conversation dropped the room for every other view — and two views of one conversation is the documented expected shape. Now refcounted, with joins deliberately left unconditional: a duplicate join is idempotent, and gating it on 0→1 would strand a replaced socket in no room at all.

**Silent data loss.** `mintId`'s counter fallback restarted at 0 on every page load, and is taken whenever `crypto.randomUUID` is absent — any non-secure context, i.e. an onprem deployment over plain HTTP. A re-minted `opId` was read as a replay, so a split or resize returned the recorded rev and applied nothing: no error, no retry, no toast.

**An access asymmetry.** `GET .../plan` gated on ownership while its four siblings route through `canAccessConversation`, so on a shared conversation the plan chip 404'd for exactly the collaborators it was shared with. Now the shared predicate; `DELETE` stays owner-only, deliberately and with the distinction documented.

Plus: an existence oracle in the chat entry's strategy selection, a fabricated `createdAt` on truncated message refs, an ambient `process.env.SHELL` read, a rewrite rule that aliased nothing (with the test that pinned the typo), an unused option that was also a memo hazard, a dead export, two stale comments, and — per `CLAUDE.md` — the **CHANGELOG entries this epic owed**.

Two deferrals that had lost their owner are now on the epic's open-task list rather than in a migration header and a `Set` inside a unit test: 0250's four `NOT VALID` constraints, and the four pane-grid verbs' missing renderers.

**Two places where the docs claimed more than the code delivers**, both raised independently by two reviews, both now corrected *and enforced*:

- "One pipeline behind two routes" names the ENTRY. The two strategy functions are ~2,080 and ~1,460 lines and share **165 byte-identical substantive lines**, clustered in the epilogue — where the billing settle, `releaseHold` and the exactly-once mention latch live. Measured, not estimated. The spec said the one duplicated stretch had been shared outright, which read as a claim the rest was merely similar; it is not. `turn-duplication-ratchet.test.ts` now makes new copy-paste between the two files fail, and makes the figure in both prose homes track the measurement — so it can neither grow silently nor be described wrongly.
- §4 listed only *renamed* and *frozen*, so anything not called out as frozen read as done. The missing third category is *unfinished*, and it was the largest: ~30 exported `AgentSession*` names, the `{ session: … }` response key, four `Session*` component files. Documented with counts, and the two sites reviewers pointed at (`AgentPanesProps.sessionId`, `workspace-conversations.ts`) now say at the point of use that they hold `agent_workspaces.id` and not the deprecated `conversations.sessionId` — the standard `session-tools.ts` sets for its own frozen wire vocabulary. The sweep itself is deliberately deferred: at ~570 files it buys naming and spends review attention.

Verification: `typecheck` 17/17 · `lint` 15/15 · apps/web **16705 passed** · realtime 960 · infrastructure 341 · evidence manifest 15/15 proven, 0 gaps. Every behavioural fix carries a test observed to fail against the prior code.

## Known, flagged rather than silently resolved

- ~~The 8 renumbered migration files still cite their pre-renumber numbers.~~ **Resolved** — see post-review remediation above. Every stratum is now consistent: the runbook, the spec, the DSR runbook, 28 source files and the migration SQL itself. UPGRADE.md carries a drift guard.
- ~~`set_plan` / `clear_plan` (from #2359) write `conversations` directly without bumping `rev` or emitting.~~ **Resolved** — both now go through `conversationRepository.setConversationPlan`, which bumps `rev` and emits in the same statement, guarded by `isDistinctFrom` so re-binding a plan a conversation already has writes nothing. The rule and master's new writer meet.
- **The `CodeQL` check is red on one new alert**, `js/user-controlled-bypass` (high) at `page-agents/consult/route.ts:304`. It is a false positive of the same class as the **13 instances of this exact rule already open on master** (#228–234, #240, #275): the flagged condition is `if (!requested || !access?.allowed)`, where the "user-provided value" is the `conversationId` selecting which row to authorize and the guard is the authorization verdict on that row. Every resource-addressed authz check has this shape. The alert exists *because* this branch added a check where there was none — that route previously returned another member's private transcript to anyone who could name the conversation. **Deliberately not dismissed**: dismissing a high-severity security alert is a maintainer's call.
- CodeQL alert #275 predates this branch (the `/help` short-circuit from #2329) and remains untriaged on master.

https://claude.ai/code/session_01XnzUuAixdTJ8xDpP5S92Ts


2witstudios:
## Review-coverage note: both automated reviewers abstained

Recording this explicitly, because the two bot comments above are easy to scroll past and together they mean **this PR received no automated review at all**:

- **CodeRabbit** — skipped: 529 files against a 100-file limit.
- **Codex** — skipped: account review-usage limits reached.

There are also zero human review threads. So none of the usual automated safety nets ran on this diff. What follows is the compensating verification I ran directly against the branch, concentrated on the **master-integration surface** — the part no sub-PR review ever covered, since all 26 merged PRs were reviewed before the `origin/master` merge (`f9dc155`) rebuilt the migration chain and ported `/plan` into the collapsed pipeline.

### Found and fixed

**`infrastructure/UPGRADE.md` cited the pre-renumber migration numbers** (21 references, all off by one). This is not cosmetic drift. A renumber *shifts* numbers rather than orphaning them, so each stale citation resolved to a different **real** migration:

- The rename section described **`0253`** as `ALTER ... RENAME`, "Nothing is dropped, recreated or rewritten", "no row is touched". In this release `0253` is `0253_drop_chat_messages` — the epic's **only irreversible** migration, and the one the PR body tells operators to verify a snapshot before. The runbook and the deploy instructions contradicted each other on the single step that cannot be undone.
- `never apply 0251 in the same run as 0246` named master's `0246_sticky_venus` and `0251_messages_unification_validate_fk` — neither is the pane-grid pair it means.
- The chat_messages soak-gate rule pointed at 0248/0250/0252 instead of 0249/0251/0253.

Fixed in 6fa37c3, each citation re-verified against the migration it names, plus a drift guard asserting every `NNNN_name` the doc cites resolves to a real file — **observed to fail** on the original text and pass after the fix.

The PR's existing "known, flagged" note about stale numbers is correctly scoped to the SQL bodies (byte-identity is forced there — the runner hashes the file). That constraint never applied to a markdown runbook.

### Verified, no change needed

Each of these was a claim I tried to falsify rather than confirm:

| Claim | Result |
|---|---|
| 8 renumbered migrations are byte-identical | sha256-verified all 8 against `e3ceadf6e` — identical |
| Journal ordering survived the renumber | 246 to 254 strictly monotonic; the 8 non-monotonic entries are pre-existing on master |
| No schema drift | `drizzle-kit check` clean |
| Evidence manifest holds | 15 guarantees, 15 proven, 0 gaps |
| Page titles cannot forge prompt structure | `clipDescription` collapses whitespace **and** the C0/C1 control range, so newline injection is genuinely blocked |
| Page-chat uses the principal-aware check | confirmed `canPrincipalViewPage`; global surface is `allow: ['session']` only, so its user-level default is correct |
| Alias rewrite does not bypass auth | middleware matcher is a broad negative lookahead covering both spellings; rewrite is `afterFiles` |
| Compat views are auto-updatable | single-table column aliases — genuinely updatable in Postgres |
| Tool contract is frozen | pinned as explicit literals, not `.snap`, so `--update` cannot silently re-pin |
| Phase 5's "assert reclaim trigger present" | present, live-fires on `agent_workspaces`, and refuses to silently skip |

Two things I probed as suspected defects and concluded are **correct as shipped**:

- **`set_plan`/`clear_plan` not bumping `rev`** (flagged in the PR body). Bumping without emitting would be *worse*: clients hold a rev watermark and refetch on a gap (`conversation-apply.ts`), so a silent bump would make the next real message event look like a gap and force a spurious full refetch on every plan bind. The full fix is bump **+** emit, which belongs with the contract PR. Shipped behaviour is coherent.
- **`planPageId` absent from the Art 15 export.** `agentPageId` is projected only to classify a message's `pageId` (`type='client'`), not as a disclosure decision — so there is no asymmetry to correct.

### Local validation

`typecheck` 17/17 clean · `test:unit` 8756 passed / **0 test failures** (12 file-level failures are all local test-DB connection errors, the known env-only baseline — CI runs these against a real DB and they pass) · `infrastructure` 341/341 · evidence manifest 15/15 · `drizzle-kit check` clean · branch at `origin/master` tip, 0 conflicts.

One caveat worth stating plainly: this covered the integration surface and the PR's own claims, not all 214k added lines. The three adversarial reviews the description cites remain the primary evidence for the other 26 PRs' internals.

2witstudios (commented):
## Audit: shortcuts, leftovers, unproven tests, wiring gaps

Swept the branch (195 commits, 573 source files) for deferred work that lost its follow-up, dead/half-finished refactors, tests that read as proof without being proof, and authz/wiring asymmetries.

**Green before anything else:** `typecheck`, `lint`, `knip`, and `scripts/check-evidence-manifest.ts` (15/15 proven, 0 gaps) all pass. Every name in `CONVERSATION_CONTENT_EVENTS` / `CONVERSATION_DIRECTORY_EVENTS` has both an emitter and a listener — no orphan events. The rolling-deploy shims are enumerated in the spec's PR 17 checklist *and* mechanically guarded (`agent-workspaces-rename-migration.test.ts` asserts the views announce their own expiry). The tenant-export column drift guard and the CI env-var fixes for the dark suites are real. That's a high bar, so the list below is short.

---

### 1. MAJOR — 0250's four constraints are permanently unvalidated, and the PR that was supposed to validate them silently didn't

`0250_cooing_klaw.sql` adds four constraints `NOT VALID`:

- `ai_stream_sessions_conversation_id_conversations_id_fk`
- `conversations_global_context_null_chk`
- `conversations_page_context_present_chk`
- `conversations_drive_context_present_chk`

Its header names the follow-up explicitly: *"`VALIDATE CONSTRAINT` … is left to the PR that has looked at real data (Phase 4 PR 14, alongside 0249's chat_messages FK)."* And its runtime NOTICE says the violating rows *"must be repaired before VALIDATE CONSTRAINT."*

PR 14 is `0251_messages_unification_validate_fk.sql`. It validated **only** 0249's `chat_messages` FK — on a table that 0253 then dropped. `grep -rn "VALIDATE CONSTRAINT" packages/db/drizzle/` returns exactly one executable statement, `0251:155`. Nothing validates the four above, and they are **not** on the PR 17 checklist in `docs/2.0-architecture/agent-sessions.md` §4, which the spec calls "the epic's one remaining open task." The deferral has no owner left.

This is not theoretical — the consequence is already written down in `packages/lib/src/compliance/erasure/purge-stream-state.ts:26`:

> `0250's FK is NOT VALID, so rows whose conversation was hard-deleted before it landed still dangle and are unreachable by any cascade.`

The Art. 17 path compensates with a user-scoped delete, so erasure is covered. What isn't covered is the **conversation-scoped** case the retention engine relies on (`retention-engine.ts:140-145` documents purging one soft-deleted conversation as taking its stream checkpoints via the 0250 cascade) — for a pre-0250 dangling row there is no cascade to ride, and those rows carry `parts`, i.e. message content, forever.

**Ask:** either add a `0255` that runs the four `VALIDATE CONSTRAINT`s (0250's own pre-audit NOTICE already tells the operator the counts, and VALIDATE takes only `SHARE UPDATE EXCLUSIVE`), or add all four to the PR 17 checklist so they leave this epic with an owner instead of a dangling reference to a PR that came and went.

---

### 2. MODERATE — four tools added by this PR ship with no renderer, and the gap lives only in a test file

`registry-coverage.test.ts`'s `PENDING_RICH_RENDERERS` now carries 16 entries; `list_panes`, `resize_pane`, `move_pane`, `arrange_panes` are **new in this branch** (2f3e39b9b). The ledger itself is well built — it fails both directions and can't grow silently — and the stated reason (a grid card is a layout-diagram design task) is fair.

The problem is where the gap is recorded. The pane-grid verbs are Phase 3's headline capability, and every one of them renders to the user as a raw payload. That's tracked in a `Set` inside a unit test, not in the spec's open-task list and not in an issue. The ledger stops it from being *forgotten by CI*; it does nothing to stop it being forgotten by the epic. Please add the pane-grid renderer set to the PR 17 checklist or open an issue and cite it in the ledger comment.

---

### 3. MODERATE — the plan endpoint is owner-only while every other conversation-scoped surface is "owner OR (isShared AND page access)"

`api/ai/conversations/[conversationId]/plan/route.ts:55` and `:90` both gate on `row.ownerId !== userId`. Everything else conversation-scoped in this epic routes through `canAccessConversation` — the realtime `join_conversation` handler, `/stream-join`, `/active-streams`, and the new `/api/ai/conversations/revs` — and `conversation-access.ts` is emphatic that the point of that module is that the enforcement points *cannot drift*.

Effect on a shared page conversation: a collaborator can read the messages and join the room, but `GET .../plan` 404s, `PlanChip`'s fetcher throws, `data` stays undefined, and the chip renders nothing (`PlanChip.tsx:84`, `:101`, `:118`). So the plan the agent is visibly following is invisible to exactly the people the conversation was shared with — and every one of their views fires a failing request that SWR then retries.

The route header explains why it distinguishes 404 from `{plan: null}`, which is a different question. If owner-only is deliberate here, it needs the same kind of stated reason `conversation-access.ts` gives for excluding shared *global* conversations. If it isn't, this should be `canAccessConversation` like its four siblings.

---

### 4. MINOR — the Phase 5 rename stops at the module boundary in `apps/web`

ac640f324 says "the workspace store says workspaceId on the inside too," but the store's callers still name the workspace address `sessionId`:

- `AgentPanes.tsx:87, 154, 163, 183, 194, 204, 328, 363`
- `useSessionRecord.ts:49`
- `panes/workspace-conversations.ts:120, 174, 260`
- `AgentPageView.tsx:280, 308`, `useSpawnSession.tsx:95`

To be clear, these are **not** the compat field — they read `session.workspaceId` correctly. It's the old word on a local holding the new thing, which produces lines like `AgentPanes.tsx:363`'s `session.workspaceId === sessionId`. Given §5's own standard ("a description the code no longer enforces is worse than no description"), the same hazard applies one layer down: the next reader has to prove to themselves that `sessionId` here isn't the deprecated key. Cheap sweep, and it's the one this epic keeps claiming to have finished.

---

### 5. MINOR — dead export

`packages/lib/src/realtime/conversation-event-names.ts:52` — `ConversationEventName` is exported and unused (knip). The two narrower unions above it are both consumed; this one isn't. Either drop it or note why it's kept for external consumers.

---

### 6. NIT — `chat-pipeline` / `handleChatTurn` oversell what got consolidated

What's actually shared is five preamble steps plus `start-chat-generation.ts`. `page-chat-turn.ts` (2290 lines) and `global-chat-turn.ts` (1659) stay separate, with ~89 lines of *exact* duplication — and it's concentrated in the two places you'd least want a copy: the `streamText` config block and the placeholder-terminalization cleanup (`page 1758-1776 / 1837-1845 / 1894-1915` ≡ `global 1312-1327 / 1373-1381 / 1412-1429`). That these two already drifted once is on the record — 5c1bc5410, "bring the global assistant to parity with the page chat."

The module header is genuinely honest about this (the divergence table is excellent, and the argument against ~20 conditionals in the app's highest-risk path is right). The residual risk is purely that the *directory name* says "one pipeline" and the next person doesn't read 60 lines of header before assuming a fix lands on both surfaces. Extracting just the terminalize-placeholder cleanup would remove the duplicated block that actually bites.

2witstudios (commented):
## Maintainability / separation-of-concerns review

Not a correctness review — no bugs claimed except where marked. This is about where the
implementation stops short of its own thesis, and where the seams are in the wrong places.

The branch's thesis is *one source of truth*: one `messages` table, one `conversations` table,
one relational pane layout, one chat entry point. The data layer delivers that. **The code layer
above it still forks page/global at four levels**, and the `agent_sessions → agent_workspaces`
rename stopped at the file/column boundary. Everything below is a variation on those two themes.

---

### 1. The turn strategies are two functions, not two modules

`runPageChatTurn` is ~2,080 lines (`page-chat-turn.ts:209`–2290) in **one function**, with a
single `try` spanning `:341`–`:2190` and ~20 hoisted `let`s carrying state across phases
(`userId`, `holdId`, `agentRun`, `resolvedProvider`, `assistantMessagePersisted`,
`mentionNotified`, `lifecycle`, `activeStreamId`, …). `runGlobalChatTurn` is the same shape at
~1,460 lines.

The divergence table in `handle-chat-turn.ts:40` is a genuinely good piece of engineering — it
argues persuasively that the two strategies should not collapse into one function with 20
conditionals, and I agree. But that argument doesn't imply each strategy must be one function.
The turn has clean internal phases that are the same on both surfaces even when their
*implementations* differ:

```
resolve context → authorize → mint/resolve conversation → assemble tools
→ assemble prompt → gate credits → persist user message → stream → settle
```

Extracting those as functions that take and return a typed `TurnContext` would replace the
hoisted `let`s with values, and make the mega-`try` a sequence of narrowly-scoped ones. The
highest-value single extraction is the **epilogue** — stream construction, `onFinish`, terminal
persist, hold settle, telemetry (`page-chat-turn.ts:1754`–2190 ↔ `global-chat-turn.ts:1290`–1600).
Normalized diff finds 4 matching runs ≥9 lines in that region alone; it is also where billing
settle, `releaseHold`, and the exactly-once mention latch live. Two copies of the money path is
the duplication I'd retire first.

Measured overall: 428 identical normalized code lines between the two files (of 1,052 in the
global one), mostly in short bursts — so the divergence table is honest, and I'm *not* asking for
a merge. Just for the shared spine to be named.

### 2. Two repositories over one `conversations` table

`conversation-repository.ts` (901) and `global-conversation-repository.ts` (492) both write the
same table, split by `type`, with six overlapping methods:

| conversationRepository | globalConversationRepository |
|---|---|
| `listConversations` | `listConversations` |
| `createConversation` | `createConversation` |
| `getConversation` | `getConversationById` |
| `upsertConversationTitle` | `updateConversationTitle` |
| `autoTitleConversation` | `autoTitleConversation` |
| `softDeleteConversation(ById)` | `softDeleteConversation` |

This is exactly the structure Phase 4 eliminated for messages. A reader asking "where does a
conversation get soft-deleted?" gets two answers with different signatures and different
authorization postures (`userId`-scoped on one side, `agentId`-scoped on the other).

`conversationRepository` is also wide beyond CRUD: text processing (`extractPreviewText:79`,
`generateTitle:112`), aggregation (`listConversations`, `countConversations`), sharing
(`setConversationShared`), and an audit write (`logConversationDeletion:912`).

### 3. `messageRepository` kept the page/global fork the table dropped

Six method pairs (`message-repository.ts:583`+): `savePage`/`saveGlobal`,
`insertPageStreamingPlaceholder`/`insertGlobal…`, `materializePage…`/`materializeGlobal…`,
`editPage`/`editGlobal`, `softDeletePage`/`softDeleteGlobal`, plus the private
`savePageMessageRow`/`saveGlobalMessageRow`.

`editPageMessage:891` and `editGlobalMessage:969` are ~95% identical — the transaction bodies are
line-for-line the same. `softDeletePageMessage:936` and `softDeleteGlobalMessage:1030` differ in
exactly two expressions: the scope descriptor (`{kind:'page', pageId}` vs `{kind:'global',
ownerId}`) and the legacy channel id (`args.pageId` vs `globalChannelId(ownerUserId)`).

That descriptor is already a union in the codebase, and `contextForMutation:426` already takes
it. `editMessage(args, scope: MessageScope)` with a `legacyChannelFor(scope)` helper collapses
all six pairs and removes the possibility of the two legs drifting (which they have already
done once — the global edit doesn't refetch/convert parts, the page edit does; the comment at
`:1003` records this as deliberate parity-preservation, which is fine, but it should be a
documented branch inside one method rather than an emergent difference between two).

### 4. Three message edit/delete routes never got the `handle-chat-turn` treatment

- `api/ai/chat/messages/[messageId]/route.ts` (322)
- `api/ai/page-agents/[agentId]/conversations/[conversationId]/messages/[messageId]/route.ts` (284)
- `api/ai/global/[id]/messages/[messageId]/route.ts` (266)

872 lines running the same shape — authenticate → load message → authorize → delegate to the
repository. The routes are thin and correctly delegate, so this is low-risk duplication, but the
consolidation pattern this very branch introduced for POST (one entry, strategy chosen from the
conversation) applies verbatim and isn't applied.

### 5. `workspace-layout-verbs.ts` is five modules in one file

1,173 lines holding: fraction math and normalization (`:118`–`:410`), the pane-grid algebra
(`newWorkspace`–`reorderColumns`, `:425`–`:762`), the verb zod schema + `applyVerbLocal`
dispatcher (`:775`–`:1069`), the grid↔persisted wire mapping (`:1071`–`:1173`), and the open-
placement policy (`isReplaceable`/`resolveOpenPlacement`, `:860`–`:936`).

The fraction layer alone (~300 lines of `quantizeFraction`/`settleToOne`/`scaleToTotal`/
`rebalanceFractions`/`resizeShare`) is pure numeric code with no knowledge of panes and is the
part most likely to be reasoned about in isolation. `layout-fractions.ts` / `pane-grid.ts` /
`layout-verb-schema.ts` / `layout-grid-wire.ts` / `open-placement.ts` would each be independently
testable, and `workspace-layout-wire.ts` (78 lines) shows the split is already started.

Same note, smaller: `contract.ts` (445) carries four unrelated contracts — workspace DTO, pane
schemas, shell/PTY (`shellConnectPayloadSchema`, `SHELL_BRIDGE_ROUTES`, scrollback limits), and
assorted caps.

### 6. `agent-workspaces-runtime.ts` has outgrown "DI wiring"

1,000 lines, 25 exports. Its header claims "ZERO decision logic lives here, by mandate" and that
discipline mostly holds — but the module now fronts six subsystems: lifecycle
(`spawnSession`/`endSession`), access (`checkSessionAccess`/`checkAccessForSubject`/
`checkSessionEndAccess`), advisory locking (`withSessionListingLock`, `SessionListingLockBusyError`),
conversation CRUD (`claim`/`create`/`close`/`reopen`/`count`), sandbox provisioning
(`provisionSessionSandbox`, `ensureGlobalSandboxSession`, `ensureDriveSessionForConversation`),
and storage measurement (`measureWarmSessionStorage`). Splitting by subsystem — keeping the lazy
memoized store/host singletons in one shared `*-deps.ts` — preserves the DI discipline while
making the import graph say what depends on what.

Related: five directories now hold agent-workspace code with boundaries that aren't self-evident
from their names — `packages/lib/src/agent-workspaces/` (pure), `packages/lib/src/services/
agent-workspaces/` (`*-store.ts`), `apps/web/src/lib/agent-workspaces/` (`*-runtime.ts`),
`apps/web/src/stores/agent-workspace/` (client), `apps/web/src/app/api/agent-workspaces/`. The
`-store` / `-runtime` suffixes carry real meaning here; a one-paragraph README at
`packages/lib/src/agent-workspaces/` stating the four-layer rule would keep the next contributor
from putting a decider in a runtime.

### 7. Client store: per-workspace state lives in five containers

`useAgentWorkspaceStore.ts` keeps `workspaces`/`sync`/`focus`/`queueErrors` in zustand, but the
transport half of a workspace's lifecycle lives in four module-level containers beside it:
`queueGeneration:305`, `retryTimers:331`, `labelRefreshTimers:343`, `registeredEditing:374`.
`drop():450` has to remember all of them by hand — and the exported
`__resetWorkspaceQueuesForTests:321` (test-only API shipped in production code) exists precisely
because that's easy to get wrong.

**Minor leak:** `drop()` cleans the timers and the editing registration, but `queueGeneration` is
only ever `bumpGeneration`'d — the only `delete` in the file is `.clear()` inside the test reset.
An entry is inserted per workspace id and never removed, so the Map grows for the lifetime of the
tab. Small (one int per workspace ever opened), but unbounded, and the "module-level so it
survives the sync entry" comment at `:298` explains why it can't simply be deleted in `drop()` —
which is itself the argument for a per-workspace controller object (generation + timers +
editing registration) created and disposed with the sync entry, giving one place to clean up and
one honest lifetime.

### 8. The rename is ~80% done, and the remaining 20% is the part users and callers see

The table, columns, directories, URLs, and store internals are renamed. The **exported vocabulary
is not**: ~190 residual `AgentSession*` identifiers across ~30 exported names in non-test source —
`AgentSessionRecord` (25), `AgentSessionRowStamps` (17), `AgentSessionStore` (12),
`AgentSessionListFilter` (11), `AgentSessionDTO` (8), `spawnAgentSession` (7), `endAgentSession`
(7), `listAgentSessions` (5), `agentSessionDtoSchema` (5), …

Concrete edges:

- **The wire still says `session`.** `GET /api/agent-workspaces/[workspaceId]` returns
  `{ session: AgentSessionDTO }` (`[workspaceId]/route.ts:115,178`; `route.ts:490,554,590`), and
  `useSessionRecord(sessionId)` fetches `/api/agent-workspaces/${sessionId}` — a workspace id in a
  variable named `sessionId`, hitting a workspace URL, reading a `session` key.
- **An alias shim renames the new name back to the old one:**
  `agent-workspaces-runtime.ts:108` — `export { MAX_ACTIVE_WORKSPACES_PER_OWNER as
  MAX_ACTIVE_SESSIONS_PER_OWNER }`.
- **The components layer is untouched:** `SessionPanes.tsx`, `SessionChat.tsx`,
  `AssistantSessionChat.tsx`, `useAgentSessionChat.ts`, `useSessionRecord.ts`,
  `useSpawnSession.tsx`, `EndSessionDialog.tsx`, `session-groups.ts`.
- **Stale header:** `agent-workspaces-runtime.ts:3` still says "services/agent-sessions".
- **Directory spelling:** `apps/web/src/stores/agent-workspace/` (singular) against
  `agent-workspaces` (plural) everywhere else.
- **Spec doc:** `docs/2.0-architecture/agent-sessions.md`, cited by ~6 source file headers.

The cost isn't cosmetic. "Session" now means four unrelated things in one tree: an auth session
(`packages/db/src/schema/sessions.ts`), a browser session (`browser-session-id.ts`), a terminal
PTY session (`terminal-session-map.ts`), and an agent workspace. `session-tools.ts` handles this
exactly right — it freezes `sessionId` on the *wire* because the tool contract is frozen for
running agents, and says so at `:54` and `:285`, mapping to `conversationId` at the zod boundary.
Everything else should either finish the rename or adopt that same explicit freeze-and-document
posture. Right now it's neither.

**Dead config:** `next.config.ts:195` — `{ source: "/api/agent-workspaces", destination:
"/api/agent-workspaces" }` rewrites a path to itself. Almost certainly a sed casualty from the
rename (intended source: `/api/agent-sessions`). Harmless — the `:path*` rule below it already
covers the bare path — but it should go with the rest of the block, which the comment says "the
contract PR deletes".

### 9. Two live realtime protocols carrying the same facts

Every message mutation emits twice: the authoritative `conversation:*` event and the legacy
`chat:*` broadcast. The new names are centralized in one module
(`packages/lib/src/realtime/conversation-event-names.ts`, each name appearing exactly once); the
legacy names are bare string literals appearing 5–8 times each across routes, hooks,
`socket-utils.ts`, and the repository. Being mid-migration is fine; what's missing is any
in-code record of what retires the legacy leg and when, so the asymmetry reads as permanent.

### 10. One rule written twice in two languages

`unifiedPageScope()` and `unifiedPageScopeSql()` (`unified-message-scope.ts:74,87`) encode the
same predicate in the drizzle builder and in raw SQL. The module doc is explicit that scattering
this rule is the failure mode it exists to prevent — and then keeps two copies, with 5 call sites
of the raw form in `conversation-repository.ts`. If the hand-written aggregates genuinely can't
use the builder, a test asserting the two forms select identical row sets would close the drift
hazard the module was written to close.

### 11. Known gaps are recorded in prose, not in anything greppable

Zero `TODO`/`FIXME`/`@deprecated` markers across all 257 changed source files — which reads as
clean until you notice the gaps are documented in paragraphs instead:

- `page-chat-turn.ts:279` — "The durable fix (idempotent `createMentionNotification` per
  user+message) is a filed epic D task."
- `next.config.ts:193` — "The contract PR deletes this block."
- `useAgentWorkspaceStore.ts` — the abandoned-queue path deliberately shows the server snapshot
  over a local edit.

The prose is excellent and I'd keep every word of it. But "what is still owed on this code" is
currently unanswerable by grep, and the PageSpace epic board isn't visible from an editor. A
single convention — `// FOLLOW-UP(#NNNN):` on the line above the paragraph — would make the
backlog enumerable without losing the explanation.

---

### What I'd sequence

1. Extract the shared turn epilogue (§1) — highest risk-per-line in the branch, currently
   maintained twice.
2. Collapse the six `messageRepository` page/global pairs onto the scope union (§3) — mechanical,
   well covered by existing tests, removes the largest remaining fork.
3. Finish or formally freeze the rename vocabulary (§8) — cheap now, and it gets more expensive
   with every new caller of `AgentSessionRecord`.
4. Split `workspace-layout-verbs.ts` (§5) and give the client queue one owner (§7).
5. Merge the two conversation repositories (§2) and the three message-mutation routes (§4).

2witstudios (commented):
## Purity audit of the churn hotspots

Ranked the diff by lines-changed × commit-touches and audited the top ~30 source files against the purity/functional-core claims the epic makes (`docs/2.0-architecture/agent-sessions.md` §1, and the ~180 in-source `pure` claims).

**The claims that are made hold up. The problem is where they stop.**

### What actually verified clean

- `packages/lib/src/agent-workspaces/workspace-layout-verbs.ts` (1173 lines) — no clock, no random, no I/O, no env. Every transition allocates; the only in-place writes (`settleToOne`'s `settled`, `splitRight`'s `columns.splice`) are on locally-allocated copies. `rebalanceFractions` / `scaleToTotal` / `resizeShare` are total and referentially transparent, and the fraction invariant really is established in one place.
- `stores/agent-workspace/verb-queue.ts`, `stores/agent-workspace/pane-labels.ts`, `applyPaneLabels` (`workspace-layout-runtime.ts:206`), `paneScopesOfVerb` (`authorize-pane-scope.ts:79`), `buildActivePlanPrompt` (`plan-binding.ts:123`), `planConnect` / `planColdTailPersist` / `latestActivityAt` (`shell-handler.ts`), `extractPreviewText` / `generateTitle` (`conversation-repository.ts:77,110`), `processMessageContentUpdate` (`message-repository.ts:1451`) — all pure exactly as documented.
- No module-level mutable state in the top-12 churn files other than the two places that document themselves as the imperative shell (`useAgentWorkspaceStore.ts:284-374`, `conversation-rev-sync.ts:34,101`).
- Every `.reverse()` / `.sort()` in the diff operates on a freshly-allocated query result, not on a caller-owned array. No argument mutation found.

### 1. The two biggest files in the diff are the opposite of the pattern — `page-chat-turn.ts:209`, `global-chat-turn.ts:196`

`runPageChatTurn` is **one 2082-line function** with ~194 branch points and 54 inline `await`s. `runGlobalChatTurn` is **1464 lines**, ~110 branches, 41 awaits. Decision and effect are fully interleaved — no extractable decision core, nothing unit-testable without the DB and the model provider.

This is the same monolith that was `route.ts` `POST` on master (lines 149–2270, ~2120 lines). Phase 5 relocated it and thinned the route to 30 lines, which is a real improvement to the route — but the function itself came across intact. Four sibling modules in this same PR carry the header *"Pure decision logic over injected deps, per the repo rule that branching logic belongs in a pure decision"* (`claim-` / `close-` / `create-` / `reopen-conversation-in-workspace.ts`). That rule was applied to ~40-line helpers and not to the 3500 lines where it would pay.

Not asking for a rewrite in this PR. Asking that the epic doc stop implying the discipline is uniform, and that the credit-gate / model-resolution / plan-binding / tool-assembly decisions inside these two get named as follow-up extraction targets.

### 2. …and the two of them share 166 identical substantive lines

166 non-comment lines ≥40 chars are byte-identical between `page-chat-turn.ts` and `global-chat-turn.ts` (out of 535 / 439 distinct substantive lines respectively — ~38% of the global file). Commit `5c1bc5410` *"bring the global assistant to parity with the page chat"* is the cost of that already being paid once inside this PR. This is the exact "two writers kept in step by convention" failure that `workspace-layout-verbs.ts:8-11` congratulates itself on having eliminated for the grid — it's alive and well one directory over.

### 3. Real bug: the pane/op id counter fallback can silently discard a user's layout change — `useAgentWorkspaceStore.ts:284-291`

```ts
let paneCounter = 0;
function mintId(prefix: string): string {
  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') { ... }
  paneCounter += 1;
  return `${prefix}-${paneCounter}`;
}
```

The counter is module-scoped, so it **restarts at 0 on every page load**. `crypto.randomUUID` is unavailable in non-secure contexts — which for this product means any onprem/self-hosted deployment reached over plain HTTP on a LAN address, not a hypothetical.

Both id families it mints are compound-PK'd per workspace:

- `agent_workspace_pane_columns` PK is `(workspaceId, id)` (`agent-workspace-layout.ts:137`) — so a reload re-minting `pane-1` / `col-1` into the same workspace produces the duplicate-id write `verb-queue.ts:59-69` documents as "a 500 five retries later when the queue gives up and discards the user's work".
- Worse, `agent_workspace_layout_ops` PK is `(workspaceId, opId)` (`agent-workspace-layout.ts:137`), and `enqueueVerb` mints `opId` from the same counter (`useAgentWorkspaceStore.ts:479`). A re-minted `op-1` hits `store.findOp` (`workspace-layout-runtime.ts:103`), is read as a **replay**, and returns the recorded rev without applying the verb. Silent no-op: no 500, no retry, no toast, the user's split/resize just doesn't happen.

Suggest seeding the fallback with something per-load-unique (page-load timestamp + counter, or a `Math.random()` prefix) so the degraded path degrades to "ugly ids" rather than "collides with the previous session's ids".

### 4. Minor — ambient input in an otherwise-pure resolver, `shell-handler.ts:293`

`resolveShellCommand` reads `process.env.SHELL` (`:302`) while sitting between three siblings explicitly labeled `Pure:` (`planConnect`, `planColdTailPersist`, `latestActivityAt`). Same file, same testing story; the shell binary should be a parameter like everything else that function needs.

### 5. Minor — clock read in the event payload builder, `conversation-events.ts:250-256`

`shapeMessageForEvent` falls back to `new Date().toISOString()` when a message carries no `createdAt`, directly beneath the `PURE audience computation` block. That stamps a *wrong* ordering key onto the truncated `messageRef` rather than omitting it or failing loudly. Given `createdAt` is what the client orders on, "now" is the one value guaranteed to be wrong for a message being backfilled or replayed.

---

Findings 3–5 are actionable in this PR. 1–2 are scope calls for you — my ask there is only that the docs match the code.

2witstudios (commented):
## 🔬 Code review — agent-session single source of truth

Reviewed `master...pu/broken-sessions` (247 non-test source files, 9 migrations `0246`–`0254`, ~590 files total). Ran churn against the diff, read the epic board (`qrtixghswpq6f1r4gbp7aac0` — phases 0–4 and 6 complete, phase 5 in progress), the spec (`docs/2.0-architecture/agent-sessions.md`), and every migration end to end.

**Verdict: 1 major, 2 minor, 3 nits.** No blockers. This is unusually disciplined work — the migrations reason about corpora this repo has never seen, the authorization split (`authorize-pane-scope.ts` write half / `resolvePaneLabels` read half) is exhaustive over the verb union by construction, and `conversation-apply.ts` reduces the whole delivery protocol to one pure, total function. `bun run lint` and `bun run build` are green; unit tests are 8817 passing with 13 DB-integration suites failing only because no local Postgres is up — and failing *loudly*, which is what `packages/db/src/test/require-db.ts` was added to guarantee.

---

### MAJOR 1 — the `conv:<id>` room join is not refcounted, so one unmount evicts the whole tab

`apps/web/src/hooks/useConversationSubscription.ts:150-162`

```ts
useEffect(() => {
  if (!socket || !conversationId) return;
  const join = () => socket.emit('join_conversation', conversationId);
  join();
  socket.on('connect', join);
  return () => {
    socket.off('connect', join);
    socket.emit('leave_conversation', conversationId);   // ← unconditional
  };
}, [socket, conversationId]);
```

The rev-watch registry three effects down is *deliberately* refcounted — `registerConversationWatch` carries a `count` field whose comment says "refcounted so two panes on the same conversation don't unregister each other" (`conversation-rev-sync.ts:24`, `:37-56`). The room membership underneath it is not, and it needs to be for the same reason: `socket.leave(room)` is per-socket, and a tab has one socket.

Two subscribers on one conversation is not hypothetical — it is documented as the expected shape:

- `GlobalChatContext.tsx:281` subscribes `currentConversationId` app-wide, with the comment *"Panes are different: each one subscribes to its own conversation… Duplicate delivery across the two is harmless — every cache action is idempotent by construction."* That is true of **join** and false of **leave**.
- `useAssistantSessionChat.ts:147` subscribes the same id per pane.
- `useAgentChannelMultiplayer.ts:62` does the same for a page-agent conversation.

**Failure scenario.** User has the Global Assistant open on the dashboard (provider watermarks conversation `C`) and also opens `C` in an Agents-console pane. They close the pane. Its cleanup emits `leave_conversation(C)`; the server drops the socket from `conv:C`; the provider is still mounted and still rendering `C`, but no longer receives `conversation:message_created`. A dispatch from another surface goes invisible until the next socket reconnect or an unrelated gap-heal — the epic's canonical bug, reintroduced through its own subscription primitive. Same shape for two console panes on one conversation, and for a pane whose `conversationId` changes from `A` to `B` while another surface still shows `A`.

The rev-gated protocol makes this *silent* rather than obviously broken: the surviving surface's watermark simply stops advancing, so nothing errors.

**Suggested fix:** fold the room join/leave into the existing refcounted registry (it already keys on `conversationId` and already knows the mount/unmount boundary), or add a sibling module-level `Map<conversationId, count>` that only emits `join_conversation` on 0→1 and `leave_conversation` on 1→0. The `connect` rejoin handler wants the same treatment so a reconnect re-joins each watched conversation once.

---

### MINOR 2 — `CHANGELOG.md` is untouched

`CLAUDE.md` states *"**Changelog**: Update any user-visible notes when changes land."* The last commit touching `CHANGELOG.md` is `1c4e08a9e` (Aug 4), before this branch's work, and nothing in the diff adds to `## [Unreleased]`.

This epic is dense with user-visible change that belongs there: the pane grid becoming server-authoritative and live across devices/members, panes and sidebars that stop showing blank while the DB holds messages, spawned workers appearing immediately instead of on a 20s poll tick, one unified chat history, `/plan` shipping as an editable starter skill, and `list_sessions` discovering shared workspaces while redacting other members' private thread titles.

---

### MINOR 3 — strategy selection on `/api/ai/chat` is an existence+type oracle over conversation ids

`apps/web/src/lib/ai/chat-pipeline/handle-chat-turn.ts:213-260`

With no `chatId`, the entry reads the caller-supplied `conversationId` with `conversationRepository.getConversation()` — no ownership predicate — purely to pick a strategy. The module doc argues this is fail-closed, and for *access* it is: `resolveOrCreateConversation` independently re-enforces owner + `type='global'` + `isActive`. But the two branches answer differently, and that difference is observable:

| body | response |
|---|---|
| `{messages, conversationId: X}` where `X` exists and is `type='global'` (anyone's) | routed to the global strategy → `404 "Conversation not found"` (`global-chat-turn.ts:468`) |
| `X` absent, or any other type | falls through → `400 "chatId is required"` (`page-chat-turn.ts:394`) |

So an authenticated account can distinguish "this id is an existing global conversation" from everything else. Impact is genuinely small — cuid2 ids aren't enumerable and nothing beyond existence leaks — but this codebase is meticulous about exactly this ("Refusal is UNIFORM across 'forbidden' and 'does not exist'… a caller probing ids learns nothing from the difference", `authorize-pane-scope.ts:30-32`; the same policy is restated on the verbs route). The lookup already returns `userId`; gating the widening on `conversation.userId === auth.userId` collapses both cases back to the unchanged 400 at no cost.

---

### NIT 4 — `ConversationSubscriptionContext.streamOptions` has no callers

`useConversationSubscription.ts:106-112`, consumed at `:279`. All three call sites (`GlobalChatContext.tsx:281`, `useAssistantSessionChat.ts:147`, `useAgentChannelMultiplayer.ts:62`) omit it. It is also the one dependency of the `cacheHandlers` memo that is an object, so if a future caller passes an unmemoized literal the handlers re-register every render — the churn class `16e3d2b90` was fixing. Either drop the option or note the memoization requirement on it.

### NIT 5 — stale comment naming a table that no longer exists

`packages/db/src/schema/chat.ts:43` — `// Soft-delete flag for rollback support (matches messages/chatMessages pattern)`. `chat_messages` is dropped by `0253`. `276a2b43c` and `7242d3f82` swept comments in `apps/web`; this one is in `packages/db`.

### NIT 6 — the "a conversation's page" rule has one home, and one place still spells it out again

`unified-message-scope.ts:51-53` establishes the derivation as `type='page' → contextId`, `type='client' → agentPageId`, explicitly because getting it subtly different at each call site was the failure mode. `workspace-layout-runtime.ts:442` then derives a pane's `agentPageId` as `row.type === 'page' ? row.contextId : null`, dropping the `client` branch. Harmless while panes never bind client threads, but it is a second spelling of the rule the epic just consolidated — worth routing through the shared helper or stating in a comment why a pane can't be a client thread.

---

### Reviewed and found clean

- **Migrations `0246`–`0254`.** The `DISTINCT ON` dedup in `0247`/`0252` (client-authored blob vs compound-PK rows), `0252`'s directional drop guard with the `pagespace.workspace_state_force_drop` escape hatch for the wedged tenant upgrade the migration itself blocks, `0249`'s ownership ladder and its honest departure from decision 4 (`conversations.userId` is `NOT NULL`, so tier 4 must skip, not quarantine), `0251`'s pre-validate audit that proves the `VALIDATE` scan is empty before running it, `0253`'s stale-twin repair for the dual-write window (with the argument for *why* those two shapes prove a lost write), and `0254` being rename-only so the `pgs-ses-` Sprite key fold and the GDPR reclaim trigger survive by construction. Each guard is re-runnable and each destructive step refuses rather than guesses.
- **Authorization.** `canAccessConversation` is genuinely single-sourced across the realtime `join_conversation` handler, `/stream-join`, `/active-streams`, the pane-scope write gate and the label read gate. `paneScopesOfVerb` and `mintedPaneId`/`mintedColumnId` are exhaustive with `never` fall-throughs, and the comments explain why a `default:` would be the dangerous shape in each. The label-free broadcast + per-viewer GET is the right call for a room wire that can't express per-viewer redaction.
- **Standards.** Zero `any`, zero `@ts-expect-error`, zero `TODO`/`FIXME` across 247 changed source files. Async `params` correctly awaited in every new route. No hand-rolled permission checks. Compat debt is enumerated as the PR 17 checklist in `agent-sessions.md §4` and the two 120s backstop polls it names are exactly the two that exist.

### Note on the working tree (not part of this PR)

The worktree has uncommitted work in `session-tools-runtime.ts`, `session-tools.ts`, `conversation-repository.ts` and its test — apparently in-flight fixes for a "MAJOR `list_sessions` ownership-is-not-access" finding and "MINOR 2" un-share eviction. It currently fails typecheck:

```
src/lib/repositories/__tests__/conversation-repository.test.ts(100,68):
  error TS2556: A spread argument must either have a tuple type or be passed to a rest parameter.
```

`vi.fn(() => Promise.resolve())` is typed zero-arg, so `(...args: unknown[]) => mockKickUserFromRooms(...args)` won't compile. `vi.fn((..._args: unknown[]) => Promise.resolve())` fixes it. Flagging so it doesn't reach CI.

PR #2484, opened 2026-08-24
fix(agents): a spawn refuses rather than run a worker with tools it was never granted

### LINKED ISSUE #2460: [agents] spawn_session with agent param: workers never receive sandbox tools, configured tool list silently ignored
## Context
- Date: 2026-08-22, drive: PageSpace - Marketing
- Agent: "Scraper Runner" (AI_CHAT agent), configured via `update_agent_config` with 24 enabledTools including sandbox tools (`bash`, `writeFile`, `readFile`, `editFile`, `git_clone/pull/add/commit/push/status`, `spawn_shell`, `send_shell`, `read_shell`, `kill_shell`, `list_sessions`) plus page tools.
- The config was accepted and echoed back on every submission — 24 tools confirmed, tried both `toolExposureMode: "search"` and `"upfront"`.

## What happened
Spawned workers via `spawn_session` with `agent: <agentId>` (three separate spawns across both exposure modes). In every case the worker's runtime tool surface was **page-family tools only** — no `bash`, no `readFile`, no `spawn_shell`. Not "permission denied" — simply absent. `tool_search` for sandbox keywords returned nothing.

Worse, the surface was **inconsistent between workers spawned under the same agent with the same config**:
- Worker 1 reported 2 non-core tools: `edit_sheet_cells`, `trash_page`
- Workers 2–3 reported ~10: the page family + `regex_search`/`glob_search`
- None ever matched the 24-tool config verbatim

Exact errors the workers hit:
- `read_file` → `Tool "read_file" is not permitted for this agent.` (`read_file` isn't even the name in the config — I configured `readFile`)
- `bash` → `Unknown tool "bash". Call tool_search("keyword") to discover available tools.`
- `spawn_shell` → `Unknown tool`

Control: Dev Coder (PageSpace - Dev) carries the same camelCase-style config (`bash`, `writeFile`, `readFile`, `git_*`, `gh_*`, `spawn_shell`...) and works, so name casing isn't the cause.

## What unblocked us
`spawn_session` WITHOUT the `agent` param — the worker runs under the conversation's own agent and gets the full sandbox + page tool surface. That worker completed the job.

## Impact
Agent delegation for sandbox work is impossible as-is. The silent divergence between stored config and runtime surface is the worst part: three spawns, three different wrong tool sets, no error at spawn time. If a worker context can't honor an agent's configured tools, failing the spawn loudly would be far better than silently running it crippled.

2witstudios:
Fixes #2460.

## The reporter's diagnosis is wrong, and the difference is the fix

The issue says `enabledTools` is ignored. It is not — the allowlist is applied faithfully
(`page-chat-turn.ts` step 3). The sandbox family is stripped **downstream of it**, by
`filterToolsForSandboxEnablement` (step 3b), which removes every `SANDBOX_TOOL_NAMES` entry
when `pages.sandboxEnabled` is false. The allowlist cannot re-grant them, and the column
defaults to `false`.

The trap: **`update_agent_config` had no `sandboxEnabled` parameter.** It was writable only
through the settings UI's PATCH. An agent configured entirely through tools — exactly what the
reporter did — could hold 24 tool names, have them echoed back intact on every write, and be
permanently `sandboxEnabled: false`. Nothing at any layer mentioned the divergence.

## The two observations the issue could not explain

**"Three spawns, three different wrong tool sets."** Two independent causes, and neither is
randomness. What a worker *reports* having in `'search'` exposure is whatever `tool_search`
returned for the queries it happened to run — the mode hands the model a search over its tools,
not a list of them — so three workers asking different questions describe three different
surfaces while holding the same one. Underneath that, compute eligibility genuinely does vary:
it keys on the payer of the workspace each worker landed in, which is why a spawn now warns
when that workspace will not run code (item 11).

**The "ghost" `read_file` permission entry.** The reporter reasoned that being told
`Tool "read_file" is not permitted for this agent` — for a name they never configured — meant
"the permission surface workers were checked against wasn't even derived from the config I
wrote." It was. `read_file` is not a tool at all (the sandbox tool is `readFile`); the model
hallucinated the name and called `execute_tool` with it, and at that time `execute-tool.ts`
ran its allowlist check *before* its unknown-name check, so an unknown tool was reported as a
forbidden one. That ordering has since been corrected in `execute-tool.ts` — an unknown name
now answers "Unknown tool" with suggestions drawn from what the caller may actually run. No
change needed here, but the inference deserves retiring: the allowlist was theirs all along.

## What changed

One new pure module, `apps/web/src/lib/ai/core/agent-tool-surface.ts`, answers "what of this
stored config survives?" — granted / blocked (with the gate that dropped each) / runtime-conditional /
deferred by exposure mode — and formats one sentence per divergence. Every consumer uses it, so the
config echo and the spawn refusal cannot drift apart.

1. **`update_agent_config` writes `sandboxEnabled`** (AC1), on the same edit permission the
   settings PATCH uses. By the end of this PR, **five doors** write the field under one policy —
   that tool, the settings `PATCH`, the SDK's `PUT`, `POST /api/ai/page-agents/create`, and
   `POST /api/pages` — because an agent that cannot be created with the switch is an agent born
   in the same trap. HTTP doors reject a non-boolean rather than coercing it.
2. **`spawn_session` fails loudly on a self-contradictory config** (AC2):
   `reason: 'agent_tools_ungrantable'`, naming the tools, the gate and the one-call fix, before
   anything is created or dispatched. If the check itself cannot run, it degrades to a warning
   that says the surface was *not checked* — never to a failed spawn.
3. **Every echoing door reports effective, not merely stored** (AC3): `effectiveTools`,
   `blockedTools: [{tool, gate}]`, `toolsNeedingComposerToggle`, `toolsReachedBySearch`,
   `sandboxEnabled`, `toolExposureMode`, plus `warnings[]`. `enabledTools` is still reported
   unchanged beside them — stored and effective, always both, always distinguished. For an
   agent with **no** allowlist the list fields are `null` (unrestricted) rather than ninety
   names or a misleading `[]`.
4. **The gate itself is untouched** (AC4). Nothing was weakened.

### Found while verifying the claims this PR makes

Each of these was found by checking whether something the PR *asserts* is true everywhere it
claims. Each is mutation-checked against its own test file.

5. **The composer-toggle pair is neither granted nor blocked** (codex review). `web_search` and
   `generate_image` never pass through the allowlist — `page-chat-turn.ts` lifts them out before
   it and restores them only for a request whose composer toggle is on, and a dispatched worker
   turn carries no toggles. They report as `toolsNeedingComposerToggle`, never as effective, and
   never refuse a spawn. The set is pinned equal to `ALWAYS_UPFRONT_TOOLS` so drift fails CI.
6. **`enabledTools: []` meant "every tool"** — `filterToolsForAgentAllowlist` reads `[]` as "none"
   and `null` as "unrestricted", and `update_agent_config` folded `[]` into `null`. An agent
   someone tried to lock down came back holding the entire registry.
7. **`sandboxEnabled` was coerced, not validated** (CodeRabbit). `Boolean("false")` is `true` on
   the one field that decides whether a stored sandbox allowlist is granted. Every door that
   writes it now rejects a non-boolean.
8. **ONE SWITCH, EVERY SURFACE — four of five tool sets weren't asking.** A sweep of every
   `pageSpaceTools` consumer (after the second finding, I stopped assuming I'd found them all)
   turned up four surfaces that build their own set from the allowlist and never asked
   `sandboxEnabled`:
   - the `@`-mention / consult engine (`agent-communication-tools.ts`);
   - `POST /api/ai/page-agents/consult`, the HTTP/SDK door onto the same capability;
   - the voice path's **advertised** set (`realtime/system-context.ts`);
   - the voice path's **executable** set (`voice-runtime-deps.ts`) — fixing only the advertised
     one would have been half a gate, since `tool_search` searches the executable set and
     `execute_tool` dispatches from it.

   Precisely what that granted: `canRunCode` still refused every COMPUTE call, so none of these
   was a way into the sandbox. But the chat-side session family is deliberately *not*
   compute-gated, so `spawn_session` and its siblings were genuinely usable from a switched-off
   agent. Voice gates only a BOUND agent; an unbound Global Assistant call has no agent whose
   switch it would be.
9. **An agent could be BORN in the contradiction.** `POST /api/ai/page-agents/create` and
   `POST /api/pages` both accept `enabledTools` and had no way to set the switch — and the
   agent-create response listed the stored names as "Agent has access to N tools: bash, …",
   the exact sentence the reporter believed through three failed spawns. Both now take the
   switch; the create echo reports the effective surface like every other door.
10. **The SDK could neither set the switch nor see the truth.** `agents.updateConfig`'s input is
    a `strictObject`, so `sandboxEnabled` was *rejected before the wire*; its output is a plain
    object, so every effective-surface field and warning the route returns was silently stripped
    before an SDK caller could see it. Both fixed, with the new output fields optional so the
    client still parses a response from a server that predates them. The CLI inherits it:
    `agents config --set` derives its input type from this operation and JSON-parses values, so
    `--set sandboxEnabled=true` sends a real boolean into the route's new strict check — the same
    path an existing test already pins for `--set visibleToGlobalAssistant=true`
    (`packages/cli/src/commands/__tests__/agents.test.ts`, "JSON-coerces non-string values"), so
    this is a checkable claim rather than my reading of the coercion helper.
11. **The tier leg was still silent.** The stored config explains most of the issue but not the
    part the reporter found least explicable — compute eligibility keys on the payer of the
    workspace the worker landed in, so one agent resolves differently in two workspaces. After
    creation, when the workspace is known, a spawn asks the same question the worker's own turn
    will ask (`canRunCode` against that workspace's drive and owner) and WARNS when compute the
    agent is configured for will not be granted there. A warning, never a refusal: no caller can
    fix a payer's tier by spawning differently. It states the outcome and lists what decides it,
    because `canRunCode` folds together the kill switch, the tier and the caller's drive role, and
    naming one would be a guess dressed as a reason.
12. **The diagnostic could fail a spawn that used to work.** A transient read failure in the
    surface check now degrades to a warning that says the surface was *not checked*, rather than
    throwing out of `spawn_session`.

That is every door that writes the field (two creates, three updates) and every tool set built
from it (five).

## Why this is one PR, and what it costs

**Why not split it.** The first four items are the issue. Items 5–11 are what checking each of
this PR's own claims turned up, and most of them are the *same* invariant missing somewhere else.
A gate that answers differently depending on which surface asks is not a gate, so shipping the
switch as settable while three other surfaces still ignore it would leave the PR's own
documentation false on merge. The SDK and create-door work is the smallest piece and the easiest
to argue for deferring — say the word and I will lift it out.

**What it costs at runtime.**
- One indexed PK read per `spawn_session`, for the agent's config, plus — after creation — a
  workspace lookup and one `canRunCode` for the compute question. A spawn is one-per-worker, not
  a hot path.
- One indexed PK read per voice bridge tool call, to read the switch server-side rather than
  carry it on a versioned payload between two separately-deployed services. That choice is
  explained where it is made; the alternative leaves the gate open for the whole deploy window.
- Nothing added to a page-chat turn: it already read the row and already applied the switch.

## A trade-off worth arguing about

A stale sandbox name in `enabledTools` has two causes and the stored config cannot distinguish
them: tools configured for an agent that never had the sandbox (the issue), or a deliberate
revoke — turning the switch off in the settings tab hides those tools from the picker but leaves
them in the saved list, because only edited fields are sent. So the refusal fires in both cases,
and it names both readings with the one-call fix for each rather than assuming.

Pruning the names on revoke would remove the second case entirely. I rejected it: it destroys a
selection the owner may want back when they re-enable, to save them one call. Happy to be
overruled.

## The product expectation in the issue's follow-up

The reporter's follow-up makes a second, sharper point: the worker appears in the sidebar under
the session's workspace, so the UI promises environment inheritance — *"either the worker carries
the environment's surface, or it shouldn't be presented as part of that environment."*

This PR deliberately delivers the loud-failure half and **not** inheritance, because inheritance
is the wrong half. A worker spawned under agent X runs as agent X, and its tool surface is X's
configuration. If a spawn silently widened that to whatever the CALLER can reach, any caller could
grant an agent access its owner never configured — a spawn would become a way to escalate another
agent's permissions. So the answer to "spawned under environment → has the environment's access"
is: no, it has ITS OWN agent's access, and when that contradicts what the config asks for you now
find out at spawn time, in one sentence, with the one call that fixes it.

The reporter's own post-issue verification confirms the diagnosis: adding the sandbox tools
through the UI (which sets `sandboxEnabled`) made every subsequent spawn work, with the same spawn
mechanism that failed three times through the API. The `read_file` "ghost permission" they noticed
is covered too — it is not a tool that exists, and it now reports as `not_registered` with the
reason, rather than as a permission refusal for a name nobody configured.

## Fail vs warn — chosen deliberately (AC2)

**Fail** for the one case that is a contradiction inside the agent's own stored config: sandbox
tools named while `sandboxEnabled` is off. Deterministic, not a runtime accident, one
`update_agent_config` call fixes it either way, and spawning can only produce the crippled worker
the issue asks us to refuse.

**Warn** — on the success payload, as `toolSurfaceWarnings` — for everything else: a name this
deployment does not register (the caller cannot fix it by retrying; on a deployment with
`CODE_EXECUTION_ENABLED` off this is what the sandbox names become, and enabling the switch would
change nothing); `'search'`-mode deferral (nothing is lost); and the composer-toggle pair. The
payer-tier caveat is attached to the CONFIG doors only, never per spawn — it would fire on every
sandbox worker forever while saying nothing about that spawn.

## Does `'search'` exposure contribute to the reported symptom? (AC6)

**Yes, to the appearance of it.** In `'search'` mode the model directly sees only
`CORE_TOOL_NAMES` + `tool_search`/`execute_tool` — precisely the "page tools only" shape the
reporter describes, even with an intact allowlist. It does not explain their case (`tool_search`
for sandbox keywords returned nothing, which is the `sandboxEnabled` strip happening *before*
exposure), and it is not itself a bug. **No behavioral change there.** What changes is that
deferral is reported as its own thing (`toolsReachedBySearch`) and never as a block, because a
deferral reported as a block sends the next reader after the wrong fix.

## Verification

- **Unit**: `agent-tool-surface.test.ts` (17), spawn fail/warn/degrade cases in
  `session-tools.test.ts`, a description-claim pin in `session-tools-contract.test.ts`,
  effective-echo/switch/empty-array cases in `agent-tools.test.ts`, all five doors that write the
  field, the consult engine and the consult route (both asserted on the tool set handed to
  `generateText`), and the voice path (asserted on the advertised tool list, both switch positions
  plus the unbound call).
- **Integration, against a real Postgres and the REAL registry**:
  `agent-tool-surface.integration.test.ts` (12) runs the production deps over real agent rows and
  a real workspace. Its first
  CI run FAILED and was right to: CI leaves `CODE_EXECUTION_ENABLED` off, so the sandbox names were
  absent from the registry entirely and classified `not_registered` — every assertion about the
  switch would have passed for the wrong reason. The suite now sets the switch before the registry
  loads and fails loudly in `beforeAll` if the family is still missing. The kill-switch-OFF world
  has its own unit case, because the right answer there is different.
- **Mutation-checked**, each against its own test file: the sandbox gate in `agent-tool-surface.ts`
  (5 red), the spawn refusal, the degrade-on-failure path, the `sandboxEnabled` write on every door,
  stored-as-effective echo on each echoing door, the empty-array storage, the strict-boolean checks,
  the consult-engine and consult-route filters, the voice exposure filter, the page-service insert,
  and the integration suite's own fixture guard. Every one went red; every one restored.
- **Not** mutation-checked locally: the three voice **executable**-set assertions, which are
  DB-backed and therefore only run in CI. Stated rather than glossed.
- **Typecheck/lint/full test suite deferred to CI** — this machine serialises builds across agents
  and I am not permitted to run them locally. Everything above was run per-file.
- **`packages/cli`'s tests could not be run here at all**: they import `@pagespace/sdk` by package
  entry, which needs its built `dist`, and building it is a slot-gated `tsc`. They DO run in CI —
  `turbo run test:coverage --filter='!@pagespace/processor'` covers every package including
  `sdk` and `cli` — and passed on the tip before this one. So the coverage exists; it just is not
  mine. Saying so rather than leaving the impression every package was exercised locally.
- Wire contract: `session-tools-schema.test.ts`'s pinned literal updated and
  `docs/2.0-architecture/agent-sessions.md` gains §3e in the same PR, per §5's rule.
- **AC5 (end-to-end in a running app with a live agent session) is NOT done** — it needs a web
  build, and build slots are rationed on this machine (concurrent builds have crashed it before).
  Requested; this section will carry the transcript once a slot is granted. Nothing above should
  be read as a live-app claim.

  Being precise about what a slot would add, since the integration suite already drives the
  production deps against a real Postgres and the real registry:
  1. that the app **builds and boots** with these changes — the one thing no test here proves,
     and the place a dynamic-import or module-cycle mistake would surface;
  2. `PUT /api/ai/page-agents/:id/config` returning the effective echo over real HTTP with a
     seeded session and CSRF token — reachable with curl, no model involved;
  3. the `spawn_session` refusal as a genuine tool call. `update_agent_config` and
     `spawn_session` are AI-tool-registry tools, not SDK operations, so the MCP/CLI surface
     cannot invoke them — reproducing the reporter's exact flow needs a real model turn against
     a configured provider key. Without one, (1) and (2) are what a slot buys.

## Deliberate boundaries

- **`send_session` does not warn.** A send addresses a worker that already exists and whose
  surface was reported when it was spawned; repeating the same sentence on every message is how
  warnings stop being read. The spawn is where a worker is created under a config, so it is where
  the config is judged.
- **The refusal is only for the self-contradiction.** Every other divergence warns, because no
  caller can spawn their way out of an unregistered tool name, a search-mode deferral, or a
  workspace's payer tier.
- **Nothing prunes `enabledTools`.** See the trade-off section above.
- **`sandboxEnabled` is NOT added to the page state hash**, though every sibling agent-config
  field is in `PageStateInput`. Adding it would change the computed hash for every page on its
  next mutation, and consumers that compare a stored hash against a recomputed one would read
  that discontinuity as a conflict — real blast radius, for no benefit here: undo/redo restores
  from the activity log's `previousValues`, which are populated from `updatedFields` independently
  of the hash. Flagging the asymmetry rather than quietly widening it.

## Not in scope, flagged

`api/v1/chat/completions/route.ts` and `workflows/workflow-executor.ts` already apply the switch,
but they rebuild the same filter recipe by hand — four copies of one pipeline, now with a fifth
(voice) that also assembles its own. Worth one shared assembler, deliberately not bundled here.

https://claude.ai/code/session_01R8ZEaEgGRA8n7PW22kGgep


2witstudios:
**Since the last reviews (codex @ 441c769, CodeRabbit @ same):** the PR roughly tripled, and not by scope creep — every addition came from checking a claim this PR itself makes. Rather than re-reading 30 commits, the short version:

**Both review findings are addressed** (composer-toggle pair → its own `conditional` category, pinned equal to `ALWAYS_UPFRONT_TOOLS`; `sandboxEnabled` rejected-not-coerced, applied to *both* HTTP doors rather than the one flagged).

**What checking the claims turned up — four more surfaces and two more doors:**
- The `@`-mention engine, `POST /api/ai/page-agents/consult`, and the voice path's **two** independently-built tool sets (advertised *and* executable) all applied the allowlist and never asked `sandboxEnabled`. Fixing only the advertised voice set would have been half a gate — `tool_search` searches the executable one and `execute_tool` dispatches from it.
- Precisely what that granted: `canRunCode` still refused every **compute** call, so none was a way into the sandbox — but the chat-side session family is deliberately not compute-gated, so `spawn_session` and friends were genuinely usable from a switched-off agent.
- Agents could be **born** in the contradiction (`page-agents/create`, `POST /api/pages`), and the create response listed the stored names as "Agent has access to N tools: bash, …" — the exact sentence the reporter believed through three failed spawns.
- The SDK's `strictObject` input **rejected** `sandboxEnabled` before the wire, and its output schema stripped the new effective-surface fields.

**One new behaviour worth a look:** a spawn now warns when the *workspace* the worker landed in won't run code (item 11) — the tier leg, which is the real reason one agent produced three different tool sets and looked random. Warning, never refusal; states the outcome rather than naming a cause, since `canRunCode` folds three questions together.

**Two decisions I'd like pushback on if you disagree:** the refusal fires on a deliberate sandbox *revoke* too (the stored config can't distinguish that from the issue's case, so it names both readings instead of pruning the stale names), and this is one PR rather than three — the SDK/create-door work is the piece I'd lift out first if asked.

Full detail in the description. AC5 (live app) is still unmet and marked as such: it needs a build slot on a machine that serialises them.

2witstudios:
@coderabbitai review

The PR has grown substantially since your last pass (rate-limited on the intervening pushes). Since then: four more tool-assembly surfaces now apply the per-agent sandbox switch, two agent-creation doors accept it, the SDK contract exposes it, and a spawn warns when the workspace itself will not grant compute. Your `sandboxEnabled` coercion finding is fixed on both HTTP doors, plus the two creation doors that came later.

Areas where a fresh pass would be most valuable:
- `apps/web/src/lib/ai/realtime/voice-runtime-deps.ts` — the switch is read server-side per bridge call rather than carried on the versioned payload; I argue the alternative leaves the gate open for a whole deploy window, but the cost is a DB read per voice tool call.
- `apps/web/src/lib/ai/core/agent-tool-surface.ts` — the null-vs-empty semantics (`effectiveTools: null` means unrestricted, `[]` means none).
- `apps/web/src/lib/ai/tools/session-tools.ts` — the fail-vs-warn split, and whether refusing on a deliberate sandbox revoke is the right call.

PR #2485, opened 2026-08-24
fix(agents): a long agent job no longer wedges itself partway through (#2461)

### LINKED ISSUE #2461: [agents] Worker session execute_tool calls fail ("tool_name stripped from envelope") after sustained high-volume tool use
## Context
- Date: 2026-08-22
- Worker spawned via `spawn_session` (no agent param — running under the conversation's agent, full tool surface, same session sandbox).

## What happened
The worker's job: read ~30 JSON chunk files (`readFile`) and apply each to a SHEET via `edit_sheet_cells` (~240 cells per call, payloads of several hundred KB each).

- First ~4 files applied fine (960 cell writes, roughly 1MB of cumulative payload through the session).
- After that, **every** subsequent `execute_tool` call failed with the identical validation error:

```
Invalid input for tool execute_tool: Type validation failed.
Error: [{ "expected": "string", "code": "invalid_type", "path": ["tool_name"],
"message": "Invalid input: expected string, received undefined" }]
```

- ~10 consecutive retries across differently-structured calls — same error every time. The `tool_name` field was being stripped from the envelope before validation.
- Earlier identical calls in the same session had succeeded, so it wasn't the payload shape.

## Recovery
A freshly spawned worker picked up from the next chunk and completed the remaining 26 files (~6,000 more cell writes) with zero transport failures, including individual 240-cell calls.

## Suspicion (experience-level)
Something in the worker session transport degrades after a cumulative volume of large tool calls — envelope fields silently dropped. Correlates with either total payload bytes or call count; a fresh session resets it.

## Repro candidate
Spawn a worker; have it make ~30 sequential `execute_tool` calls, each carrying a few-hundred-KB JSON payload; watch for envelope corruption (tool_name stripped) partway through.

2witstudios:
Fixes #2461.

**Repro-first.** No code was changed until the failure was reproduced. The reproduction is checked in at `apps/web/src/lib/ai/tools/__tests__/execute-tool-envelope-repro.test.ts` and runs in ~50ms with no provider.

## What actually happens

The reporter's phrase "tool_name stripped from the envelope" is a symptom description, not a mechanism. There is no parent→worker envelope carrying `tool_name`: `spawn_session`/`send_session` ship only a signed `{ text }` and the worker runs its own model loop. The failing `tool_name` is emitted by the **worker's own provider**.

Driving the real `streamText` loop against a mock provider reproduces the reported error exactly when the provider emits an `execute_tool` call carrying **no argument tokens at all**. The AI SDK coerces that to `{}` before validating:

```js
// ai@6.0.212, parseToolCall → doParseToolCall
const parseResult = toolCall.input.trim() === ''
  ? await safeValidateTypes({ value: {}, schema })   // ← the reported path
  : await safeParseJSON({ text: toolCall.input, schema });
```

so zod reports the required `tool_name` as missing, and `parameters` (which has `.default({})`) reports nothing. That is the reported error field for field, and the `Value: {}` in the message is the tell. **Nothing stripped a field — the model sent no arguments whatsoever.**

The obvious competing suspect is ruled out by test: a payload truncated **mid-argument** fails as `JSON parsing failed`, a *different* error. So the incident is specifically the zero-argument-tokens case.

## Does it actually fix #2461?

The reporter's agent **alternated**: `readFile` a JSON chunk (a several-hundred-KB result), then `edit_sheet_cells` it (a several-hundred-KB argument), ~30 times. Driving that exact mixture through the real `streamText` loop, capped and uncapped:

| | uncapped | capped |
|---|---|---|
| step 5 | **1.14 MB** (~285k tokens) | 0.58 MB |
| step 30 | 8.24 MB | 0.58 MB |
| peak | grows without bound | flat |

Step 5 is the number that matters. The issue says *"the first ~4 files applied fine — after that, **every** subsequent `execute_tool` call failed."* At step 5 the uncapped transcript is past a 200k-token window. The context-exhaustion account **predicts** that failure point rather than merely permitting it, which is the difference between a diagnosis and a story that happens to fit. Capped, the same workload stays flat at 0.58 MB — 14× smaller by step 30, and bounded rather than linear.

## The trigger, measured

Compaction and elision both run once per **turn**, in `prepareHistoryForModel`. A turn is up to `AGENT_MAX_STEPS` (100) model calls, and between steps `prepareStep` only re-marked cache breakpoints. So the transcript handed to the provider grew monotonically for the whole turn. Measured, driving the real loop with ~103 KB payloads:

```
prompt bytes per step: 71, 103746, 207421, 311096, 414771, 518446, 622121, 725796, 829471, 933146
```

Dead linear at **+103 KB per step**, every byte of it the model's own arguments. At the reporter's several-hundred-KB payloads that exhausts a 200k-token window in about four calls — which is exactly where their worker stopped. A fresh worker starts from zero, which is why one recovered the job.

## Verdicts on the two leads I was handed

**Lead 1 (un-elidable context growth) — right family, wrong mechanism.** Confirmed: `execute_tool` *is* excluded from `DEFAULT_ELIDABLE_TOOLS`. But **adding it would not have fixed this**, for a reason that has nothing to do with the list: elision, compaction and `capToolResultSize` all run **between turns**, in `prepareHistoryForModel`, and this is an **in-turn** problem — one turn is up to 100 model calls, and nothing ran between them. Listing `execute_tool` would not have changed that, and the exclusion is correct as written anyway (in search-exposure mode it dispatches writes whose results must not be replayed), so it stays.

Two further corrections to the brief's framing: compaction is **not** admin-gated — `canUseCompaction` is now `user != null` and the worker path does pass a user, so that comment is stale. And elision would have been the wrong tool even in-turn for half of this, since it replaces tool **outputs** while the argument half of the growth is tool **inputs**.

**Lead 2 (persisted `input: undefined` replaying) — ruled out.** `foldChunksToParts` persists a `tool-input-error` as `state: 'output-error'` (not `input-error`), and `sanitizeMessagesForModel` keeps only `state === 'output-available'`, so the part is dropped before the next turn ever reaches the model. It cannot replay as `tool_name: undefined`. The repeated failures were also *within one run*, where the SDK already feeds a `tool-error` result back to the model (`ai/dist/index.mjs:4792-4804`) — so the model did see the error; it just could not act on it, which is the other half of this fix.

## The fixes

**1. Bound in-turn growth — the cause.** New pure module `cap-step-tool-payloads.ts`, wired into the `prepareStep` seam of **every** multi-step tool loop in the codebase — all six:

| call site | loop | budget |
|---|---|---|
| `lib/ai/chat-pipeline/global-chat-turn.ts` | `streamText` | `stepCountIs(100)` |
| `lib/ai/chat-pipeline/page-chat-turn.ts` | `streamText` | `stepCountIs(100)` |
| `app/api/v1/chat/completions/route.ts` | `streamText` | `stepCountIs(100)` |
| `lib/ai/tools/agent-communication-tools.ts` | `generateText` | `stepCountIs(20)` |
| `app/api/ai/page-agents/consult/route.ts` | `generateText` | `stepCountIs(20)` |
| `lib/workflows/workflow-executor.ts` | `generateText` | `stepCountIs(100)` |

Their no-tools sibling branches are left alone: with no `tools` the model cannot emit tool calls, so there is nothing to accumulate.

It caps **both directions**, because both grow. Measured on the real loop: arguments the model writes accumulate at ~+103 KB/step, and results it reads at ~+100 KB/step — the same dead-linear shape. The reporter's agent did both at once (`readFile` a chunk, then `edit_sheet_cells` it), so capping only arguments left #2461 reproducible for anything read-heavy.

What survives is deliberately asymmetric:

- **Arguments** — everything but the newest *step's* calls. Arguments are content the model itself just wrote, with the result of that call still in front of it, so an older payload is the cheapest thing in the transcript to give up. Exemption is by message rather than by count, because one step can issue several tool calls in parallel and they all land in a single assistant message — keeping one of three and capping its siblings would be an arbitrary line through a single step.
- **Results** — everything but the newest few *steps*. A result is information the model *read* and may still be consuming; a gather-then-act agent reads two or three files before writing anything, and a window of one would break it mid-thought. Retention is by step rather than by result part, because one step can read four files in parallel and land four results in a single tool message — counting parts would keep three of those four and stub the rest, breaking exactly the agent the window protects. What matters is that retention is constant in the number of steps, not that three is magic.

An output's own `type` is preserved when stubbing — `error-text` tells the provider the tool failed, and rewriting it to `text` would silently turn a failure into a success.

**One boundary worth a reviewer's opinion — now measured, not asserted.** `content` outputs (the multimodal array `read_page` returns to a vision-capable model, bulk = base64 image bytes) are not capped. I originally wrote that up as a blanket gap; measuring it showed the gap is narrower:

| path | per step | at step 8 |
|---|---|---|
| via `execute_tool` (search-exposure — where high-volume agents live) | **capped** | flat **1.53 MB** |
| `read_page` as a top-level tool (`content` output) | +0.51 MB | 3.56 MB, linear |

`type: 'content'` comes from `read_page`'s own `toModelOutput`, so it only appears on the top-level path; through `execute_tool` the image arrives as ordinary JSON and is capped like anything else. So the agent shape that caused #2461 is covered.

Worth adding that capping there costs nothing: `execute_tool` has no `toModelOutput` of its own, so the image reaches the model as **base64 text inside a JSON blob, never as an image content part**. A vision model cannot see it either way. Capping that payload removes bytes the model could not use — which is a separate pre-existing limitation of vision-via-`execute_tool`, not something this PR introduces, but it does make the capping unambiguously safe on that path.

The top-level path is left alone deliberately: a stub is no substitute for an image the way it is for re-readable text, image delivery already has its own design in `read-page-vision-output.ts`, and what a mid-comparison vision agent may lose is a product call. Happy to extend it if you'd rather — it's a small change. (`execution-denied` carries no value, so there is nothing to cap.)

It preserves tool call ids, names, and the object-ness of the argument value so call/result pairing survives conversion, and it is byte-stable once capped, so cache breakpoints still work: each oversized payload changes bytes exactly once, when it falls out of the retention window.

**Capping only rewrites what is SENT.** The run's recorded steps keep every payload at full size, so persistence and the activity log see exactly what the agent really sent and received — asserted by a test, because getting this wrong would silently destroy history.

The two chat turns also re-mark cache breakpoints per step, so they share `agentLoopPrepareStep` in `context-assembly.ts` — the per-step counterpart to `prepareHistoryForModel` in the same seam. That extraction was forced by CI: writing the wiring into both turns tripped the repo's `turn-duplication-ratchet`, which forbids adding byte-identical lines between `page-chat-turn.ts` and `global-chat-turn.ts`. It was right to, and the result is a **net reduction** — 161 shared lines, one below master's 162, with the ratchet constant and both prose homes moved to match.

A source-scan merge guard (`lib/ai/__tests__/tool-payload-cap-callsites.guard.test.ts`, following this repo's existing `gate-callsites.guard.test.ts` idiom) brace-matches every `streamText`/`generateText` call, selects those that declare a step budget **and** pass `tools`, pins the set at those six, and fails the build if one stops capping. This was a missing seam rather than one broken loop; a seventh loop added later would otherwise reintroduce #2461 silently.

The threshold (24 000 chars) is deliberately looser than the existing 8 000-char cap on tool *outputs* — an output is a refetchable read, an argument payload is content the model authored and may still be reasoning about. The job is to stop one turn's payloads from ending the turn, not to trim ordinary writes.

**2. Make the residual failure honest — the legibility bug.** Zod's stock "expected string, received undefined" reads as *"you forgot a field"*, which is a lie about what happened and, worse, unactionable: an agent that believes it sent `tool_name` has nothing to change, so it re-sends the identical call. That is the reporter's ~10 identical failures.

The message must stay accurate for two different causes, because it cannot tell them apart: the callback is on the **field**, and a zod field-level issue carries only the value at its own path, so `{}` (nothing arrived) and `{"parameters": {...}}` (arguments arrived, the name did not) both surface as `tool_name: undefined`. Object-level `.check()`/`superRefine` is not an escape hatch either — zod does not run it when the shape parse fails. So the wording claims only what is true in both cases and gives each cause its own concrete recovery:

> execute_tool requires tool_name, and this call did not carry one. If you did set it, the response was cut off before the arguments finished — an oversized parameters payload is the usual reason, and retrying unchanged would fail the same way, so re-send with fewer items per call. If you left it out, re-send the same call with tool_name set.

A genuinely wrong-typed `tool_name` still gets its own plain message.

The provider-facing JSON Schema is **unchanged** — verified byte-for-byte and pinned by a test, so the fix cannot manufacture the omission it explains.

## CI

All 10 checks green on `ea2e570` — build, typecheck, lint, the full unit suite, the agent-session E2E stories, and every security job. `mergeStateStatus: CLEAN`, no conflicts, branch contains master's tip.

## Verification

- 48 tests green across the reproduction, the cap module, the call-site guard, and the existing `execute-tool` suite.
- **Mutation-checked throughout, one test file at a time — every mutation caught.** Reverting the cap to a no-op fails the growth test; reverting the schema to `z.string()` fails both message tests; reverting to the old over-claiming wording fails the omitted-field test; stripping the cap from `workflow-executor.ts` turns the call-site guard red and names the file; five mutations of the argument path (dropping the newest-call exemption, the size threshold, the no-op short-circuit, the object-shaped stub, the tool call id) four of the result path (removing it entirely, collapsing the retention window, dropping the newest-call exemption, stubbing `error-text` as plain `text`), and restoring the single-position exemption that mishandled parallel calls, each turn the suite red.
- The read-heavy end-to-end test runs the identical loop **twice, capped and uncapped**, so it asserts a measured difference rather than a threshold chosen to pass.
- **Typecheck, lint and the full suite run on CI, not locally** — concurrent local builds are restricted on this machine, so CI is the authoritative gate for those.

  That gate earned its keep: it caught a type error vitest structurally cannot, because vitest strips types without checking them. `part is { type: string; output?: unknown }` reads as equivalent to `part is ToolResultPart` and is not — *every* member of the content union satisfies `{ type: string }`, so narrowing kept all of them and `part.output` failed on `TextPart`. Fixed by narrowing to the SDK's own `ToolCallPart` / `ToolResultPart`. Worth noting for confidence in the rest: that was the **only** type error across the whole change, so the six call sites, the rename and the guard all typechecked on the first pass.

**Review found two things I got wrong; both are worth knowing about.**

1. *Codex* — the cap reached two of six loops. Chasing that turned up that it also covered only one of two payload directions.
2. *CodeRabbit* — results were retained by individual part rather than by step, so four parallel reads in one step would lose one. I had fixed precisely this for *calls* one commit earlier and then argued in the code that results should stay by-count. The argument was wrong and I'd have kept believing it.

A mutation check also caught the call-site guard passing on a loop whose wiring had been **deleted** — the comment above it still named the helper, and the guard matched the prose. Slices now strip comments and the check requires a call form. A guard a comment can satisfy is the same failure this whole PR is about: a mechanism that reports success without doing anything.

**On running-app verification.** I never ran the app locally — concurrent builds are restricted on this machine. But that gap is smaller than it first looks, and I'd rather be precise than leave a vague caveat hanging: the `E2E (agent-session user stories)` job does a production build, boots web + realtime + proxy, waits on `/api/health` and a live same-origin socket.io handshake, and then drives the agent-session specs with Playwright against a mock provider. That is a running app exercising this code path, and it passes.

What genuinely remains unexercised is a **real** provider at the payload volume that triggered #2461 — thirty several-hundred-KB calls in one turn. I can't stage that honestly on a laptop, and a build slot wouldn't get me there either, so I'm not requesting one. The measured bound in the table above is the evidence for that case; the E2E job is the evidence that the wiring works in a live stack.

https://claude.ai/code/session_01USWfw4mwrDriSHgqdbLWdU


2witstudios:
@coderabbitai review

Context for the review: the first attempt hit the rate limit at `8352254` and never ran. The PR has changed substantially since — it now caps tool payloads in **both** directions (arguments and results) across all six multi-step model loops, not two, and adds a source-scan call-site guard.

Two spots where a second opinion would be most valuable:

1. `cap-step-tool-payloads.ts` — the retention asymmetry. Arguments keep only the newest step's calls; results keep the newest few, because a gather-then-act agent may still be reading them. Is that window defensible, and is preserving the output's `type` when stubbing (so `error-text` cannot silently become a success) handled correctly for every member of the union?
2. `content` (multimodal) outputs are deliberately **not** capped — this is a known, stated gap, reasoning in the module docblock and the PR description. Worth challenging if you disagree.

2witstudios:
@coderabbitai Thanks — that P1 is correct, and it lands on exactly the point I reasoned my way into getting wrong. Fixed in `3091f04`.

The commit before yours fixed the parallel case for tool **calls** (exempt the newest *message*, not the newest position) and I explicitly argued results should stay exempted by count. Your counterexample kills that argument: one step reading four files in parallel delivers four result parts in a single tool message, so `resultPositions.slice(-3)` kept three and stubbed the fourth — breaking the gather-then-act agent that is the entire reason results get a window and arguments do not. I had made the same mistake in both directions and only fixed one.

Retention is now by tool-result **message**, so every result part of a retained step survives however many that step produced. `KEEP_RECENT_TOOL_RESULTS` is renamed `KEEP_RECENT_RESULT_STEPS` — a constant that counts steps shouldn't be named for parts.

Both tests you asked for:
- more oversized parallel results in the newest step than the window has slots → all kept
- steps behind the window still expiring whole → still capped

Mutation-checked: restoring the `slice(-N)` parts form turns the parallel test red, and collapsing the window to zero steps turns four tests red.

Noted also that you confirmed the capped output-type handling is correct for every variant, and that you're not flagging the `content` exclusion. That was the judgement call I was least sure of, so an independent read agreeing the boundary is documented and needs a product decision is genuinely useful.

One caveat for anyone reading: your review ran against `3716f08`, and the branch has moved twice since — the shared `agentLoopPrepareStep` seam (extracted because CI's turn-duplication ratchet caught me copying wiring into both chat turns) and now this fix.

PR #2486, opened 2026-08-24
fix(documents): line-range edits corrupt html-mode documents and lie about line counts (#2463)

### LINKED ISSUE #2463: [documents] Line-range edits on html-mode DOCUMENTs corrupt non-HTML content and misreport line counts
## Context
- Date: 2026-08-22
- Pages: `Lead Gen/registry` — a DOCUMENT with `contentMode: html` (created before content modes existed) holding raw JSON, ~81 lines. Also `Lead Gen/_index`, html-mode, holding markdown report text.

## What happened
Called `replace_lines(1, 81)` on the registry with a ~91-line JSON payload. Response: success — `linesChanged: 81, newLineCount: 9`.

The result was corrupt: most of the new content landed, but a stale duplicated tail from the old content survived at the end (9 lines of a half-written old entry), leaving the JSON invalid. It took a follow-up full read + a second delete pass to repair. The reported `newLineCount: 9` for a 91-line payload was itself the tell that line accounting was off.

Second case, same session: `replace_lines(1, 18)` on the html-mode `_index` doc with an 18-line markdown payload reported `totalLines: 1` afterward while the content was ~18 lines.

Both docs behaved correctly once I did a full read and full-range rewrite — but the intermediate corrupt state is exactly what an agent leaves behind if it doesn't verify after every edit.

## Suspicion (experience-level)
The html content pipeline appears to normalize/re-chunk lines for storage (wrapping, paragraph assembly?), so client-visible line numbers and the stored representation drift. Payloads that aren't valid HTML fragments (raw JSON, markdown) hit this hardest — nothing validates content-shape on write. Markdown-mode docs edited with the same calls behaved exactly as expected.

## What would have helped
- Warning/validation when `replace_lines` targets an html-mode doc whose content isn't HTML, or treating line numbers as unreliable there
- New machine-written docs defaulting to markdown mode (these predate me and defaulted to html)

2witstudios:
Fixes #2463.

An agent edited an html-mode DOCUMENT holding raw JSON, was told `linesChanged: 81, newLineCount: 9` for a 91-line payload, and left the page holding invalid JSON: most of the new content applied, a stale tail of the old content surviving. Reported as success.

## What was actually wrong

**The count was measured on the wrong thing, on both surfaces.**

- `apps/web/src/lib/editor/line-edit.ts` pushed the replacement into the line array as ONE element, so `newLineCount` was `surrounding + 1` no matter how many lines the payload held. 8 + 1 = the reported `9`.
- `apps/web/src/app/api/mcp/documents/route.ts` did split the payload — but then ran `addLineBreaksForAI` a **second** time over content that already had newlines, while returning a `totalLines` computed **before** that pass. So the route told the agent a number the next read contradicted. That divergence is the mechanism that produced the reporter's stale tail: a document reported as N lines was N+8 on disk, and the next `replace(1, N)` left 8 lines behind.
- `addLineBreaksForAI` only knew block tags. A document laid out with `<br>` separators has none, so it came back with zero newlines added → `totalLines: 1` for an 18-line document. That is the second symptom in the report.
- `create_page` hard-forced `contentMode: 'html'` at both the repository call and the response echo, so an agent that did not know to ask for markdown always landed on the fragile path.

## The fix

**One rule: the stored content IS the line projection, and the reported count is measured on it.**

`line-edit.ts` is now the shared line-accounting core for both surfaces — `replaceLines`, `insertLines`, `deleteLines`, plus `projectLines`/`canonicalizeForLineEditing`. It splits multi-line payloads, canonicalizes the assembled document exactly once, and takes `newLineCount` from that stored string. The MCP route calls into it (its own three implementations and the double-normalization are gone) and reports `totalLines`/`previousTotalLines` from the result. A read of the page therefore cannot return a different count than the write reported — that identity is asserted directly in tests by writing, feeding the stored content back, and reading.

**`addLineBreaksForAI`:**
- `<br>` and `<hr>` now break lines (plus `dl/dt/dd/caption/address/details/summary/hgroup` added to the block list).
- The adjacent-tag pass used `\s*`, which **swallowed** a real newline between two blocks — a removal, in a function whose contract is that it only adds. Now `[^\S\n]*`. This matters more than cosmetics: idempotence is what makes "store the projection" safe.

**Loud failure instead of silent partial application:**
- New optional `expectedTotalLines` on all three operations — the count you last read. A mismatch is refused: 409 on the route, a structured `success: false` in the AI tool, `--expect-lines N` in the CLI. This is the reporter's exact scenario turned into a refusal.
- Out-of-range edits now answer with the real line count (400 + `totalLines`) rather than a bare message, and neither refusal writes anything.

**Content-mode mismatch is now stated, not discovered:** an html-mode DOCUMENT whose content has no HTML block structure returns `contentModeWarning` on every read and every write. It deliberately does **not** tell the agent to call `/api/pages/{id}/convert-content-mode` — that route is session-auth only, so an MCP principal would just get a 401. (Widening that route's auth is real permission work and out of scope here; flagged rather than half-done.)

**Machine-written documents default to markdown:** both hard-forcing sites in `page-write-tools.ts` now use one `resolvedContentMode` (they could previously disagree — `type === 'DOCUMENT'` at the write, `isDocumentPage(type)` at the echo), and `POST /api/pages` defaults DOCUMENTs to markdown **when the caller is an MCP/API-key principal**. Browser sessions keep html — the rich-text editor *is* html, and flipping the UI default is a different change. Explicit `contentMode` always wins; existing pages are untouched and keep working (round-trip tests cover html-mode pages holding HTML, JSON and markdown).

**SDK/CLI:** the SDK's `z.object` output schemas would have silently **stripped** `contentModeWarning`/`previousTotalLines` on the way to the CLI, and `strictObject` inputs would have rejected `expectedTotalLines` outright — both modelled now, and the CLI prints the warning to stderr on `pages read` and `pages replace-lines`.

## How this was verified

Unit + route tests, and **mutation checks** on the line-counting logic (criterion 6):

| Mutation | Result |
|---|---|
| `newLineCount` from the pre-normalization array | **3 tests red** (core + route) |
| Replacement back to a single array element, count from the array (the original #2463 shape) | **11 red** across all three suites |
| Staleness guard made a no-op | **14 red** |
| `<br>`/`<hr>` pass removed | **6 red** |
| adjacent-tag regex back to `\s*` | **1 red** (blank-line preservation) |
| `contentMode` default back to html | **2 red** (tool + API route) |
| Route stores double-normalized content again | **7 red** |

One mutation **survived** and is worth stating: pushing the replacement back into the array as a single element *without* also changing the count computation stays green — because the count is now measured on the joined, stored content, so the array's shape no longer affects it. That is the fix working as intended (the arithmetic no longer depends on that detail), not missing coverage; the combined mutation above is the one that reproduces the reported bug, and it goes red loudly.

Suites run locally (single files / directories only): `line-edit`, `line-breaks`, `text-edit`, `page-serializer`, `page-write-tools`, all `mcp/documents/__tests__`, `api/pages/__tests__/route`, plus the full CLI (1229) and SDK (804) suites, and the wider `src/lib/ai`, `src/lib/workflows`, `src/lib/webhooks`, `src/components/workflows` sweep — 3642 passed, 3 failures all pre-existing DB-env-only ("could not reach Postgres" / `role "test" does not exist`).

**Not run locally: typecheck, lint, and the full web suite** — the machine owner has serialised builds/typechecks across agents and these are deferred to CI, which is the gate. Live-app verification is likewise pending a build slot; everything here is verified under vitest against the real route handler and the real tool `execute`, including the write→read round trip.

## Follow-ups since the first push

- **Lint fix (all four red checks, one root cause).** `isCodePage` became unused in `page-write-tools.ts`; `next build` runs ESLint, so it failed Lint & TypeScript, Unit Tests, E2E and Static Security Analysis identically.
- **Self-review, before any reviewer asked:** the adjacent-tag pass was *dropping* the horizontal whitespace between two blocks while the file's header promises it only ADDS (the "never removes characters" test now covers the spaced and tabbed forms and goes red if the replacement is reverted); the structural-markup regex is built once instead of per call; `expectedTotalLines` is `.int()` on the route, matching the SDK — a fractional value could never equal a line count, so it would have refused every edit that supplied one.
- **Review thread from `chatgpt-codex-connector` (P1), addressed in b2ef45f.** The code-shipped `writing-documents` skill still told agents `create_page` defaults DOCUMENTs to html — so an agent following the *required* skill would omit `contentMode`, believe it had an html page, and write HTML into what is now markdown. That skill, the DOCUMENT bullet in `inline-instructions.ts`, the SDK's `pages.create` contract note and the MCP documents API doc all say markdown now. The same skill body had two more claims this branch had just falsified — the line-breaker documented as block-tags-only, and whitespace between blocks described as collapsed — both corrected, along with the new guard, the warning, and the count guarantee.

## A self-review pass found the fix had opened a hole the size of the bug

Worth reading if you review nothing else here. Adding `<br>`/`<hr>` to the normalizer, combined with "the stored content IS the projection", meant:

```
{"leads":[{"note":"call<br>then email"}]}   →   {"leads":[{"note":"call<br>\nthen email"}]}
```

A newline injected **inside a JSON string value**, and the write path stores it — the page ends up holding invalid JSON, which is #2463's own failure mode, reproduced by the fix for it. The gate was "does the content contain a tag anywhere", which JSON carrying scraped markup passes. (The pre-existing block-tag passes did the same to `{"body":"<p>x</p>"}`, so this was latent before the branch too.)

The answer: **an HTML document opens with a block element** — Tiptap wraps everything — so `looksLikeHtmlDocument` anchors the test at the start of the string, and gates both the normalizer and the mismatch warning. The warning had the matching hole: it stayed silent for exactly the JSON-carrying-markup case it exists to flag, because it used the same weak test. Mutation-checked: reverting to the tag-anywhere sniff turns the normalizer test and the route round-trip red.

Three more from the same pass:

- `<hr>` **opening** a document gained a blank line 1, renumbering everything below it against a read the agent had already done; a document **ending** in `<br>`/`<hr>` gained a phantom trailing line. Both guarded.
- `POST /api/pages` applied the markdown default without looking at the `content` it also accepts, so an existing API-key client posting Tiptap HTML with no `contentMode` got a **markdown page holding HTML** — one line, unusable for line editing, and unwarned, since the warning only looks the other way. The default reads the body now.
- The in-app `read_page` returned no `contentModeWarning` while the MCP `read` did, so an agent following the skill's own read-then-write rule would meet the warning only *after* writing. It is on the read now.

## Review rounds

- **Codex (P1)** — the code-shipped `writing-documents` skill still told agents `create_page` defaults to html. Fixed in b2ef45f, along with two more claims this branch had just falsified in the same file; replied in the thread. Left open for the reviewer, since I fixed it rather than confirming it was already fine.
- **CodeRabbit (2 × P2)**, both correct, fixed in 9454a68: `replace_lines`' refusal carried the real line count only inside `message` while the comment above it promised a field and the MCP route returned one — `totalLines` is a field on both surfaces now; and `expectedTotalLines` was `z.number()` in the tool against `.int().min(0)` on the route and SDK, so `3.5`/`-1` parsed and then refused every edit forever. Mutation-checked both.

## Known behavior a reviewer should weigh, not discover

Storing the canonical projection means an html-mode payload is line-broken before it is saved, including inside `<pre>` — so a code block written by the AI tool can gain a trailing newline it would not have had. This is not new: the MCP route (the surface agents actually use) has always stored `addLineBreaksForAI` output, so the behavior is now consistent across both surfaces rather than differing between them, and it is idempotent so it cannot accumulate. Making the normalizer skip `<pre>` regions would change line numbering for every existing page and needs verification in the running editor, which is a separate change.

I also declined to add a symmetric warning for HTML written *into* a markdown-mode page: markdown-mode content is rendered through `marked`, which passes raw HTML through, so "HTML in a markdown page" is not reliably wrong and I will not ship a warning I have not verified against the running renderer. The skill documents it instead.

https://claude.ai/code/session_01DC1j7xVprYDtggiH8o1MTd


2witstudios:
@coderabbitai review

The first review was rate-limited before it ran. The branch has moved a fair amount since — the substantive part to look at is `apps/web/src/lib/editor/line-breaks.ts` (`looksLikeHtmlDocument`, which decides whether content is normalized at all and therefore whether a write can reshape it) and `apps/web/src/lib/editor/line-edit.ts` (the shared line-accounting core both editing surfaces now go through).

2witstudios:
Both fixed in 9454a68 — you were right on both counts.

1. `totalLines` is now a field on the refusal, not prose inside `message`. It is `projectLines(page.content, isRawText).length`, i.e. the same canonical projection the count would have come from on a successful edit, so the two surfaces answer the same question the same way. Both existing failure tests assert it (89 for the stale case, 3 for the out-of-range one).
2. `expectedTotalLines` is `z.number().int().min(0)` in the tool now, matching the route and the SDK. Added a schema test that 3 parses while 3.5 and -1 do not — the failure mode you describe (parses, then refuses every edit forever) is worse than a plain rejection.

Mutation-checked: reverting either change turns its test red.

PR #2568, opened 2026-09-08
fix(ios): correct the AASA, and prep 1.4 for App Store review

2witstudios:
## Why

The iOS app (`ai.pagespace.ios`) has shipped to TestFlight but **has never been through App Review**. It is a 100%-remote WebView (`server.url = https://pagespace.ai/dashboard`), which puts **Guideline 4.2** squarely in play — and `apps/ios/README.md` names universal links as part of the defense against it.

Universal links have never worked. Verified live, not inferred from source:

```
$ curl https://pagespace.ai/.well-known/apple-app-site-association
{"applinks":{"apps":[],"details":[{"appID":"TEAM_ID.ai.pagespace.app", ...
HTTP 200  content-type: application/octet-stream
```

Three defects in one file: an unsubstituted `TEAM_ID`, the wrong bundle id (`ai.pagespace.app` vs the real `ai.pagespace.ios`), and a non-JSON content type.

## What changed

**Universal links.** Caddy routes `/.well-known/*` to marketing (only `oauth-*` reaches web), so `apps/marketing/public/.well-known/apple-app-site-association` is the copy that serves — corrected to `M96WTV3CKX.ai.pagespace.ios`. The web copy was unreachable and is deleted, leaving one source of truth. `next.config.ts` now sets `application/json` on that path, since the file is deliberately extensionless and Next was defaulting to `application/octet-stream`.

`paths` is narrowed to `/invite/*` — the only one of the three claimed routes that exists (`/join/*` and `/open/*` both 307 to sign-in). The OAuth callback paths from the old web copy are deliberately **not** carried over: native sign-in returns through the `pagespace://auth-exchange` custom scheme, so claiming them would hijack the web fallback.

**Native config.** `MARKETING_VERSION 1.4` / `CURRENT_PROJECT_VERSION 5`. Release `CODE_SIGN_IDENTITY` was `"Apple Development"` — a development identity on the configuration the archive uses — now `"Apple Distribution"`. Added `ITSAppUsesNonExemptEncryption=false` (kills the export-compliance prompt on every upload) and `NSCameraUsageDescription` (no camera plugin, but a web `<input capture>` would hard-crash without it). The launch screen used `systemBackgroundColor`, which resolves white in light mode behind a `#0f0f0f` app; now a fixed dark ground, and the stale declared image size (1366² for a 2732² asset) is corrected.

[...]

PR #2569, opened 2026-09-08
feat(ios): route universal links, and close the iOS billing gaps

2witstudios:
Follow-up to #2568. That PR deferred claiming `/invite/*` because nothing routed it; this adds the routing and restores the claim **together**, plus the two remaining code-doable items from the submission epic.

## Universal links now actually work

`DeepLinkHandler` handles two cases that are **not** interchangeable, and shipping one without the other leaves half the links dead:

| Start | Delivery | Why a listener alone isn't enough |
|---|---|---|
| **Cold** | `App.getLaunchUrl()` | The URL is spent before any listener can exist — the app just loads `/dashboard`. |
| **Warm** | `appUrlOpen` | The native side only *notifies plugins*; it never calls `loadUrl`. With no listener the WebView silently stays where it was. |

Navigation goes through the router, never `window.location`: in the iOS shell a top-level location change reaches Capacitor's `WKNavigationDelegate`, which cancels anything outside `server.url`'s `/dashboard` prefix and opens Safari, blanking the WebView. `/invite/*` is outside that prefix, so this distinction is load-bearing.

`resolveDeepLink` is an **allowlist**, not "navigate to whatever path arrived" — the URL is external input and a universal link can name any path on the host. Two deliberate choices:

- It ignores `pagespace://auth-exchange` outright. `/api/auth/desktop/exchange` redeems its code with no PKCE binding, so whichever app receives the code can take a session. Routing it here would extend that surface.
- Anything unrecognised **on the claimed host** goes to the browser rather than being swallowed, so a half-configured AASA degrades to today's behaviour instead of killing the link.

## Guideline 3.1.1 — two real gaps

- **`/settings/billing`** rendered a bare "Billing — Your subscription, payment methods, and invoices" header over nothing on iOS. Every *section* was gated; the *route* was not. It also fetched `/api/stripe/invoices` and `/api/stripe/upcoming-invoice` on mount regardless of platform. Now wrapped in the same `BillingGuard` that `/settings/plan` already uses — which also prevents the child mounting, so those fetches never fire.
- **`/settings/usage`** stays reachable (usage is not a purchase, and its nav item is deliberately not `mobileHidden`), but its two credit banners were ungated purchase copy — *"Credit purchase canceled. You can buy credits anytime."* — now behind `showBilling`.

## Facebook SDK — documented, not removed

It's a hard dependency of `@capgo/capacitor-social-login`'s own target, so removing it means forking the plugin. Checked each link in the chain instead: no `FacebookAppID`/`FacebookClientToken` in `Info.plist`, `FacebookProvider.init()` only builds an `ISO8601DateFormatter`, its `initialize()` is empty, and we only ever call `SocialLogin.initialize` with `apple` or `google`. Linked but never initialized, so `NSPrivacyTracking = false` is accurate. Deliberately did **not** add `FacebookAutoLogAppEventsEnabled` — with no app ID there is nothing to disable, and dead config rots.

## Verification

20 pu agents were live and this worktree has no `node_modules`, so no vitest run locally — CI is the gate for the test files. But `deep-links.ts` imports nothing, so I executed **the real resolver** under `node --experimental-strip-types` against all 13 cases:

```
PASS  https://pagespace.ai/invite/abc123        -> {"kind":"route","path":"/invite/abc123"}
PASS  https://pagespace.ai/invite/a/b           -> {"kind":"external",...}
PASS  https://evil.example/invite/abc123        -> null
PASS  https://pagespace.ai.evil.example/...     -> null
PASS  http://pagespace.ai/invite/abc123         -> null
PASS  pagespace://auth-exchange?code=stolen     -> null
all 13 cases pass
```

Then mutation-checked the guards, which **found a gap in my own tests**: swapping exact-match for `endsWith` did not flip any assertion. `pagespace.ai.evil.example` fails `endsWith` too — the host that slips through is a *prefix* lookalike, `evilpagespace.ai`. The code already rejected it; the test did not guard it. Added that case. Dropping the `https:` check and loosening the invite regex both flip correctly.

## Still needs a device — not in this PR

**The AASA claim must be device-verified before archiving 1.4.** Unit-level evidence is not device evidence: after merge and a marketing deploy, reinstall the app (Apple's CDN caches the AASA) and confirm an `/invite/` link opens it from **both** a cold start and with the app already running. The two paths are handled separately and only one is exercised by any given tap.

Same for the billing sweep — the gate reads `window.Capacitor`, so it only means anything on a real build.

Epic: `iOS App Store Submission Epic` (`pcmu8yj2cvfk60y9d4r10ovo`).

https://claude.ai/code/session_014B1XUKffgMpNUtvFR1xLeH

## Summary by CodeRabbit

- **New Features**
  - Added support for opening `/invite/*` links directly in the iOS and web apps from cold or warm starts.
  - Unrecognized claimed links now open in the browser, while unsupported or unrelated links are ignored.
  - Billing settings and usage purchase alerts are now shown only when billing access is available.

- **Documentation**
  - Updated Android and iOS setup, review, and verification guidance for universal links.
  - Documented the Facebook iOS SDK configuration.

- **Tests**
  - Added coverage for deep-link routing, validation, and lifecycle behavior.

### DIFF
--- a/apps/android/README.md
+++ b/apps/android/README.md
@@ -232,10 +232,11 @@ succeed — which is prerequisite 1.
 
 ### The filter to add, once both prerequisites are met
 
-iOS is deferred for the same reason and claims no paths either — see
+iOS now routes these links — `apps/web/src/components/DeepLinkHandler.tsx` consumes
+`appUrlOpen` and `getLaunchUrl`, so **Prerequisite 2 above is met for both platforms**;
+what remains for Android is the signing cert and `assetlinks.json`. Paths mirror
 `apps/marketing/public/.well-known/apple-app-site-association` (the copy Caddy actually
-serves) and the "Universal links" section of `apps/ios/README.md`. Whenever the two
-platforms do start capturing links, they should capture the same ones. Widening beyond these paths should still wait on prerequisite 2 — whole-host
+serves), currently `/invite/*` only, so the two platforms capture the same links. Widening beyond these paths should still wait on prerequisite 2 — whole-host
 capture would strand users on `/dashboard` from every marketing, blog, or docs link.
 
 A third prerequisite applies to the auth-callback paths specifically: `/api/auth/desktop/exchange`
--- a/apps/android/README.md
+++ b/apps/android/README.md
@@ -232,11 +232,14 @@ succeed — which is prerequisite 1.
 
 ### The filter to add, once both prerequisites are met
 
-iOS now routes these links — `apps/web/src/components/DeepLinkHandler.tsx` consumes
-`appUrlOpen` and `getLaunchUrl`, so **Prerequisite 2 above is met for both platforms**;
-what remains for Android is the signing cert and `assetlinks.json`. Paths mirror
-`apps/marketing/public/.well-known/apple-app-site-association` (the copy Caddy actually
-serves), currently `/invite/*` only, so the two platforms capture the same links. Widening beyond these paths should still wait on prerequisite 2 — whole-host
+**Prerequisite 2 is now met** — `apps/web/src/components/DeepLinkHandler.tsx` consumes
+`appUrlOpen` and `getLaunchUrl`, and it lives in the shared web layer, so the routing is
+already there for Android the moment it starts receiving links. What remains for Android is
+prerequisite 1: the release signing certificate and a real `assetlinks.json`. Until those
+exist **Android captures nothing** — no https intent-filter is registered.
+
+When it does, mirror `apps/marketing/public/.well-known/apple-app-site-association` (the copy
+Caddy actually serves), currently `/invite/*` only, so both platforms capture the same links. Widening beyond these paths should still wait on prerequisite 2 — whole-host
 capture would strand users on `/dashboard` from every marketing, blog, or docs link.
 
 A third prerequisite applies to the auth-callback paths specifically: `/api/auth/desktop/exchange`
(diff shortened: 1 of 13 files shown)


2witstudios:
Pushed `bcc538c2d`. All seven threads addressed — the P1 was a genuine hole and the rest were real.

**P1 — mount outside the dashboard layout.** Correct, and it broke exactly the case the feature exists for: a signed-out `/dashboard` is rewritten by middleware to `/auth/signin`, so `DashboardLayoutClient` never mounts and `getLaunchUrl()` is never read. An invite tap is very often signed out. Moved to the root layout beside `ClientTrackingProvider`, which is the existing precedent for an app-wide null client component.

**Major — register the listener before awaiting the launch URL.** Fixed; ordering is now listener → launch URL. Added the duplicate-order policy you asked for: a link that surfaces through *both* paths is acted on once, so it does not push a duplicate history entry. Both are covered by tests (one asserts the actual call order).

**Major/P2 — malformed escape aborts listener setup.** Correct and the consequence was worse than the throw: it ran inside `setup()`, so the broad catch swallowed it and the session was left with no warm-start listener at all. Now degrades to handing the URL to the browser. Verified against the real resolver — `/invite/%` and `/invite/%zz` both return `external` instead of throwing.

**P2 — purchase alerts render before platform detection.** Both of you flagged this at the call site, but the root cause is the hook: `showBilling` defaulted to `true` until `isReady`, which is backwards for a flag whose whole job is gating purchase surfaces. There are **six** callers and every one gates purchase UI, so I fixed it at the door rather than repeating `isReady && showBilling` six times. `hideBilling` deliberately keeps its old default — it drives redirects, and bouncing before detection would throw web users off their own billing pages. Added `useBillingVisibility.test.ts`, which had no test at all.

**Android README overclaim.** Right, thank you — I wrote that the two platforms capture the same links. Android captures nothing today; no https intent-filter is registered until the signing cert and a real `assetlinks.json` exist. Reworded to say what is actually true: the shared routing prerequisite is met, prerequisite 1 is not.

PR #2570, opened 2026-09-08
fix(sandbox): resolve a probe on its own sentinel — the runtime holds the exec socket open 5–10s after exit

2witstudios:
The Ports pane's first real use in production failed the way it was built to fail — loudly — with `probe-failed: timed-out`, four times running. This is the diagnosis, measured, and the fix.

## What was measured

Through the **shipped driver** (`createSpritesSandboxClient` → `runCommand`), instrumented per phase, against a fresh sprite and the long-lived env sprite the user was on:

| | fresh | env (running) |
|---|---|---|
| socket open | +151 ms | +194 ms |
| first stdout | +161 ms | +223 ms |
| **`exit`** | **+5 348 ms** | **+10 483 ms** |

— for `true`, `ss -ltn` and `ss -ltnp` alike. The command finishes in ~200 ms; the Sprites runtime then holds the exec WebSocket open for 5–10 s before closing it, and the SDK emits `exit` only on close (`websocket.js` `handleClose`). So every run through this driver carries a 5–10 s tail the command did not spend, a 5 s cap can never succeed, and `-p` had nothing to do with it. `probeRelayRuntime` (10 s) shares the path and would have failed on the env sprite's first relay create.

## Why I had not seen it

Every probe test ran against a fake `exec` or the SDK's `execFile` from a laptop — never the shipped driver. And 5 s was set by feel: it was the shortest exec cap in the codebase, whose only other 5 s is fenced to a just-created sprite, while every user-gesture exec elsewhere uses 10–30 s.

## The fix: an opt-in stdout sentinel

`RunCommandArgs.stdoutSentinel`. When set, `runSpawned` resolves as soon as a stdout **line** begins with the token, reads the exit code from that line, strips it from the result, and SIGKILLs the lingering socket itself. **Inert unless set** — a long build or a shell keeps the close-based contract untouched. `findSentinel` requires a complete line, so a chunk boundary inside the code (`12` of `127`) cannot fake a result.

Both probes now end their command with `echo "<sentinel> $?"` and pass the token. Re-measured through the driver against the env sprite: **120 ms, 206 ms, 270 ms** for the full `probeListeningPorts`, returning `:3000` with its pid. The cap is raised to 20 s as a backstop (matching `sandbox-storage-measure`); with the sentinel it is not expected to fire.

## Verification

- Mutation-checked: disabling the sentinel branch fails the new driver tests (a hanging command with a sentinel must resolve; the same command without one must still time out).
- sandbox-client 114/114 · lib preview + client suites 587/587 · realtime 1266/1266 · `tsc` clean on lib, web, realtime.
- **After deploy:** Scan in the same `next dev` session should list `:3000` in well under a second. I'll be watching the row and realtime.

## Worth knowing beyond this PR

Every `runCommand` caller in the codebase pays this 5–10 s tail today (file browser, storage measure, agent tools). This PR deliberately does not change their contract; adopting the sentinel where a caller controls its own command is a cheap follow-up.

https://claude.ai/code/session_01P8A8KLYCoViP7ouH5QzbnB

### DIFF
--- a/packages/lib/src/services/sandbox/preview/__tests__/dev-preview-effects.test.ts
+++ b/packages/lib/src/services/sandbox/preview/__tests__/dev-preview-effects.test.ts
@@ -232,7 +232,10 @@ describe('probeRelayRuntime — socat is preferred only when the sprite proves i
     let seen: { cmd: string; args?: string[] } | undefined;
     const runtime = await probeRelayRuntime(async (args) => { seen = args; return { exitCode: 0, stdout: '/usr/bin/socat\n', stderr: '' }; });
     assert({ given: 'exit 0', should: 'socat', actual: runtime, expected: 'socat' });
-    assert({ given: 'the probe', should: 'ask sh for socat', actual: seen, expected: { cmd: 'sh', args: ['-c', 'command -v socat'], timeoutMs: 10_000 } });
+    // Ends with a sentinel and asks the driver to resolve on it: the runtime
+    // holds the exec socket open 5–10s after exit, so waiting for `exit`
+    // would spend the whole budget on a close (see `RunCommandArgs.stdoutSentinel`).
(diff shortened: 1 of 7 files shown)


2witstudios:
@coderabbitai review

PR #2587, opened 2026-09-09
feat(marketing): real App Store screenshots, as a narrative, iPad in landscape

2witstudios:
Ten real simulator captures off the shipping 1.4 build, plus the harness changes they needed. Before this, `public/screenshots/ios/` held only a README — `bun run capture` would have failed its own missing-capture check.

## The five tell a story

Ask → build → result → triggers → schedule, from one session: *"Build out a drive for my coffee roasting company."*

| # | slug | what it shows |
|---|---|---|
| 1 | `ask` | the request, and the agent asking clarifying questions back |
| 2 | `builds` | **Ran 7 commands (Create Page ×7)** — the agent actually working |
| 3 | `result` | 22 pages across six folders, full tree in the sidebar |
| 4 | `triggers` | task agent triggers: *run when due date arrives / when task is completed* |
| 5 | `workflows` | Create Workflow with a cron schedule |

The order is load-bearing: only the first two or three appear in App Store search results. Frames 2 and 4 are the strongest **Guideline 4.2** evidence available — nobody looks at an agent creating pages and calls this a website in a wrapper.

## iPad is landscape

`2752×2064`, not portrait. PageSpace is a three-pane product and portrait squeezes it into something closer to a large phone. App Store Connect accepts either orientation for 13" iPad provided the dimensions match the one you pick. iPhone stays portrait, as phone screenshots universally are.

That split means `StoreScreenshot` now carries two layouts: stacked for portrait, side-by-side for landscape — the latter mirroring the landing hero, so the store and the site read as one product.

## All dark, on purpose

Both light and dark captures were taken. Dark wins because these get composited over `hero-space.webp` beneath a dark scrim — light captures fight that backdrop — and because dark matches the app icon, the splash (`#0f0f0f`) and the landing page.

## Verification

Every file confirmed at its exact required size before filing:

```
iphone/  ask builds result triggers workflows   1320x2868
ipad/    ask builds result triggers workflows   2752x2064
```

**Geometry was tuned against rendered composites, not guessed.** I rendered the same geometry offline and looked at it, which caught two things a spec review would not have: a fifth of each canvas sat empty below the device, and on iPad the device landed 36px from the right edge — close enough to read as a mistake rather than a deliberate bleed. Both corrected.

**What I could not verify:** the real React render. This worktree has no `node_modules` and ~20 pu agents were live, so Next + Playwright was off the table under the no-local-builds rule. The offline preview shares the geometry but not the fonts or the `Tag` styling, so **run `bun run --cwd apps/marketing capture` and eyeball the ten outputs before uploading anything to App Store Connect.** The spec asserts exact dimensions, so a mis-sized export fails loudly rather than at upload.

Epic: `iOS App Store Submission Epic` (`pcmu8yj2cvfk60y9d4r10ovo`).

https://claude.ai/code/session_014B1XUKffgMpNUtvFR1xLeH

## Summary by CodeRabbit

- **New Features**
  - Added refreshed App Store screenshot layouts for iPhone and iPad, including landscape iPad compositions.
  - Updated device presentation with official hardware frames, glass effects, ambient glow, and reflections.
  - Introduced a six-screenshot story covering builds, team collaboration, agents, tasks, publishing, and permissions.
  - Tailored screenshot content and routes to each supported device.

- **Bug Fixes**
  - Prevented the development indicator from appearing in captured screenshots.

- **Documentation**
  - Updated iOS screenshot guidance with revised dimensions, ordering, capture instructions, and validation details.


2witstudios:
Good catch, and it found the root cause of something I had been working around by hand — every render in this branch was preceded by `pkill -f "next dev"`, which was me papering over exactly this.

Fixed, though not quite as proposed. `reuseExistingServer: false` on its own makes capture **fail** whenever a dev server happens to be running on 3004, and the error even advises setting it back to `true`. So I took the other suggestion in your comment and gave capture its own port: it boots on **3005**, never collides with the dev server, and always carries `CAPTURE=1`. A clean export becomes a property of the config rather than something you have to remember to do first.

The specs also drop their second hardcoded copy of the port and use Playwright's `baseURL`, so there is one place to change it.

Verified the way the bug would actually show up rather than by inspection: started a plain `bun run dev` on 3004, left it running, ran the full capture, then cropped an export at 1:1 in the corner where the badge used to sit. Clean, 12/12 passing.

PR #2590, opened 2026-09-09
feat(env-bridge): Local Environments ready for cloud GA — server policy, the exec click, audit/Stop/settings, posture docs, and the exit gate

2witstudios:
Local Environments — an agent runs on the user's own computer — from "works on my machine" to a feature that can be released. Four verified waves on `pu/local-env-ga`: **#2582** (server layer), **#2583** (approvals + the exec click), **#2585** (audit, activity, Stop, settings), **#2584** (security posture docs), **#2588** (the exit gate, run). Behind `LOCAL_ENVS_ENABLED`, which is **off everywhere**.

## The one-sentence proofs

- **PageSpace's say is real.** With `exec` off in `serverPolicy` the server refuses to sign and **no frame reaches the daemon** — proven on hardware, `GATE P19 … grantId NULL`.
- **A machine is driven by its owner only** ([D-6]). `bindPolicy` is the single value `owner` in the type, the CHECK and the gate; a drive admin is 403 on policy, Stop, activity and approvals, in a real browser.
- **A chat click can only unblock a request the machine itself froze.** The daemon byte-compares the re-issued request against the frozen one; anything else is `approval_mismatch`.
- **The injection case executes nothing.** A page written by another member, carrying `curl … | sh`, read by the owner's agent: the model quoted the payload back and declined; when made to comply, the machine froze it and the card showed the literal command. `grep -c '"verdict":"allow'` on the daemon's audit returns **0**.
- **Stop reaches the process.** Not just the server row: a signed `pause` frame kills the running process group (`exit 137`, pid gone) and the machine signs the acknowledgement.

## What each wave did

| Wave | What changed |
|---|---|
| 1 · #2582 | `decideSign` refuses `flag_disabled → revoked → paused → server_denied` **before the signing key is touched**; `serverPolicy` written at mint from an explicit dialog choice; owner-only PATCH as a compare-and-set; `no_server_ops` at bind; `admins`/`members` binding removed structurally |
| 2 · #2583 | Approvals re-keyed `(env, user, op, subject)` and made durable, so a new chat does not re-prompt and approving `git status` never authorises `rm`; `exec` is never scaffolded into the machine allowlist, so it reaches the `ask` verdict by construction; the owner's click in chat, verified by the machine |
| 3 · #2585 | Server-side audit rows joined to the daemon's log by grant id (**fail-closed**: no row, no grant); live activity to the owner's room only; Stop; drive and account settings; the approval mirror and a reconnect replay that **blocks** the env until the machine acknowledges |
| Docs · #2584 | `docs/security/local-environment-bridge.md` + a customer page, every claim citing the file that makes it true |
| Gate · #2588 | The 26 negatives re-scoped row by row, 14 added, **40/40 PASS, 0 SKIP** on a production build |

## What this does NOT claim

Stated plainly here because the posture doc states it plainly: **`exec` is not sandboxed.** A command you approve runs as you, with your credentials and your network. Roots confine what an agent can *name*, not what a process *does* ([D-1], accepted). Unattended `exec` without a sandbox is **materially less safe than Codex and roughly Claude Code minus the watching human**; the mitigation at general availability is the owner's click on every exec class, and OS confinement is a post-general-availability milestone. Approval is remembered **per program, not per command line**. An owner who hand-edits `exec` into their machine allowlist turns the click off — honoured as their choice ([D-7]), and the daemon now says so loudly at connect. The residual register in the posture doc carries all twelve risks with owners.

## Review found what CI could not

Fourteen defects across the waves, one continuous-integration failure among them. Three broke a wave's own guarantee: Stop changed server state but never reached the running process; an unacknowledged Stop was never retried; a failed reconnect replay released the block so grants could sign while the machine still held a revoked approval. The gate then found a **destructive-on-refusal** bug — `DELETE …/envs/[envId]` revoked the machine *before* the `live_sessions` guard, so a delete the product **refused** had already deleted the machine key and revoked 22 sessions. Fixed here: the revoke is now the delete transaction's hook, under the row lock, after the guard. The three sibling paths were audited and do not have it (register **R-12**).

The gate's own first run set two database fields by hand and did not put them in the seed script. Codex caught it; the run was re-done from a clean seed with a new control row proving the admin *can* start an ordinary session, so the refusal under test is the environment gate and not non-membership. Without that fix the injection check would have passed **for the wrong reason** — the agent could not read the page at all.

## Migrations

`0291` (bind policy) · `0292` (grant audit) · `0293` (paused) · `0294` (approval mirror) · `0295` (daemon epoch). Each generated by `db:generate`, a no-op on the second run, `drizzle-kit check` clean.

## Wire protocol

The signed `hello` now **requires** `daemonEpoch`, and `pause` / `pause_result` / `approval_revoke_result` are new frames. A CLI built before this branch cannot connect. Acceptable now only because the flag is off everywhere and the CLI ships from this tree.

## Before turning the flag on

The posture doc's flag-on checklist — every item a command or a task id, not a sentence. The gate report is on task `ukqmwy41zkf192hwgh6sqsxf`; the epic is `j945few5ssv75k5ad0bowbb4`.


2witstudios:
**On hold pending two hardening fixes — do not merge yet.**

A threat review of the merged branch (prompted by the question "what would have to go wrong for someone's computer to be exposed?") found two gaps that the epic's own design intent does not survive. Both are being fixed on this branch before it merges. Nothing is exposed meanwhile: `LOCAL_ENVS_ENABLED` is off on every deployment.

**Gap 1 — a file write is a command in disguise.** Confinement answers *where*, never *what*. There is no deny-list on filename or content, `mode` is chosen by the request and applied to existing files, and the decision layer never even sees `mode` (it is dropped in `grant-args.ts`). So a prompt injection on a shared page — no hack, no key theft, no server compromise — can have the owner's own agent write `.git/hooks/pre-commit` mode 0755 inside the enrolled root, and the next ordinary `git commit` runs it as the owner. That bypasses the Tier B click entirely. Fix: a sensitive write escalates to the same owner click `exec` uses, rather than being refused, so ordinary writes stay headless.

**Gap 2 — the owner's click is server-attested, not machine-verified.** The daemon checks the server's signature and that the request byte-matches what it froze. It has no way to observe that a human clicked; the owner-only check lives entirely in the web route. An attacker holding the signing key and a position as the server the daemon dials can therefore drive the ask → approve loop alone. This works only in chat-ask mode, which is the headless path built for GA. Fix: the click carries a passkey assertion bound to the frozen request, verified by the daemon against credentials pinned at enrolment. PageSpace already has the WebAuthn stack; the missing piece is that assertions were verified server-side and discarded.

Board: hardening A `k4mtz7rn4ewh1amx71ki9y60`, hardening B `nsic3ddzk6sekmacy4mf19tr` on epic `j945few5ssv75k5ad0bowbb4`. This PR will be re-pointed at the updated tip once both land and the affected gate rows are re-run from a clean seed.

PR #2612, opened 2026-09-12
feat(oauth): Phase 0 contracts — ADR 0004, profile scope, step-up boundary, redirect rules

2witstudios:
Phase 0 of the "Sign in with PageSpace" epic (`yv08hib74nrtmksdzxmf5nkw`, phase page `kb5oh1ngdlwn1aw5rd48ml6g`): the pure contracts every later phase stands on. Pure functions in `packages/lib` plus one ADR — no routes, no schema, no UI, no SDK, no CLI.

> **Rewritten 2026-09-12.** The previous description had been patched across a dozen edits and had drifted — it said "10 → 22 schemes" where the count is 35, "17-row table" where it is 22, and "five self-review commits" above a list of six. Every number below is recounted from the source, not restated from an earlier version of this text. The drift itself is recorded in the leaf-task evidence rather than quietly erased.

## The five leaves

- [x] **1 · ADR 0004** — `docs/adr/0004-third-party-oauth-clients.md`. All **14** architecture decisions with `file:line` ground truth, a **22-row** fail-closed posture table (G1–G22), the pure-function signatures, D-8/D-9/D-10/D-11 as decided, and **6 Phase 1 obligations** with acceptance lines.
- [x] **2 · `profile` scope** — `scopes.ts`, `consent.ts`, `metadata.ts`, `grant-scope-summary.ts`. RED first: 20 failed / 4 passed.
- [x] **3 · `requiresStepUp`** — `step-up-boundary.ts`. **100%** stmts/branch/funcs/lines; 7/7 mutants killed.
- [x] **4 · Redirect URI rules** — `clients.ts`. **100%** on all four metrics; 10/10 mutants killed; the 16 pre-existing CLI redirect tests pass unchanged.
- [x] **5 · `validateClientRegistration`** — `client-registration.ts`. **100%** on all four metrics.

**20 commits.** `src/auth/oauth`: **327 → 478 tests**, all green.

## What changes behaviour

- **`profile`** releases exactly what `/api/auth/me` already returns and nothing else. `profile account` → new `profile_account_conflict`; `profile manage_keys`/`all_drives` reuse the existing conflict shapes; `name:` alongside `profile` → `name_without_mint_grant`. `profile drive:<id>:member offline_access` parses; `offline_access` alone still rejected. Canonical order keeps the top-level flags alphabetical. `profile ⊄` drive sets, `⊄ account`, `⊄ all_drives`.
- **`isPureDriveGrant` excludes profile-bearing sets.** Without it, `profile drive:abc123` falls into the `applyKeyGrant` branch (`oauth-repository.ts:250`) and silently mints a long-lived `mcp_` key where the user consented to a sign-in.
- **The loopback port wildcard is `firstParty`-only.** A third party registering `http://127.0.0.1/callback` gets nothing — not the wildcard, not its own exact port.
- **Registration and authorize cannot diverge**: `validateClientRegistration` calls the same `validateRedirectUri` with `firstParty: false`, so "registration accepts exactly what authorize honours" is true by construction. `allowedScopes` holds scope SHAPES; a concrete `drive:abc123:member` is rejected as firmly as `account`, and `[]` is a mistake rather than "no cap".

## Defects found and fixed after the first review

Two reviews (Codex + human) raised 9 threads; two independent `/code-review` passes and my own probes raised 11 more. All 20 are addressed, and the second pass independently verified the first pass's fixes. The ones that mattered:

**Never-throw contract violated by a plain JSON body.** `String(scope)` invoked caller-controlled coercion, and `{"toString":"no"}` shadows it with a non-callable — `TypeError`, i.e. a 500 from any registration body. Now a fixed `[non-string]` placeholder. The poisoned case went into the **existing** "never throws" suite, which had walked seven hostile inputs and missed it: every one was hostile *data*, none was hostile *behaviour*.

**`requiresStepUp` enumerated the table it was guarded by.** The `satisfies` clause closes the *table*; a hand-written body can still omit an entry, and the new scope then compiles, carries an entry, and never steps up — fail-open. Body is now derived. Verified by simulating that exact regression (`filter((f) => f !== 'account')` → 4 red, including a new 512-combination sweep compared against a reference predicate written from the ADR rather than from the implementation).

**Three consent-screen lies, which turned out to be one class.** `name:` on a profile grant narrated "create a key named ci" over an exchange that mints nothing; the `profile` line made a *set-level* claim from a per-scope formatter, so `/api/account/oauth-grants` rendered "No access to any drive or content" directly above "Act as you in Acme Drive"; and `profile` is advertised before the enforcement ADR 0004 promises for it exists. What they share: **three surfaces each re-derive what a scope means.** That is now a Phase 1 rule in the ADR — a surface that narrates scopes calls `describeGrantScopes` rather than re-deriving it.

**A deny-list documented as an allow-list.** `filesystem:`, `view-source:`, `jar:`, `chrome:`, `content:`, `intent:`, `mailto:` and friends classified as `private_use` and were registrable; the URL-wrapping ones hand the authorization code to whatever the inner URL addresses. Now **35 entries** in four documented groups (33 reachable through `classifyRedirect`; `http:`/`https:` are answered for first and kept so the set is independently correct). The docstring states the fail direction honestly and says why private-use *cannot* be allow-listed — RFC 8252 §7.1 makes the scheme one the app itself picks. Squatting is recorded in the ADR as accepted-not-solved, with the reverse-DNS option and why it isn't taken now (it would refuse `swipesend://callback`, US5's own shape).

**`https://localhost./callback` walked past the `localhost` refusal** — `hostname` is `'localhost.'`, so the equality check never matched. Fixed; then probing *that* fix found it stripped one dot while the parser preserves repeats, so `localhost../callback` still passed. Now strips all of them, as a loop rather than `/\.+$/` (CodeQL polynomial-backtracking shape).

Plus: name/description hygiene against bidi and zero-width characters beside the "Unverified app" badge, `allowedScopes` capped, redirect dedupe on normalized `href`, two unreachable branches removed, and five zero-trust behaviours that were correct but unasserted (prototype pollution, homograph hosts, trailing-dot FQDN, host case, percent-encoded paths) pinned by test.

**Two escalated, not fixed here** — both in `apps/web`, which this phase may not touch, both now blocking acceptance lines on the Phase 1 board page and in ADR 0004: neither live consent surface has a `profile` branch (a `profile` grant is approved against a **blank** capability list), and `profile` is reachable through the live device-authorization door today.

## Mutation evidence

Re-run against the tip so the line numbers are reproducible, not historical. Mutate by LINE INDEX (twin lines make a first-occurrence string replace hit the wrong entry), one at a time, source restored between each, control re-run at the end.

`step-up-boundary.ts` — **7/7 killed**:

| line | mutation | result |
|---|---|---|
| 43 | `drives: () => false` | 9 failed / 21 passed |
| 45 | `allDrives: () => false` | 3 failed / 27 passed |
| 47 | `account: () => false` | 4 failed / 26 passed |
| 49 | `manageKeys: () => false` | 4 failed / 26 passed |
| 51 | `updateKeyId: () => false` | 3 failed / 27 passed |
| 53 | `activateKeyId: () => false` | 3 failed / 27 passed |
| 59 | `profile: () => true` (opposite direction, as a control) | 8 failed / 22 passed |

`clients.ts` — **10/10 killed**:

| line | mutation | result |
|---|---|---|
| 213 | drop the `localhost` refusal | 3 failed / 54 passed |
| 218 | drop the loopback-literal check | 2 failed / 55 passed |
| 220 | scheme deny-list off | 2 failed / 55 passed |
| 257 | wildcard check off | 2 failed / 55 passed |
| 266 | candidate userinfo/query/fragment off | 1 failed / 56 passed |
| 274 | **loopback first-party gate off** | 2 failed / 55 passed |
| 284 | registered userinfo/query/fragment off | 1 failed / 56 passed |
| 288 | kind-must-match off | 1 failed / 56 passed |
| 294 | drop `pathname` from the loopback compare | 3 failed / 54 passed |
| 300 | `href` → `origin` | 8 failed / 49 passed |

The first sweep left three survivors — a scheme regex no input could reach, and two mutually-redundant wildcard checks. Rather than write tests around dead branches, the source was simplified until each remaining rule was load-bearing, then re-swept.

**Exhaustiveness guard proved, not asserted:** adding a probe field to `ScopeSet` produces `step-up-boundary.ts: error TS1360`, so a new scope field cannot be added without stating whether it may be approved without a second factor.

## Gate

- **CI green on the tip.** Run [34691690114](https://github.com/2witstudios/PageSpace/actions/runs/34691690114) on `67bf22144` — the current tip, byte-identical to `origin/pu/signin-phase0` — is **success** on all three jobs: Unit Tests, Lint & TypeScript Check, E2E. Every Unit Tests step passed, including `Run tests with coverage (Node 24)` (where `packages/lib`'s coverage thresholds gate), both integration suites, `Check evidence manifest`, and `Coverage report`.

  **It took two attempts, and the first failure is disclosed rather than buried:** one test of 12,908 — `distributed-rate-limit.integration.test.ts > boundary bursting` — failed on attempt one. It is timing-fragile by construction (blocking requires landing <1000 ms into the next bucket, so the boundary alignment plus two Postgres round-trips must fit inside a second), this diff touches no rate-limit or security file, and a re-run of only that job on the identical SHA with zero code change went green. Full reasoning in [this comment](https://github.com/2witstudios/PageSpace/pull/2612#issuecomment-5645840576).

  This base branch is not in `pull_request.branches` for `test.yml`/`security.yml`, so opening a PR here triggers nothing — CI was obtained via `workflow_dispatch`. Wiring `pu/signin-with-pagespace` in (`pull_request` only, not `push`, per `test.yml`'s own header) is an epic-level call outside this diff, and it affects every remaining phase PR.
- `bun run typecheck` (monorepo) — 20/20 tasks successful. `eslint` clean on every touched file.
- Coverage, per module: `step-up-boundary.ts`, `clients.ts`, `client-registration.ts` — all **100%** stmts/branch/funcs/lines. The two uncovered branches in `consent.ts`/`grant-scope-summary.ts` are pre-existing custom-role fallbacks this PR does not touch.
- `knip:check` — **green in CI on this branch** (`Lint & TypeScript Check`, step 9, `Knip (unused code gate)`). Locally it reports 4 findings in `env-bridge`/`drive-envs`, none in this diff; that is the known pu-worktree artifact, and the definitive control is a *different directory*, which CI is.

## ⚠️ One deliberate deviation from the phase spec — flagged, not absorbed

**Phase 0 requirement 7** says `describeScopeForConsent({kind:'profile'})` should say, in substance, *"See your name, email, and avatar. No access to any drive or content."* As of `e26224353` that second sentence is emitted only when the caller passes `profileIsSoleAccess`.

I changed a spec'd behaviour because **the spec conflicts with itself**: requirement 2, two lines above, mandates that `profile drive:<id>:member offline_access` parses — so the grammar this phase is required to build makes requirement 7's second sentence *false for a legal grant*. Verified live rather than argued: `/api/account/oauth-grants` rendered

```
See your name, email, and avatar. No access to any drive or content.
Act as you in Acme Drive — everything you can currently do there …
```

ADR 0002 Decision 5 opens by saying zero trust includes the user being able to trust what the consent screen says, and US10 requires consent to tell the truth, so I resolved it that way. The exact spec'd sentence is still produced for the case requirement 7 describes — a `profile`-only grant — through `describeGrantScopes`, the path that actually renders it.

**This is a spec question, not mine to settle.** If the literal reading is preferred, the revert is one ternary in `consent.ts` and three tests. Also flagged on the leaf task page and in the Point Guard channel.

## Notes for review

- The `apps/web` edits are exactly one mechanical addition — `profile: false` in seven `ScopeSet` test fixtures, required because the field is non-optional. No route, component or production file is touched.
- `packages/lib/package.json` gains `./auth/oauth/step-up-boundary` and `./auth/oauth/client-registration` exports entries, since Phase 1 imports both from `apps/web` and a missing subpath fails `web#build` while passing vitest.
- `getRegisteredClient` and `PAGESPACE_CLI_CLIENT_ID` keep their exact signatures; Phase 1 adds the DB-backed lookup behind the static one.
- `description: ""` means "no description" and is dropped from the accepted value. I originally rejected it for consistency with `name` and flagged it here for a second opinion; the second opinion was better than my reasoning — an absent description and an empty one render identically, so rejecting `''` bought nothing while leaving no way to *clear* one on a partial update and failing any create form that posts its empty optional fields. Whitespace-only is still an error: `''` is unambiguously empty, `'   '` is invisible content.

https://claude.ai/code/session_0142ugFtSbPe4xzkTdyzyCR9


packages/lib/src/auth/oauth/metadata.ts
```diff
@@ -19,8 +19,11 @@ const CODE_CHALLENGE_METHODS_SUPPORTED = ['S256'] as const;
 
 const TOKEN_ENDPOINT_AUTH_METHODS_SUPPORTED = ['none'] as const;
 
-/** ADR 0002 Decision 1 grammar. `drive:*` stands for the templated `drive:<driveId>[:role]` family. */
-const SCOPES_SUPPORTED = ['account', 'offline_access', 'manage_keys', 'drive:*'] as const;
+/**
+ * ADR 0002 Decision 1 grammar, as amended by ADR 0004 (`profile`). `drive:*`
+ * stands for the templated `drive:<driveId>[:role]` family.
+ */
+const SCOPES_SUPPORTED = ['account', 'profile', 'offline_access', 'manage_keys', 'drive:*'] as const;
```
2witstudios: **P1 anchor.** This line makes `profile` a publicly advertised scope in the RFC 8414 discovery document, while the enforcement ADR 0004 promises for it ("short-circuits the same way `manage_keys` does") does not exist in any layer yet — see the PR-level comment for the full trace and the empirical confirmation.

The phase spec does require this line, so it is not a defect here. Raising it so the sequencing is owned by a task: either land the `isProfileOnly` short-circuit alongside `manageKeysNoDriveAccess`, or hold `profile` out of `SCOPES_SUPPORTED` until Phase 1 does, so the advertised grammar and the enforced grammar never disagree.

packages/lib/src/auth/oauth/step-up-boundary.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * True iff approving this grant requires the consent step-up ceremony —
+ * i.e. iff it grants content access (`drive:*`, `all_drives`, `account`) or
+ * touches key material (`manage_keys`, `update_key:*`, `activate_key:*`).
+ *
+ * False only for identity-alone grants: `profile` and `profile
+ * offline_access`. Every field is listed below so that dropping one is a
+ * visible deletion rather than an invisible omission.
+ */
+export function requiresStepUp(scopes: ScopeSet): boolean {
```
2witstudios: **P2 — the compile-time guard covers the table, but not this consumer.**

The `satisfies { readonly [K in keyof ScopeSet]: … }` clause on `STEP_UP_CONTRIBUTION` is real and load-bearing, and the TS1360 probe in the commit body proves it. But this function then hand-enumerates nine calls. Add a `ScopeSet` field, add its table entry (compiler satisfied), forget the line here — it compiles, and the new scope silently never steps up. That is the fail-open direction, and it halves the guarantee the table was written to provide.

Worth noting the 7 mutants in the commit body all mutate *table entries*, so the existing verification is orthogonal to this gap rather than a refutation of it.

Deriving the body from the table closes it:

```ts
export function requiresStepUp(scopes: ScopeSet): boolean {
  return (Object.keys(STEP_UP_CONTRIBUTION) as (keyof ScopeSet)[]).some((k) =>
    (STEP_UP_CONTRIBUTION[k] as (v: ScopeSet[typeof k]) => boolean)(scopes[k]),
  );
}
```

Separately, the module docstring says "This module is the single place that distinction is computed." That is an aspiration until Phase 1 wires it: production still decides step-up via `isCredentialEscalatingGrant` (`verify/route.ts:82`, `decision/route.ts:98`). Right now there *are* two expressions of the rule — the thing the doc warns about. Worth saying "Phase 1 replaces `isCredentialEscalatingGrant` at both call sites" so the claim reads as a plan rather than a description.

packages/lib/src/auth/oauth/clients.ts
```diff
@@ -43,15 +79,76 @@ export function getRegisteredClient(clientId: string): RegisteredClient | null {
...
+
+/**
+ * Which rule, if any, this URI plays by. Deliberately total and
+ * allow-listed: anything that is not recognizably one of the three shapes is
+ * `reject`, so a scheme nobody thought about is refused rather than
+ * defaulting into the most permissive branch.
  */
-export function validateRedirectUri(client: Pick<RegisteredClient, 'redirectUris'>, redirectUri: string): boolean {
-  if (typeof redirectUri !== 'string' || redirectUri.length === 0) return false;
+function classifyRedirect(url: URL): RedirectKind {
```
2witstudios: **P2 — this is a deny-list, and the docstring above calls it an allow-list.**

The comment says *"Deliberately total and allow-listed: anything that is not recognizably one of the three shapes is `reject`, so a scheme nobody thought about is refused rather than defaulting into the most permissive branch."* The implementation does the opposite — any scheme not in `NEVER_PRIVATE_USE_SCHEMES` falls through to `private_use`. Verified against Node's URL parser:

```
filesystem:https://evil.com/temporary/x  -> private_use  (registrable, honoured)
view-source:https://evil.com             -> private_use
jar:https://evil.com!/x                  -> private_use
content://com.evil/x, chrome://settings, mailto:..., intent://...  -> private_use
```

Two things follow. The doc and the code disagree on the fail direction — fix one. And RFC 8252 §7.1 (which the test file cites) recommends private-use schemes be reverse-DNS names under a domain the app controls; nothing enforces or nudges that, so `slack://oauth` is registrable by anyone and scheme-squatting on a shared device is the classic hijack this rule exists to blunt. A `scheme.includes('.')` requirement, or an explicit ADR note that squatting is accepted for now, would close it.

(`intent://…#Intent;…;end` happens to be caught by the fragment check — accidentally, not by design.)

packages/lib/src/auth/oauth/client-registration.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/**
+ * Per-field schemas, checked independently so one bad field never masks
+ * another — a registration form should show every problem at once, and an API
+ * caller should not have to fix errors one round trip at a time. The
+ * `allowedScopes` entries are `string` here; which strings are declarable is
+ * `classifyScopeShape`'s job below, so the two rejections stay distinguishable.
+ */
+const NAME = z.string().min(1).max(100);
```
2witstudios: **P3 — `name` reaches the consent screen with weaker validation than an mcp key name.**

`z.string().min(1).max(100)` accepts `'   '`, `'\u0000'`, and RTL-override / zero-width sequences. `scopes.ts:13` already defines `NAME_CONTROL_CHAR_RE` for exactly this on key names, and applies a length + control-char check at `:231`.

This string renders on the consent screen right next to the "Unverified app" badge, which is precisely where a spoofed name pays off — a name with a bidi override or trailing whitespace can visually undercut the badge that is the whole trust signal. Reuse `NAME_CONTROL_CHAR_RE` and reject whitespace-only.

packages/lib/src/auth/oauth/client-registration.ts
```diff
@@ -0,0 +1,207 @@
...
+const DESCRIPTION = z.string().max(500);
+const HTTPS_URL = z.string().refine(isHttpsUrl);
+const REDIRECT_URIS = z.array(z.unknown()).min(1).max(10);
+const ALLOWED_SCOPES = z.array(z.unknown()).min(1);
```
2witstudios: **P3 — unbounded.** `redirectUris` is capped at 10 one line up; `allowedScopes` has no `.max()`. A registration can submit an arbitrarily long array, and every entry produces an error object in the returned list. Cap it at something like `.max(20)` — there are only six legal shapes, so any larger input is a mistake or an abuse.

packages/lib/src/auth/oauth/client-registration.ts
```diff
@@ -0,0 +1,207 @@
...
+    redirectUris.forEach((uri: unknown, index: number) => {
+      const field = `redirectUris[${index}]`;
+      // The authorize-time rules, applied as the least-privileged client
+      // there is: nothing registered here can claim the first-party loopback
+      // port wildcard.
+      if (typeof uri !== 'string' || !validateRedirectUri({ redirectUris: [uri], firstParty: false }, uri)) {
+        errors.push({ code: 'invalid_redirect_uri', field });
+        return;
+      }
+      if (seen.has(uri)) {
```
2witstudios: **P3 — duplicate detection compares raw strings; matching compares normalized `href`.**

`['https://a.com/cb', 'https://a.com:443/cb']` passes this check as two distinct entries, but `validateRedirectUri` treats them as the same URI (it compares `candidate.href === pattern.href`, and the URL parser drops the default port). Same for case-normalized hosts. Dedupe on `new URL(uri).href` so the stored list matches the set the authorize endpoint will actually honour.

2witstudios (commented):
## 🔬 Review — Phase 0 contracts

Genuinely good work. Every new rule is fail-closed by construction, `isPureDriveGrant` correctly excludes profile-bearing sets, the `satisfies` table in `step-up-boundary.ts` is a real compile-time guard, and the commit bodies carry pasted mutation evidence (10/10 redirect mutants, 7/7 step-up mutants, plus a TS1360 probe proving the table is closed). All nine "Given X, should Y" requirements on the phase page are met. 445/445 tests green in `src/auth/oauth`; `knip --workspace packages/lib` reports no new findings; exports-map entries are present for both new subpaths.

One finding is serious and is **not** a defect in this PR's stated scope — it is a gap between what ADR 0004 promises and what any code implements.

### 🔴 P1 — `profile` is advertised and reachable, but the "short-circuits like `manage_keys`" enforcement does not exist anywhere

ADR 0004 line 126 and epic decision 4 both say `profile` "short-circuits the same way `manage_keys` does … no drive, page, task or channel read or write is reachable with it." No code implements that. Meanwhile this PR puts `profile` into `SCOPES_SUPPORTED` (live discovery document) and into the accepting path of `parseScopeList` → `validateAuthorizeRequest` → exchange.

Trace a `profile`-only grant:

1. `parseScopeList('profile')` → ok
2. `validateAuthorizeRequest` → ok (not a pure-drive grant, so no `name:` requirement)
3. exchange → `applyKeyGrant` returns `not_a_key_grant` → ordinary `ps_at_` pair
4. `validateOAuthAccessToken` → `account: false`, `allowedDriveIds: []`
5. `getAllowedDriveIds` (`apps/web/src/lib/auth/index.ts:756`) → `isManageKeysOnly` false → `[]`
6. `checkMCPDriveScope` (`:785`) → `allowedDriveIds.length === 0` → **`return null`, i.e. allowed, for any drive**

This is precisely the shape `validateOAuthAccessToken:349-360` refuses to issue for `all_drives`, in a comment that names the hazard: *"grant-everything via `getAllowedDriveIds`/`checkMCPDriveScope`."* `manage_keys` got the `manageKeysNoDriveAccess` sentinel (`index.ts:750`) for exactly this reason. `profile` got neither guard. 93 route files call `checkMCPDriveScope`.

Confirmed empirically with a throwaway test (since removed):

```
✓ getAllowedDriveIds(profileOnly) === []   ·  isManageKeysOnly(profileOnly) === false
✓ checkMCPDriveScope(profileOnly, 'drive_someone_elses') === null   // ALLOWED
```

The consent copy this PR adds says *"See your name, email, and avatar. No access to any drive or content."* That is currently false — US10's "consent tells the truth", inverted.

**Not a production exposure:** this PR targets `pu/signin-with-pagespace`, and the epic converges to master as one reviewed PR after Phase 6's security review. And the Phase 0 spec itself requires `SCOPES_SUPPORTED` to include `profile` (line 14) while assigning enforcement to Phase 1 — so the gap is designed into the plan, not missed here.

**The ask** is therefore a blocking Phase 1 acceptance line rather than a change to this diff:

> `checkMCPDriveScope` and `getAllowedDriveIds` deny a profile-only OAuth principal — mutation-checked.

plus an `isProfileOnly` sentinel beside `manageKeysNoDriveAccess`. If Phase 1 slips or reshuffles, this is the item that must not become the thing everyone assumed someone else did.

Related: `profile-scope.test.ts` opens with "It carries ZERO content access", and the only test near that claim (`:175`) asserts the consent *string*. A mutation check on `isManageKeysOnly` would survive. The 7 test-fixture edits in this PR all mechanically add `profile: false`; no call site was asked what `profile: true` means.

### OWASP Top 10 (2021) pass

| | |
|---|---|
| A01 Broken Access Control | ⚠️ P1 above — the only finding in this category |
| A02 Cryptographic Failures | ✅ no crypto touched; https-only logo/homepage closes the mixed-content downgrade |
| A03 Injection | ✅ no SQL/HTML construction; `isHttpsUrl` blocks `javascript:`/`data:` on consent-rendered links and images |
| A04 Insecure Design | ⚠️ P1 (advertise-before-enforce) and the `requiresStepUp` comment below |
| A05 Security Misconfiguration | ✅ exports map complete, knip clean, no new config surface |
| A06 Vulnerable Components | ✅ zero dependency changes — `package.json` gains only two exports entries |
| A07 Auth Failures | ⚠️ net materially stronger (loopback wildcard first-party-gated, kind-matching, `profile_account_conflict`, `isPureDriveGrant` exclusion). Residual: private-use scheme squatting, commented inline |
| A08 Integrity Failures | ✅ best-handled item here — `validateClientRegistration` rebuilds field-by-field so `firstParty`/`verified` cannot ride in, and there is a test asserting it |
| A09 Logging Failures | ➖ nothing to log from a pure function. Phase 1's registration route should emit a security-audit event on rejection — repeated `forbidden_scope` is an abuse signal |
| A10 SSRF | ➖ not reachable today (`logoUrl` is browser-rendered, never server-fetched). **Flag for Phase 1:** `isHttpsUrl` checks protocol and userinfo only, so `https://169.254.169.254/` and `https://localhost/logo.png` both pass. The moment anything proxies, caches or screenshots that logo it is an SSRF — `packages/lib/src/security/url-validator.ts` already has `isBlockedIP`/`validateExternalURL`/`safeFetch` |

### Tests & commits

Tests are colocated, isolated and explicit; `emptySet()` is a factory invoked per test rather than a shared mutable fixture. Edge coverage is thorough (decimal/octal/short-form IPv4 loopback, IPv6 alternate spellings, path traversal, percent-encoded traversal). Commit messages are among the best in this branch's history — conventional format, and each body names the rejected alternative and what collapsing it would have cost.

Inline comments below are all P2/P3.

2witstudios:
### Re: the Codex review — I reproduced all three. Two are real; one I'd push back on.

Verified by probe against the branch (probe file removed afterwards), not taken at face value.

**✅ `client-registration.ts:174` — confirmed, and it breaks the module's stated contract.** This is the strongest finding on the PR and I missed it. The docstring promises *"Total: every input path returns a result, none throws"*, and `validateClientRegistration` throws on a plain JSON body:

```js
validateClientRegistration(JSON.parse(
  '{"name":"x","redirectUris":["https://a.example.com/cb"],"allowedScopes":[{"toString":"no"}]}'
))
// TypeError: Cannot convert object to primitive value
```

`String(scope)` at the `typeof scope !== 'string'` branch is the trigger: the parsed object shadows `toString` with a non-callable, `Object.prototype.valueOf` returns the object itself, and `ToPrimitive` throws. Reachable from any untrusted registration body, so Phase 1's route turns it into a 500 unless it catches. Worth noting the existing `never throws on untrusted input` suite (`:59-77`) walks a list of hostile inputs and still misses this one — the fix should add a poisoned-`toString` case to that list, not just to a new test.

Cheapest fix is to not coerce at all: `scope: typeof scope === 'string' ? scope : '[non-string]'`, or drop the `scope` field on that branch since `field` already locates the entry.

**✅ `scopes.ts:479` — confirmed, and it is a consequence of this PR.**

```
parseScopeList('profile drive:abc123 name:ci')  ->  ok
hasNewKeyName(scopes)   -> true    // consent narrates "a key named ci"
isPureDriveGrant(scopes) -> false  // ...and nothing is ever minted
```

`validateAuthorizeRequest`'s mint-name guard only fires for `isPureDriveGrant || isAllDrivesGrant`, so this shape sails through, and exchange falls to `not_a_key_grant`. The user approves a screen that promises a named key and gets an OAuth pair instead. It fails in the safe direction — less is minted than promised — but the consent screen is still wrong, which is the same class as the P1 in my review above: **the system does something other than what the consent screen said.** Two instances of that class in one PR is the signal worth acting on.

Fix belongs in `parseScopeList`'s rule 13 (`name_without_mint_grant`): `name:*` requires a mint-shaped grant, and a profile-bearing set is no longer mint-shaped, so `profile` should join `updateKeyId`/`activateKeyId` in that exclusion.

**❌ `clients.ts:112` — reproduced, but I don't think it's a bypass.** Confirmed a third party can register and use `https://127.0.0.1/callback`:

```
validateRedirectUri({firstParty:false, redirectUris:['https://127.0.0.1/callback']}, 'https://127.0.0.1/callback')      -> true
validateRedirectUri({firstParty:false, redirectUris:['https://127.0.0.1/callback']}, 'https://127.0.0.1:8443/callback') -> false
```

The port wildcard — the thing the first-party gate exists to protect — is **not** granted; that second line is the whole point. This module defines `loopback` as *cleartext* loopback, and the commit message states the rule as "`http://` only for loopback", so an https numeric-IP registration is ordinary exact-match https by design, not an escaped loopback. The `localhost`-vs-`127.0.0.1` asymmetry is coherent for the same reason RFC 8252 §8.3 gives: `localhost` can be remapped, numeric literals cannot.

What's fair in the finding is that **no test pins this either way** — the one https-loopback test (`redirect-rules.test.ts:155`) uses `firstParty: true`, so third-party https loopback is unasserted behaviour. Worth a test that locks in "accepted, but no wildcard" and one line in ADR 0004 Decision 3 saying https on a numeric loopback host is deliberately ordinary https. P3, not P2.

---

Revised tally for this PR: **1 × P1** (`profile` advertised without its enforcement — epic-level, Phase 1 acceptance line), **3 × P2** (never-throw contract violated; `name:` on a profile grant mints nothing; `requiresStepUp` body not derived from its guarded table), **1 × P2-doc** (scheme classification is a deny-list documented as an allow-list), **4 × P3**.

packages/lib/src/auth/oauth/scopes.ts
```diff
@@ -423,7 +468,19 @@ export function isScopeSubset(requested: ScopeSet, granted: ScopeSet): boolean {
...
-  return !scopes.account && !scopes.manageKeys && scopes.drives.size > 0 && scopes.updateKeyId === null && scopes.activateKeyId === null;
+  // `profile` disqualifies the shape: an `mcp_tokens` row carries drive rows
+  // and nothing else, so it structurally cannot deliver the identity half of
+  // a `profile drive:*` grant. Such a grant resolves to an ordinary OAuth
+  // access/refresh pair instead (fail closed — without this, a `profile`
+  // request would silently mint a long-lived `mcp_` key at exchange).
+  return (
+    !scopes.account &&
+    !scopes.manageKeys &&
+    !scopes.profile &&
```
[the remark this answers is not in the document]
2witstudios: Confirmed and fixed in b9105ad8d — this was real, and it is the second instance in this PR of the class "the system does something other than what the consent screen said", which is what made it worth acting on rather than deferring.

Rule 13's mint-shape exclusion now includes `profile`, so `profile` joins `update_key`/`activate_key` there:

```ts
if (
  newKeyName !== null &&
  !((allDrives || drives.size > 0) && !profile && updateKeyId === null && activateKeyId === null)
) {
  return { ok: false, error: { code: 'name_without_mint_grant' } };
}
```

It rejects the promise, not the grant — `profile drive:abc123` still parses, only the `name:` token is refused.

Test: **`rejects \`name:\` on a profile-bearing drive grant`** in `profile-scope.test.ts`, covering both `profile drive:abc123 name:ci` and the `offline_access` variant, plus the positive control that the same set without `name:` is still accepted.

Recorded in ADR 0004 as grammar rule 14c and as G18 in the fail-closed table.

*(Edited to repair a mangled test name in the original reply — a shell quoting slip, no change in substance.)*

packages/lib/src/auth/oauth/client-registration.ts
```diff
@@ -0,0 +1,207 @@
...
+    // An explicit cap of nothing is a form mistake, and rejecting it is the
+    // fail-closed reading: treating `[]` as "no cap" would silently turn the
+    // safest-looking input into the widest one.
+    errors.push({ code: 'invalid_allowed_scopes', field: 'allowedScopes' });
+  } else if (allowedScopes !== undefined) {
+    const seen = new Set<string>();
+    allowedScopes.forEach((scope: unknown, index: number) => {
+      const field = `allowedScopes[${index}]`;
+      if (typeof scope !== 'string') {
+        errors.push({ code: 'unknown_scope', field, scope: String(scope) });
```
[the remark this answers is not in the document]
2witstudios: Confirmed and fixed in b9105ad8d. This was the strongest finding on the PR — it violated the module's own stated contract ("Total: every input path returns a result, none throws"), and it was reachable from any registration body.

Fixed by not coercing at all, as suggested:

```ts
const NON_STRING_PLACEHOLDER = '[non-string]';
...
if (typeof scope !== 'string') {
  errors.push({ code: 'unknown_scope', field, scope: NON_STRING_PLACEHOLDER });
  return;
}
```

A fixed placeholder rather than a serializer: `field` already locates the entry, so echoing the value buys nothing and re-opens the same question for whatever the next hostile shape is.

Two things beyond the literal fix:

1. The poisoned case is added to the **existing** `never throws on untrusted input` list (`:59-77`) — that suite walked seven hostile inputs and missed this one, so adding it only to a new test would have left the same blind spot. It now carries a poisoned `toString`, a poisoned `valueOf`, and the same shape in `redirectUris`.
2. A second test pins that every non-string entry (`7`, `null`, `true`, `[]`, `{}`) reports the identical placeholder, so the error shape leaks nothing about the input.

While fixing this I removed an unrelated unreachable `try/catch` I had just added for redirect normalization — same class of reasoning, opposite direction: rather than guard a throw that cannot happen, the invariant that makes it impossible (`validateRedirectUri` only returns true for a string it parsed) is now asserted directly in a test.

packages/lib/src/auth/oauth/clients.ts
```diff
@@ -43,15 +79,76 @@ export function getRegisteredClient(clientId: string): RegisteredClient | null {
...
+  if (url.protocol === 'https:') {
+    // RFC 8252 §8.3: `localhost` can be remapped; only the numeric literals
+    // are trustworthy, and those live on the loopback branch.
+    return url.hostname === 'localhost' ? 'reject' : 'https';
```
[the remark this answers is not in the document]
2witstudios: Reproduced, and I agree with the pushback in the PR-level comment rather than with the finding as stated — but the part of it that was fair has been acted on.

The port wildcard is what the first-party gate exists to protect, and an https numeric-IP registration never gets it:

```
validateRedirectUri({firstParty:false, redirectUris:['https://127.0.0.1/callback']}, 'https://127.0.0.1/callback')      -> true
validateRedirectUri({firstParty:false, redirectUris:['https://127.0.0.1/callback']}, 'https://127.0.0.1:8443/callback') -> false
```

`loopback` in this module means **cleartext** loopback; an https numeric-IP URI is ordinary exact-match https, and being https it carries none of the cleartext risk the rule is about.

What was fair is that nothing pinned it. Fixed in b9105ad8d:

- New test `https on a numeric loopback host is ordinary exact-match https for a third party — accepted, but with NO port wildcard`, asserting accept-exact / reject-wildcard / reject-http for `127.0.0.1`, and the same for `[::1]`.
- The `classifyRedirect` https branch now carries the rationale inline.
- ADR 0004 Decision 3 states it: "`https://` on a numeric loopback host is ordinary https, not loopback."

So the behaviour is unchanged and now asserted in both directions instead of being incidental.

packages/lib/src/auth/oauth/metadata.ts
2witstudios: Agreed on all of it, including that it is not a defect in this diff — the phase page requires this line (Phase 0 requirement 14) while assigning enforcement to Phase 1, so the gap is designed in.

I have not removed `profile` from `SCOPES_SUPPORTED`, because doing so would fail the phase's own acceptance and would move the sequencing decision from a place where it is written down to a place where it is implicit. Instead the sequencing is now **owned in writing**, in b9105ad8d: ADR 0004 Decision 4 gains a **"Phase 1 obligations this decision creates"** section whose first item is this, with your trace and your acceptance line quoted verbatim:

> `checkMCPDriveScope` and `getAllowedDriveIds` deny a profile-only OAuth principal, via an `isProfileOnly` sentinel beside `manageKeysNoDriveAccess` — mutation-checked.

It names the empty-`allowedDriveIds`-reads-as-unscoped path at `index.ts:785`, the `validateOAuthAccessToken:349-360` comment that already names this exact hazard for `all_drives`, the `manageKeysNoDriveAccess` precedent at `:750`, and the 93 route files that call `checkMCPDriveScope`. It closes by saying that until it lands, the consent copy this phase adds is a promise about Phase 1, not a description of today.

The other three items you raised are in the same section: the two live step-up expressions, `isHttpsUrl` as a future SSRF surface the moment anything fetches a logo (with the pointer to `url-validator.ts`), and security-audit events on registration rejection.

On the related point about the test file over-claiming — you are right, and I have fixed it rather than argued it. `profile-scope.test.ts`'s header said `profile` "carries ZERO content access"; it now states that these tests cover the **grammar** only, that the resolution half is Phase 1's `isProfileOnly` sentinel, and — explicitly — that "nothing below should be read as evidence that content access is enforced today; it is not."

Leaving this thread open: it is the item that must not be quietly closed, and it is the orchestrator's to carry onto the Phase 1 board.

packages/lib/src/auth/oauth/step-up-boundary.ts
2witstudios: You are right, and the distinction between the two halves is the part I had not seen: the `satisfies` clause closes the **table**, the body was still hand-enumerated, and the gap between them fails **open**. Fixed in b9105ad8d with your suggested form:

```ts
export function requiresStepUp(scopes: ScopeSet): boolean {
  return (Object.keys(STEP_UP_CONTRIBUTION) as Array<keyof ScopeSet>).some((field) =>
    (STEP_UP_CONTRIBUTION[field] as (value: ScopeSet[typeof field]) => boolean)(scopes[field]),
  );
}
```

Your observation that the 7 existing mutants all mutate table entries — and are therefore orthogonal to this gap rather than a refutation of it — is exactly right, so I added verification aimed at *this* gap specifically:

- A **512-combination sweep** over every on/off state of the nine fields, compared against a reference predicate written from ADR 0004 Decision 6 rather than from the implementation, reporting the disagreeing combinations by name.
- A second test that guards the sweep itself (a typo'd field name would silently shrink it).
- Confirmed load-bearing by simulating the exact regression you describe — dropping one field from the derived key list:

```
filter((f) => f !== 'account')  ->  4 failed | 26 passed
```

The seven original table mutants were re-run at the new line numbers and all still die.

On the docstring: also right, and also fixed. It now says the module is *intended* to become the single place, names both live expressions (`isCredentialEscalatingGrant` at `verify/route.ts` and `decision/route.ts`, plus the unconditional step-up on `POST /api/oauth/authorize`), and states that **Phase 1 replaces both** — so it reads as a plan. Same item is recorded as obligation 2 in ADR 0004's new Phase 1 section.

packages/lib/src/auth/oauth/clients.ts
2witstudios: Correct on both counts, and the `filesystem:` / `view-source:` / `jar:` set is the part that actually stung — those wrap a second URL, so accepting one hands the authorization code to whatever that inner URL addresses. Fixed in b9105ad8d.

The deny-list goes from 10 schemes to 22, grouped by why each is there: script/content-bearing (`javascript:`, `vbscript:`, `data:`, `blob:`, `about:`, `file:`), URL-wrapping (`filesystem:`, `view-source:`, `jar:`), platform-handled (`chrome:`, `chrome-extension:`, `resource:`, `content:`, `intent:`, `android-app:`, `mailto:`, `tel:`, `sms:`). Two tests: every scheme you listed is rejected, and a second asserting `view-source:https://app.example.com/...` is refused *even when the wrapped URL is itself the registered redirect* — that one would have passed before only by accident.

On the doc/code disagreement, I fixed the **doc**, not the fail direction, and said why in the docstring: private-use schemes cannot be allow-listed, because RFC 8252 §7.1 makes the scheme one the app itself chose — an allow-list would have to enumerate every app that will ever exist. So `classifyRedirect` now states plainly that http(s) are allow-listed, private-use is **deny-listed**, and that the asymmetry is the one place the module does not fail closed.

On squatting: I did not add `scheme.includes('.')`, and I want to be explicit that this is a judgement call rather than an oversight — it would refuse `swipesend://callback`, which is the reference app's own shape and named in US5 and in the phase spec. ADR 0004 Decision 3 now records squatting as **accepted, not solved**, names the reverse-DNS option and the reason it is not taken now, names what stands in front of it today (exact-match registration, per-client scope caps, the "Unverified app" badge), and hands Phase 1 the decision — because the console UI is what would have to explain the constraint to a developer.

Your parenthetical is noted in the same breath: `intent://…#Intent;…;end` was caught by the fragment check by accident, and `intent:` is now on the deny-list by design.

packages/lib/src/auth/oauth/client-registration.ts
2witstudios: Fixed in b9105ad8d, and reusing `NAME_CONTROL_CHAR_RE` rather than re-spelling it, as you suggested — it is now exported from `scopes.ts` with a comment saying why the client name is held to the mcp key name's rule.

```ts
const NAME = z.string().min(1).max(100)
  .refine((value) => value.trim().length > 0)
  .refine((value) => !NAME_CONTROL_CHAR_RE.test(value))
  .refine((value) => !NAME_INVISIBLE_CHAR_RE.test(value));
```

I added the third check because your reasoning pointed past its own example: the threat you describe is *visual* subversion of the badge, and bidi overrides and zero-width characters are not control characters by the `\x00-\x1F\x7F` definition yet do exactly that. `NAME_INVISIBLE_CHAR_RE` covers `U+200B–200F`, `U+202A–202E`, `U+2066–2069` and the BOM.

Three tests: whitespace-only (including `\t`, `\n`, NBSP), control + bidi + zero-width, and a positive control — `Swïpe­Send`, `归档`, `Café App`, `App — One` all still pass, so this reads as a control-character rule rather than an accidental ASCII rule.

packages/lib/src/auth/oauth/client-registration.ts
2witstudios: Fixed in b9105ad8d — `.max(20)`, with your reasoning as the comment:

```ts
// There are only six legal shapes, so anything longer is a mistake or an
// attempt to make the error list itself the payload.
const ALLOWED_SCOPES = z.array(z.unknown()).min(1).max(20);
```

Test asserts both boundaries (20 accepted, 21 rejected with `invalid_allowed_scopes`), matching how the `redirectUris` cap is already tested.

Worth naming the mechanism you spotted: because this validator deliberately reports *every* problem at once, an unbounded input array is an unbounded output array — the error list is the amplification. That is a property of the design choice, so the cap belongs on every array the design collects errors over, which is now both of them.

packages/lib/src/auth/oauth/client-registration.ts
2witstudios: Fixed in b9105ad8d — dedupe now compares `new URL(uri).href`, which is exactly the string `validateRedirectUri` matches on:

```ts
const normalized = new URL(uri).href;
if (seen.has(normalized)) { errors.push({ code: 'duplicate_redirect_uri', field }); return; }
seen.add(normalized);
```

Test covers the default-port case you gave plus the two other normalizations the parser performs: host case (`https://APP.example.com/cb`) and scheme case (`HTTPS://app.example.com/cb`).

One thing this surfaced and I want to flag rather than bury: I first wrapped the parse in a `try/catch`, which left an unreachable branch and dropped the module off 100% coverage. Guarding a throw that cannot happen is not safety, it is an untested branch. So the parse is now unguarded, with the invariant that licenses it asserted directly:

> `invariant: every uri validateRedirectUri accepts is parseable, so registration can normalize it unguarded`

If that contract ever changes, a test says so rather than a 500.

2witstudios:
## Review round 1 — all 9 threads addressed (`b9105ad8d`, `bfcc97fba`)

Both reviews landed together, so I worked them as one set. Every thread has a concrete reply; none are auto-resolved — the ones fixed in this round are left open for your verification.

**Acted on as stated (6):** the never-throw violation, `name:` on a profile grant, the derived `requiresStepUp` body, the deny-list/allow-list doc contradiction, the `name` hygiene rule, the `allowedScopes` cap, and the normalized-`href` dedupe.

**Pushed back on 1, and acted on the part that was fair:** Codex's `clients.ts:112` https-loopback finding. Your reading is the one I hold — the port wildcard is what the gate protects, and exact-match https never grants it — but "no test pins this either way" was correct, so it is now pinned in both directions and stated in ADR 0004 Decision 3.

**Owned rather than closed (1):** the P1. I did not pull `profile` from `SCOPES_SUPPORTED` — that would fail the phase's own requirement 14 and would move the sequencing from something written down to something implicit. Instead ADR 0004 Decision 4 now carries a **"Phase 1 obligations this decision creates"** section with your acceptance line quoted, the `index.ts:785` trace, the `manageKeysNoDriveAccess` precedent, and the 93-call-site count. Plus the other three sequencing items you raised (two live step-up expressions, `isHttpsUrl` as a future SSRF surface, audit events on registration rejection).

### Three things I'd flag back

**1. Your P1 and the Codex `name:` finding are the same failure mode, and the second one told me where to look.** Both are "the consent screen says X, the system does Y." That is what moved the `name:` fix from *nice* to *required* — a grant that fails safe is still a grant that lied. Worth naming as a review heuristic for Phases 1–6: for every scope shape, ask what the screen promises and what exchange actually does.

**2. The never-throw bug survived a test suite written specifically to catch it.** `never throws on untrusted input` walked seven hostile inputs and missed the one that mattered, because every entry was hostile *data* and none was hostile *behaviour* — a value that misbehaves when touched. The poisoned `toString`/`valueOf` cases went into that existing list rather than a new test, so the list now covers both kinds.

**3. I removed something I had added this round.** The dedupe fix initially wrapped `new URL(uri)` in a try/catch, which left a branch no input can reach and dropped the module off 100% coverage. Guarding an impossible throw is not safety, it is an untested branch — so the parse is unguarded and the invariant that licenses it (`validateRedirectUri` only returns true for a string it parsed) is asserted directly. Mentioning it because it is the same instinct that produced the original `String(scope)` bug, pointed the other way.

### Verification

- `src/auth/oauth`: **462 tests green** (was 445).
- 100% stmts/branch/funcs/lines on all three new modules, re-measured after every change in this round.
- Both mutation sweeps re-run at the shifted line numbers: **17/17 killed**, control green.
- The derived-body fix verified by simulating the regression it prevents — `filter((f) => f !== 'account')` → 4 red, including the new 512-combination sweep.
- Monorepo `bun run typecheck`: 20/20 tasks successful. `eslint` clean. `knip` unchanged (the 4 findings are in `env-bridge`/`drive-envs`, which master carries at this branch's base SHA).
- **CI**: this base branch isn't in `pull_request.branches` for `test.yml`, so the PR itself triggers nothing. I dispatched `test.yml` against `pu/signin-phase0` directly to get real signal — Unit Tests, Lint & TypeScript Check, and E2E. Wiring the base branch in is an epic-level call I've left alone; if you want it, `pull_request` only, not `push`, per that file's own header.

packages/lib/src/auth/oauth/step-up-boundary.ts
2witstudios: **Correction to my reply above.** I said the docstring was fixed in `b9105ad8d`. It was not — the edit silently matched nothing (my search string said "two independent expressions of *that* rule" where the file says "*the* rule"), so the module went on asserting it *is* the single place step-up is decided while I told you it no longer did.

Actually applied in `731114ce1`, and verified in the file this time rather than in the commit message:

```
$ grep -c "intended to become the single place" step-up-boundary.ts
1
```

The substance is what I described: it now says the module is *intended to become* the single place, names both live expressions (`isCredentialEscalatingGrant` at `verify/route.ts` and `decision/route.ts`, plus the unconditional step-up on `POST /api/oauth/authorize`), and states that Phase 1 replaces both.

The code half of this thread — the derived body and the 512-combination sweep — did land in `b9105ad8d` and is unaffected; I re-ran the suite and the `filter((f) => f !== 'account')` regression check after this commit.

Flagging the mechanism, not just the slip: three of my round-1 edits were string replacements that no-op silently on a near-miss. I have since checked every one of them against the file, and the other two had landed. Subsequent edits assert the match before writing.

2witstudios:
## Independent review pass — 5 findings, 3 fixed here, 2 escalated (`e26224353`)

I ran an independent review over this PR's own diff rather than waiting for the next one. It found five things; I reproduced all five by probe before acting on any of them. Three are mine to fix and are fixed. Two are in `apps/web`, which this phase may not touch — they are escalated below and added to the Phase 1 board page.

### Fixed here

**1. The `profile` consent line made a set-level claim from a per-scope formatter.** This one is mine and it is the sharpest thing on the list. `describeScopeForConsent({kind:'profile'})` returned *"See your name, email, and avatar. No access to any drive or content."* unconditionally — true for a profile-only grant, false for `profile drive:abc123`, which the grammar deliberately permits. `/api/account/oauth-grants` is already wired to `describeGrantScopes`, so it rendered:

```
See your name, email, and avatar. No access to any drive or content.
Act as you in Acme Drive — everything you can currently do there …
```

Verified by probe, not inferred. The claim now requires `ctx.profileIsSoleAccess`, which only a caller holding the set can affirm; absent, it falls back to the identity sentence alone. **Opt-in, not opt-out** — a caller who forgets the flag under-claims rather than lies, and on this surface the lie is the expensive direction.

**2. `https://localhost./callback` walked past the `localhost` refusal.** `new URL('https://localhost./cb').hostname === 'localhost.'`, so the equality check never matched and a third party could register and use it — the exact remappable name RFC 8252 §8.3 is cited for refusing, reached by a spelling the check didn't know. Classification now strips one FQDN dot; matching still compares raw `href`, so `https://app.example.com./cb` stays distinct from a registered `https://app.example.com/cb`.

**3. "Same surface, same rule" for the description was half true** — it had the character check but none of `NAME`'s trim refinements, so `"   "` and `""` validated.

### Escalated — real, and not this phase's to fix

**4. Neither live consent surface has a `profile` branch, so a profile grant is approved against a blank capability list.** `apps/web/src/app/oauth/consent/page.tsx:114-157` and `api/oauth/device_authorization/verify/route.ts:66-156` each hand-roll their own if-chain over `ScopeSet` fields instead of calling `describeGrantScopes`. I updated `grant-scope-summary.ts` (connected apps); those two were not in this phase's file boundary and were not updated. A `scope=profile` request therefore renders the app name above an empty list, and Allow still mints the code — **the user approves a grant the screen never named.**

That is the same class as your P1 and as the `name:` finding, now at its third instance, and it is worth saying what the three have in common: *three different surfaces each re-derive what a scope means, and `profile` had to be taught to all three separately.* The structural fix is that the two routes call `describeGrantScopes` like the third one does, which is a Phase 1 change.

**5. `profile` is reachable through the live doors today** — which sharpens your P1 rather than repeating it. `POST /api/oauth/device_authorization` accepts `scope=profile` for `pagespace-cli` now; `isCredentialEscalatingGrant` is false for it, so `/activate` needs no step-up; and `profile offline_access` passes rule 10, so the resulting credential is refreshable and long-lived. Your assessment that this is not a production exposure still holds — the branch converges to master as one reviewed PR after Phase 6 — but the reachable-today path is worth having written down next to the acceptance line, because "Phase 1 will handle it" and "nothing can reach it until Phase 6" are different claims and only the second one is load-bearing.

Both are now blocking acceptance lines on the Phase 1 board page (`o1u0sho4xvkk6hf2owjn3gtg`) alongside the four from your review.

### Verification

475 tests green (5 RED first). 100% stmts/branch/funcs/lines on `clients.ts`, `client-registration.ts`, `step-up-boundary.ts`. Mutation sweeps re-run against the tip and re-cited in the PR body with current line numbers — 17/17 killed, control green. `eslint` and scoped `tsc` clean. CI: E2E and Lint & TypeScript Check (which runs `knip:check`) green; Unit Tests still running.

2witstudios:
## Second independent review pass — logic verified, three more fixed (`df02dc8fd`, `45d12e9d0`)

I ran a second review over the commits that fixed the first one, on the principle that the newest code is the least reviewed. It independently verified the fixes hold — `canonicalHostname` closes the `localhost.` bypass without widening anything reachable, and `profileIsSoleAccess` is complete given `parseScopeList`'s exclusivity rules — and then found three things I'd introduced.

**I orphaned a docstring, two commits after fixing an orphaned docstring.** Inserting `canonicalHostname`'s doc block put it between `classifyRedirect`'s docstring and `classifyRedirect`, so a 20-line explanation of the allow-list/deny-list asymmetry ended up documenting a trailing-dot stripper, and `classifyRedirect` had none. Same mistake, same cause — inserting above an anchor without checking what was above it.

**A comment contradicted the docstring five lines above it.** My note on `http:`/`https:` claimed the deny-list "names every scheme that is not a private-use scheme"; the docstring says it is best-effort and not exhaustive. Worse, it was addressed to "a future caller that reaches for it directly" — precisely the caller who would then read *absence* from the set as proof a scheme IS private-use, which a deny-list cannot promise. It now says what is true: membership proves not-private-use, absence proves nothing.

**`description: ""` — you were right to want this questioned, and I've changed it.** I flagged it in the description as a deliberate call wanting a second opinion. The second opinion is better than my reasoning was: an absent description and an empty one render identically, so rejecting `''` bought nothing, while leaving no way to *clear* a description on a partial update (omitting means "unchanged") and failing any create form that posts its empty optional fields. It now normalizes to absent and is dropped from the accepted value. `'   '` is still an error — `''` is unambiguously empty, whitespace is invisible content, and they deserve different answers.

### On your P1, which three reviewers have now independently found

This pass reached the same conclusion as your review and as the first pass, and it added the citation that I think settles how Phase 1 should treat it. The `all_drives` guard in `validateOAuthAccessToken` is at **`:339-353`** (I had cited `:349-360`, taken from the thread rather than the file), and its comment already contains the argument:

> `all_drives` … as a bearer OAuth access token it would carry `allowedDriveIds: []` on a non-`account` token, **a shape this codebase's two authorization-helper families disagree on** (deny-everything via `isScopedOAuthAuth`/`getScopedAccessLevel`, or grant-everything via `getAllowedDriveIds`/`checkMCPDriveScope`, depending on which a given route calls). Today exactly two upstream gates keep `all_drives` from ever reaching a persisted access token … **reject here too rather than trust both to remember it forever.**

A `profile`-only token produces the identical shape. So obligation 1 isn't a new judgement call for Phase 1 to relitigate — it's applying a rule this file already argued for, to the scope that arrived after it was written. The board entry now carries the quote and the two line numbers (`isProfileOnly` beside `manageKeysNoDriveAccess` at `:750`, the guard beside the `all_drives` one at `:351`).

### Also, unprompted

Auditing my own work found four things no reviewer raised: two negative assertions that passed on an empty array (proved by stubbing `describeGrantScopes` to return `[]` — the test stayed green), a deny-list count I'd stated as 29 and 36 when it is 35, a fail-closed table described as 17 rows when it is 22, and the table header "each line is a test" quietly covering G16, which is a *compile-time* guarantee no runtime test can express. All corrected; the wrong numbers are recorded as dated corrections on the task pages rather than silently overwritten, so what was claimed and when stays legible.

### State

20 commits · **478 tests** green · 100% stmts/branch/funcs/lines on `clients.ts`, `client-registration.ts`, `step-up-boundary.ts` · both mutation sweeps re-run at current line numbers, 17/17 killed · `eslint` and `tsc` clean. CI run [34689582005](https://github.com/2witstudios/PageSpace/actions/runs/34689582005) was fully green on all three jobs; a run on the current tip is in flight.

2witstudios:
## CI green on the tip — with one flaky failure worth disclosing

Run [34691690114](https://github.com/2witstudios/PageSpace/actions/runs/34691690114) on `67bf22144` (the current tip) is **success** on all three jobs: Unit Tests, Lint & TypeScript Check, E2E.

It did not pass first time, and I'd rather you hear that from me than find it in the run history.

**First attempt: 1 failure out of 12,908 tests** —

```
FAIL src/security/__tests__/distributed-rate-limit.integration.test.ts
  > boundary bursting is blocked by weighted sliding window
AssertionError: expected true to be false          (line 154)
```

**Why I concluded it was not mine, before re-running:**

- The diff touches no rate-limit or security file; zero commits under `packages/lib/src/security`.
- The same test passed on run `34689582005`, which carried the bulk of these changes.
- Nothing landed between those runs that could influence a Postgres sliding window (a trailing-dot loop in `clients.ts`, `description` handling, consent narration, tests, docs).

**And why the test is structurally fragile**, from its own arithmetic rather than from "integration tests are flaky": it fills a 2-second bucket, sleeps exactly one window, and expects the previous bucket's decayed weight to still block. Blocking requires `1 + 2w > 2`, i.e. `w > 0.5`, i.e. landing **less than ~1000 ms** into the next bucket. Everything before the sleep — the boundary alignment plus two `checkDistributedRateLimit` round-trips to Postgres — has to fit in that second. On a loaded runner it doesn't, the weight decays past 0.5, and the request is *correctly* allowed. The assertion then fails for a reason that has nothing to do with the code under test.

**Confirmed empirically rather than asserted:** re-ran only the failed job on the identical SHA with zero code change — Unit Tests went green. Same commit, different outcome, which is the definition.

**Flagging it as a repo-quality item, not something I'll fix here** (out of this phase's file boundary): that ~1s budget will keep costing other branches CI cycles. A fix would be to measure the actual elapsed offset into the new bucket and assert against the weight that implies, rather than assuming the pre-sleep work is instantaneous.

This is also the argument for having re-run CI against the real tip instead of relying on the earlier green run: that run was green on `b9105ad8d`, five commits back, and would have let me report something I had not verified.

PR #2613, opened 2026-09-12
feat(ai): add GLM-5.3 Flash to the admin Z.ai Coder Plan provider

2witstudios:
## What

Adds **GLM-5.3 Flash** (`glm-5.3-flash`) to the admin-only `glm` provider — the direct Z.ai Coder Plan connection (`api.z.ai/api/coding/paas/v4`).

## Why

The provider's `models` map in `ai-providers-config.ts` is what `provider-factory` validates a selection against: an unlisted model id is **substituted** with the metered default rather than forwarded, so the model was unreachable on the coder plan.

## Changes

- `apps/web/src/lib/ai/core/ai-providers-config.ts` — `'glm-5.3-flash': 'GLM-5.3 Flash'` in the `glm` provider model map.
- `packages/lib/src/monitoring/ai-monitoring.ts` — pricing `0.075` in / `0.25` out per 1M, and a `1,310,720`-token context window. Both mirror the existing OpenRouter `z-ai/glm-5.3-flash` entries, so usage rows don't fall through to $0 / a default context window.

Usage remains exempt from the shared credit pool (`METERING_EXEMPT_PROVIDERS` — flat-rate subscription), but is still costed for reporting, matching how the other `glm-*` coder-plan models are handled.

## Testing

- `@pagespace/lib` ai-monitoring suite: **124/124 pass**.
- ESLint could not run in this worktree (deps not installed: `@eslint/eslintrc` missing). The diff is three literal entries in existing `as const` object maps with no new imports, so CI lint is the gate here.

https://claude.ai/code/session_01VN2stfTofVNvW79PspUsQP

### DIFF
--- a/apps/web/src/lib/ai/core/ai-providers-config.ts
+++ b/apps/web/src/lib/ai/core/ai-providers-config.ts
@@ -582,6 +582,7 @@ export const AI_PROVIDERS = {
     // Models officially supported by that endpoint; bare `glm-*` ids (no vendor prefix).
     models: {
       'glm-5.3':     'GLM-5.3',
+      'glm-5.3-flash': 'GLM-5.3 Flash',
       'glm-5.2':     'GLM-5.2',
       'glm-5.1':     'GLM-5.1',
       'glm-5-turbo': 'GLM-5 Turbo',
--- a/packages/lib/src/ai/__tests__/context-window.test.ts
+++ b/packages/lib/src/ai/__tests__/context-window.test.ts
@@ -243,6 +243,44 @@ describe('buildModelContext', () => {
     summarizerModel: 'test-model',
   };
 
+  // Regression: the admin-only Z.ai `glm` provider sends BARE model ids, which
+  // getContextWindowSize could not resolve — every glm conversation budgeted against
+  // the 200k conservative default instead of the model's real window. A ~160k-token
+  // history on glm-5.3-flash (1,310,720) is 0.12 of the real window but 0.80 of the
+  // default, i.e. past the 0.75 trigger ratio: it compacted for no reason.
+  it('budgets a bare glm id against its real window, not the 200k default', () => {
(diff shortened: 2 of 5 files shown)


2witstudios:
Confirmed and fixed in ab3150f59.

`getContextWindowSize` gated the catalog lookup on `model.includes('/')` (ai-context-calculator.ts:172) and no heuristic block below covered GLM, so every bare `glm-*` id fell through to the 200k conservative default. Through `buildModelContext`'s `triggerRatio: 0.75` / `hardRatio: 0.95`, that meant compaction at ~150k and hard truncation at ~190k. Worth noting this **predates this PR** — it hit the whole coder-plan family (`glm-5.3` and `glm-5.2` are 1,000,000 in the catalog and were budgeted at 200k too); `glm-5.3-flash` just makes the gap the widest at 1,310,720.

**Fix:** a direct lookup for `provider === 'glm'` / bare `glm-*` ids, rather than relaxing the `/` gate. That gate is deliberate — the catalog also holds bare aliases for other vendors whose rounded values would shift long-standing heuristic results (its comment says so explicitly), so accepting all bare catalog entries would be a much wider behaviour change. Ids absent from the catalog keep the 200k default.

**Tests:** catalog-window assertions for `glm-5.3-flash` / `glm-5.3` / `glm-4.5-air` / `glm-5.1`, plus two controls — an uncatalogued bare glm id still defaults to 200k, and the prefixed `z-ai/` twin is undisturbed. Also an end-to-end `buildModelContext` regression: a ~160k-token history no longer triggers compaction on `glm-5.3-flash`, while the same history still does on an uncatalogued id (so the assertion is about the window lookup, not about the history being small).

Mutation-checked: reverting only the source fix turns 4 calculator tests and the compaction regression red; the two control assertions correctly stay green. `@pagespace/lib` src/ai + src/monitoring: 616/616 pass, `typecheck` clean.

PR #2647, opened 2026-09-16
fix(billing): missed-grant reconcile review threads from #2645 (MON-2, WAL-5)

2witstudios:
Fixes the four live Codex review threads on merged PR #2645 (`packages/lib/src/billing/missed-grant-reconcile.ts`). Spec MON-2, WAL-5.

Commits: `58c495a19` adds the failing tests (each one run and seen failing), then `f0f696f33` has the fix. `719aa63c1` adds a test for the page-level lookup failure path.

## Thread → test

| Thread | Defect | Test (red before the fix → green after) |
|---|---|---|
| (A) P1 Claim only rows that are still missed grants | The claim UPDATE matched on `id` only, so a second run that overlapped the first re-matched the converted row and granted the allowance twice | `missed-grant-reconcile-concurrency.integration.test.ts` (**real Postgres**): a blocker holds `FOR UPDATE` on the row, both runs queue on the claim UPDATE (checked in `pg_stat_activity`), then they race. Before the fix: `expected 2 to be 1` |
| (B) P1 Advance past unresolved rows | `LIMIT 200` had no pagination, so 200+ unresolved rows took the whole batch on every run | `credits-flow…` › `WAL-5 bounded sweep: 250 unresolved rows ahead of a repairable one…`: the repairable row is granted within 3 runs (in practice, 1) |
| (C) P2 Derive the tier from subscription rows | Read the `users.subscriptionTier` cache | `MON-2 missed grant: the tier is derived from the subscriptions rows, NOT the stale … cache` (cache says free, subscription says pro → granted), plus the reverse case (cache says pro, subscription canceled → not granted) and the indeterminate case |
| (D) P2 Surface per-row failures | Errors were swallowed at debug level, the route returned 200 and audited success | lib: `MON-2 per-row failure: … COUNTED as failed` (transaction throws for 1 of 2 rows → `{reconciled:1, failed:1}`, logged at error with the ledgerId). Also `MON-2 page failure: a failed subscription lookup counts every row on that page as failed`. Route: `per-row missed-grant failures make the run non-2xx … NOT audited` → 500 and no success audit; also `a clean run reports zero failures with a 200` |

## Follow-up from Codex review on this PR
`f4aba0674` fixes the P2 'Preserve the legacy amount fallback'. The subscriptions table stores no amount, so an unmapped legacy price id could never reach `getTierFromPrice`'s legacy amount fallback and would stay indeterminate forever. The reconcile now passes the missed invoice's `paidCents` as the amount (exact match against `LEGACY_PRICE_AMOUNTS`). A non-matching amount still fails closed, and the cache is still never read. Test: 'a legacy subscription on an unmapped price id resolves through the invoice amount…' failed before the change; a mutant passing `null` fails it. **Policy note for review:** the invoice's paid amount (which includes tax and discounts) stands in for the price's list amount, so only an exact legacy-amount match resolves.

## [independent-review] on 332cbe6df → `fbd56cc10`
- **P2, the amount fallback classified subscriptions it didn't describe:** the invoice amount is now used only when the user has exactly one subscription row (any status). With more rows, an unmapped entitled row stays indeterminate. Test: a canceled mapped Pro row plus an unmapped active row at paidCents 1500 → `indeterminate: 1`. It failed before the fix.
- **P2, indeterminate rows raised no alert:** the route logs at `loggers.system.error` (the same sink as `failed`) with the ledger ids (`indeterminateLedgerIds`, capped at 50). The run still returns 200. The route tests check both that the alert fires and that a clean run doesn't alert.
- Mutants: always passing paidCents, never passing it, not recording ids, never alerting, always alerting. Each one fails a test.

## Independent review follow-up (`7c3e8fac7`)
Codex's review quota ran out, so a separate reviewer agent audited the full diff. It found no double-grant, keyset or deadlock bugs. Changes made from its findings:
- **Indeterminate rows are surfaced.** An entitled subscription that the price map cannot classify (an unmapped id, and the invoice amount matches no legacy price) used to be folded into `stillMissing` with no signal, even though no later sweep can resolve it. It now has its own `indeterminate` count, a per-row warning (ledgerId, userId, paidCents), and `missedGrantsIndeterminate` in the cron response and audit.
- **Corrected claim:** the amount fallback uses the invoice's *paid* amount, not the price's list amount that the webhook passes, and it applies to each of the user's subscription rows. An exact legacy-amount match is required, so a discounted or taxed total stays indeterminate.
- **Test resolver fixed:** it now mirrors the real `LEGACY_PRICE_AMOUNTS` table (1500 also maps to Pro), and the indeterminate test uses a non-legacy amount.
- **Concurrency test hardened:** it now waits until both claim UPDATEs are blocked behind the blocker's *own* pid (the second one transitively, via the tuple-lock queue), so lock waits from other suites in a shared CI database cannot release it early. The runs are always awaited, even on a timeout. Locally it passed 3/3 runs, and the dropped-guard mutant still fails it.

**Out of scope, raised for orchestrator triage (not decided here):**
1. Pro and Business share one ratio (6000 bps), so an indeterminate but entitled row *could* be granted at that shared ratio instead of waiting for a human. That is a pricing-policy choice.
2. A user who cancels before the reconcile runs never receives the grant for the invoice they paid (fails closed and counts as `stillMissing`).
3. Pre-existing: a late reconcile overwrites `monthlyAllowanceCents` with the old invoice's allowance.

## What changed
- **(A)** The claim is `WHERE id = $1 AND entryType = 'missed_grant'`. The balance update and the emit run only when the claim actually affected a row (`'granted'`). `reconciled` no longer counts a row that another run already claimed.
- **(B)** Keyset pagination on `id` (`gt(id, cursor)`, `orderBy(id)`, pages of 200) across the whole candidate set on each run.
- **(C)** Each page does one batched `subscriptions` lookup, then `deriveTierFromSubscriptions`. `reconcileMissedGrants({ priceTier })` now requires a price resolver, and the cron route passes in `getTierFromPrice`, the same map `reconcile-subscription-tiers` uses. **Policy choice for review:** an *indeterminate* derivation (an active subscription on an unmapped price) stays `still_missing` and is not granted from the lower-bound tier. Converting the row is final, so the grant must never be sized from a tier that may be wrong. This matches the fail-closed stance of D-OW-16 and `isTierDriftRepairable`.
- **(D)** `MissedGrantReconcileResult.failed` is new. Per-row and per-page lookup errors are logged at `error`. `/api/cron/reconcile-credits` returns **500** with all counts and skips the success audit when `failed > 0`.

## Mutation check (one line changed per mutant; source restored after each)
| Mutant | Result |
|---|---|
| control (no-op) | 49/49 pass |
| (A) drop `entryType` from the claim predicate | concurrency test fails |
| (B) stop after the first page | bounded-sweep test fails |
| (C) grant indeterminate derivations | indeterminate test fails |
| (D) don't count `failed` | per-row failure test fails |
| (D) route ignores `failed` | route test fails |
| (D) don't count page-level failures | page failure test fails |
| legacy fallback: pass `null` instead of paidCents | legacy-price test fails |
| indeterminate counted as stillMissing / granted | indeterminate test fails (both) |
| route drops `missedGrantsIndeterminate` | route test fails |

## Local verification
Other agents were running, so I ran single files only, from source against a local PG17. CI is the full gate.
- `credits-flow.integration.test.ts` + `missed-grant-reconcile.test.ts`: 56/56 (51 in credits-flow)
- `missed-grant-reconcile-concurrency.integration.test.ts` (real Postgres): 1/1
- `reconcile-credits/__tests__/route.test.ts`: 8/8
- `packages/lib` `tsc --noEmit`: 0 errors. eslint on touched files: 0 errors.

https://claude.ai/code/session_01FuCNrqW9utVJ8StCimVvXN


apps/web/src/app/api/cron/reconcile-credits/route.ts
```diff
@@ -26,10 +30,24 @@ export async function GET(request: Request) {
 
   try {
     const result = await backfillCredits();
-    const missedGrants = await reconcileMissedGrants();
+    const missedGrants = await reconcileMissedGrants({ priceTier: (priceId) => getTierFromPrice(priceId) });
```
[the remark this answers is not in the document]
2witstudios: Agreed, this was a real gap. Fixed in f4aba0674. The subscriptions table stores no amount, so the reconcile now passes the missed invoice's own `paidCents` as the amount: `priceTier(priceId, row.paidCents)`, which the route wires to `getTierFromPrice(priceId, amountCents)`. An unmapped legacy price id therefore goes through the same exact-match `LEGACY_PRICE_AMOUNTS` fallback the webhook uses. It still never reads the users cache (the concern in the earlier #2645 thread). An amount that matches nothing (for example a discount or tax changed the total) stays indeterminate and fails closed. Tests: `credits-flow…` › 'a legacy subscription on an unmapped price id resolves through the invoice amount…' (paid 2999 on an unmapped price → Pro grant), and the route test now asserts the amount reaches getTierFromPrice. Both failed before the change. A mutant that passes `null` instead of paidCents fails the lib test. Leaving this thread open for verification.

2witstudios:
@codex review

2witstudios:
@codex review

packages/lib/src/billing/missed-grant-reconcile.ts
```diff
@@ -75,103 +101,151 @@ export function planMissedGrantReconcile(row: MissedGrantRow, resolvedTier: stri
...
+        rows: page.length,
       });
+      if (page.length < BATCH) break;
+      continue;
     }
+
+    for (const row of page) {
+      try {
+        const derived = deriveTierFromSubscriptions(rowsByUser.get(row.userId) ?? [], (priceId) =>
+          options.priceTier(priceId, row.paidCents),
```
2witstudios: **P2  The invoice amount classifies subscription rows it does not describe, which skips the indeterminate gate for full-price invoices**

`row.paidCents` belongs to one invoice, but this closure passes it as the price amount for *every* entitled subscription row the user has. `LEGACY_PRICE_AMOUNTS` holds the *current* list prices too (1500 → pro, 5000 → business). So any missed full-price invoice turns *any* unmapped entitled row into a paid tier, and `indeterminate` never fires.

Concrete case: a user paid a $15 Pro invoice that became a missed grant, then canceled that Pro subscription. They still hold an active `subscriptions` row on a price id that is not in `STRIPE_PRICE_TO_TIER` (a subscription whose metadata kind is unrecognised and so falls back to `account_plan`, or a price removed from the env config). Before f4aba0674 that row was `indeterminate`, so the grant waited for a human. Now `getTierFromPrice(unmappedId, 1500)` returns `'pro'`, and the row is converted and granted, for a plan the user no longer holds. That is exactly what the doc comment at 162–166 says must never happen ("never be sized from a tier we know may be wrong").

Bounded: the grant is still `paidCents × ratio` for an invoice that was paid, and Pro/Business share a ratio today. That is why this is P2, not P1.

Fix direction: use the invoice amount as the fallback only when the user has exactly one entitled row. Better still, record the invoice's subscription id on the `missed_grant` row and match on it, so the amount is only used for the row it was billed on. Add a test with one mapped canceled row plus one unmapped active row at `paidCents` 1500, and assert `indeterminate: 1`.

apps/web/src/app/api/cron/reconcile-credits/route.ts
```diff
@@ -26,10 +30,27 @@ export async function GET(request: Request) {
...
+    });
+    const counts = {
+      ...result,
+      missedGrantsReconciled: missedGrants.reconciled,
+      missedGrantsStillMissing: missedGrants.stillMissing,
+      missedGrantsIndeterminate: missedGrants.indeterminate,
+      missedGrantsFailed: missedGrants.failed,
+    };
+
+    if (missedGrants.failed > 0) {
```
2witstudios: **P2  Indeterminate rows are paid-but-ungranted forever, and nothing alerts on them**

The lib's own doc says an indeterminate row "cannot resolve on its own … needs a human". But this branch only fails the run on `failed`. For `indeterminate > 0` the route returns 200, prints `console.log`, and writes a routine `data.write` audit. The only per-row signal is `loggers.api.warn` inside the lib. Nothing calls the incident notifier, and the loggers do not reach an alerting sink. So in practice a paying user can go without credits indefinitely and nobody is told. Only someone reading audit details would see `missedGrantsIndeterminate`.

The independent-review follow-up in the PR body describes these rows as "surfaced". Today that means a count in a 200 response.

Fix direction: log `indeterminate > 0` at `loggers.system.error`, like the `failed` path, with the ledger ids, or send it to whatever the cron monitor watches. Keep the 200 if a retry really cannot help, but make the signal as loud as the failure path. Route test: with `indeterminate: 2`, assert the error logger was called.

2witstudios (commented):
[independent-review] agent:ow-irv-2647 — no P1; 2 P2 (fail-closed indeterminate gate is bypassed for full-price invoices; indeterminate rows raise no alert)

What I checked, reading the code at 332cbe6df and not just the test names:
- **Thread A (double grant):** the claim `UPDATE … WHERE id AND entryType='missed_grant'` runs first in the transaction. The balance lock/refill and the `reconciled`/emit run only when it returns a row. Under READ COMMITTED a queued second claim re-checks the predicate and matches nothing. The concurrency test really waits for 2 claim UPDATEs blocked behind the blocker's own pid, so removing the guard makes it red. Resolved.
- **Webhook vs reconcile:** `credit_ledger_stripe_ref_unique` means a redelivered `invoice.paid` for a converted invoice hits `onConflictDoNothing` and does not refill. The reconcile keeps stripeRef, so there is no second grant. Lock order is ledger row, then balance, on both paths, so no deadlock.
- **Thread B (starvation):** keyset on `id` with `gt`/`orderBy` under the same DB collation, cursor starting at `''`. A failed page still moves the cursor forward. Resolved.
- **Thread C (tier from cache):** `users.subscriptionTier` is never read. Published-app dedicated subscriptions live in `published_app_subscriptions`, not `subscriptions`, so they cannot pollute the derivation. Resolved (but see P2 #1 on the amount fallback added later).
- **Thread D (swallowed failures):** row and page failures are counted, `failed>0` returns 500 with no success audit, and the route test proves it. Resolved.
- **Codex P2 (legacy amount fallback, f4aba0674):** the route really forwards `amountCents`. With `(p,a)=>getTierFromPrice(p)` the route test's `priceTier('price_legacy_unmapped', 2999)` would fail.
- Converted rows keep `consumeStatus='applied'`, so the backfill sweep never claws them back. `missed_grant` is only written when `paidCents>0`, so a row with 0 paid can never become a zero-amount grant.
- Minor, not a finding: the lib test's `LEGACY_AMOUNT_TIER` is a hand copy of `LEGACY_PRICE_AMOUNTS` and nothing ties the two together. The route test mocks `getTierFromPrice`, so no test uses the real table with the reconcile.

Already raised in the PR body and not repeated here: a cancel before the reconcile runs means no grant, the late reconcile overwrites `monthlyAllowanceCents`, and Pro/Business share a ratio.

packages/lib/src/billing/missed-grant-reconcile.ts
2witstudios: Confirmed and fixed in fbd56cc10. The invoice amount is now passed to the resolver **only when the user has exactly one subscription row, of any status**, because then the missed invoice must have been billed on it. With more rows no amount is passed, so an unmapped entitled row stays `indeterminate`. I didn't use the 'exactly one entitled row' variant: in your own scenario (a canceled mapped Pro row plus one unmapped active row) that row is the only entitled one, so it would still have granted. Recording the invoice's subscription id on the ledger row would be more precise, but it needs a schema change, so it is out of scope for this fix. Test: 'the invoice amount never classifies a subscription it may not describe — a canceled mapped Pro row plus an unmapped active row at a full-price 1500 stays indeterminate'. Before the fix it failed with `expected { reconciled: 1 } to match { reconciled: 0, indeterminate: 1 }`. The single-subscription legacy test still grants. Mutants: always passing paidCents fails the new test, and never passing it fails the legacy test. Leaving this thread open for verification.

apps/web/src/app/api/cron/reconcile-credits/route.ts
2witstudios: Agreed, fixed in fbd56cc10. When `indeterminate > 0` the route now logs at `loggers.system.error`, the same sink as the `failed` path. The log carries all counts plus `indeterminateLedgerIds`, which the lib now returns, capped at 50. The run stays 200 because a retry cannot resolve these rows. Route tests: with `indeterminate: 2` the error logger is called with `indeterminateLedgerIds: ['led_a','led_b']`, and a clean run asserts the error logger is **not** called, so a mutant that always alerts is caught too. Both mutants (never alert / always alert) fail the tests. On the wider gap: PageSpace loggers don't reach an alerting sink today, and neither does the `failed` path. Wiring a real incident notifier is a platform gap, not something this PR can fix. Leaving this thread open for verification.

2witstudios (commented):
[independent-review] agent:ow-irv-2647 — no new findings on fbd56cc10. Both P2s from review 5225246976 are genuinely fixed.

Reviewed `git diff 332cbe6df fbd56cc10` (one commit, 4 files).

**P2 #1 — the invoice amount deciding the tier of other subscriptions: FIXED.** `missed-grant-reconcile.ts:221-223` now passes `row.paidCents` as the price amount only when the user has exactly one `subscriptions` row of *any* status. Otherwise it passes `null`. In my scenario (a canceled mapped Pro row plus an active unmapped row) the user has two rows, so no amount goes through, the unmapped row returns 'free', `indeterminate` fires, and nothing is granted. Counting canceled rows is right: a single row means the user has held only one plan, so the invoice must be for it. **Checked by breaking the fix and running the tests** (source restored afterwards, tree clean):
- If line 222 always passes the amount (the old behaviour), the new test *'a canceled mapped Pro row plus an unmapped active row at a full-price 1500 stays indeterminate'* fails (it reconciles instead).
- If line 222 always passes `null`, the legacy-amount test fails. So the Codex legacy-amount fix is still protected in the other direction.
- On the unchanged code the whole `credits-flow.integration.test.ts` file passes, 55/55. That includes the thread B keyset test, the thread C tier-from-subscriptions tests, and the thread D per-row and per-page failure tests.

**P2 #2 — no alert on indeterminate rows: FIXED.** `route.ts:44-52` now logs at `loggers.system.error` whenever `indeterminate > 0`, with the counts and up to 50 `indeterminateLedgerIds`. That is the same level and logger as the `failed` path, and the response stays 200, as intended. The route test asserts `mockLogError` was called with an 'indeterminate' message and the exact ids, and that it is *not* called on a clean run. I could not run the web vitest in this worktree (it cannot find the `@tailwindcss/postcss` module). By reading: when `failed` is 0, the new block is the only call to `mockLogError`, so reverting it breaks that assertion. The ids are collected up to the cap while `indeterminate` counts every row, so the count stays exact.

**Threads A–D, rechecked:** the claim UPDATE, the balance lock, the keyset paging and the failure counting were not touched by this commit. The concurrency integration test is unchanged. Every `toEqual` on the result shape was updated to include `indeterminateLedgerIds`, so none of them was loosened.

**Remaining risk (not a finding):** the one-row rule assumes the invoice was billed on that row. It can be wrong if the subscription webhook for the billed plan never arrived *and* the user's only row is some other active subscription on an unmapped price that happens to match a legacy amount. That needs two faults at once, and the grant is still paid × ratio.

apps/web/src/app/api/cron/reconcile-credits/route.ts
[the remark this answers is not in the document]
2witstudios: Independently verified by the orchestrator on #2647 head fbd56cc10 (CI 10/10 green), and confirmed by the independent re-review of that head (no new findings): the reconcile passes the missed invoice's own `paidCents` as the amount to `getTierFromPrice(priceId, amountCents)`, so an unmapped legacy price still resolves through the legacy amount table (the subscriptions table stores no amount). Resolving.

packages/lib/src/billing/missed-grant-reconcile.ts
2witstudios: Independently verified by the orchestrator on #2647 head fbd56cc10 (CI 10/10 green), and confirmed by the independent re-review of that head (no new findings): the amount is passed only when it can describe the subscription — `const invoiceAmount = userSubs.length === 1 ? row.paidCents : null` — so with several subscription rows no amount is used and an unmapped entitled row stays indeterminate rather than being misclassified. Resolving.

apps/web/src/app/api/cron/reconcile-credits/route.ts
2witstudios: Independently verified by the orchestrator on #2647 head fbd56cc10 (CI 10/10 green), and confirmed by the independent re-review of that head (no new findings): when `missedGrants.indeterminate > 0` the route logs at error with `indeterminateLedgerIds`, so a paid-but-ungranted row now raises an alert naming the ledger ids instead of a silent 200. Resolving.

PR #2657, opened 2026-09-17
fix(spec-gate): Review 0b follow-ups — WAL-5 re-allowlist, re-add rule, Spec mirror re-sync, MON-6 literal pins

2witstudios:
Phase 0 follow-ups 2 to 5 from **Review 0b** (page `wiv716vr39v9wbx2z40aoio1`, section "Review 0b — fix round", Follow-ups). Base: `pu/org-wallets` @ 67463d41f (merge-base = tip).

## 1. WAL-5 was a false positive
Per Review 0b (Follow-up 2), WAL-5 is re-added to `scripts/spec-coverage-allowlist.txt`. WAL-5 says usage rows record the wallet charged, holds are per wallet, and ledger, reconciliation and monthly reset are keyed on wallet. No wallet exists yet. The tests that named it check user-keyed missed-grant reconciliation, so they prove MON-2 only:
- `missed-grant-reconcile-concurrency.integration.test.ts`: "MON-2 WAL-5 two overlapping…" becomes "MON-2 two overlapping…"
- `cron/reconcile-credits/__tests__/route.test.ts`: the 4 "MON-2 WAL-5 …" titles become "MON-2 …"
- `credits-flow.integration.test.ts`: describe "(MON-2, WAL-5)" becomes "(MON-2)". "WAL-5 bounded sweep" becomes "MON-2 bounded sweep" (a repairable paid row still gets its amount_paid × ratio grant). "WAL-5 billing disabled (tenant/onprem)…" loses its ID, because a deployment-mode no-op proves neither MON-2 nor WAL-5.
- Source comments in `missed-grant-reconcile.ts` and `cron/reconcile-credits/route.ts` no longer cite WAL-5.

**Whole-repo audit of the other test titles that name Spec IDs** (single- and multi-line `it`/`test`/`describe` titles under apps, packages, scripts and infrastructure):
- **Fixed (clear):** in `credit-copy.test.ts`, describe "MON-6 plan-card facts (**SEAT-2** lane A2)". The SEAT-2 was lane provenance, not a claim, but a describe title counts for every test inside it. None of those tests checks the tier table. It now reads "(lane A2)". SEAT-2 stays covered by subscription-tiers, plans and the checkout route tests.
- **Reported, not changed (borderline, left for a review to rule on):**
[...]

PR #2659, opened 2026-09-17
feat(db): org core schema — organizations, org_members, org_invitations, drives.orgId/orgVisibility, drive_members.source (Wave B1)

2witstudios:
## Wave B1 — org core schema (lands first and alone)

Leaf: izu8uzreac8uh7e3mpunqib4 · Task wqkdxnwbit8een989antta19 · Phase 1 vcs7u8l4v8rjgwfhkfft86hi. Decisions applied: D-OW-6 (materialize org membership as `drive_members` rows with `source`), D-OW-15 (drive slugs unique per org).

**The only Wave B migration: `0296_charming_dazzler`**, generated with `bun run db:generate`, no hand edits. Base is `pu/org-wallets` @ 67463d41f (latest there is 0295_groovy_risque).

### What lands
| Object | Shape |
|---|---|
| `organizations` | name, `slug` unique, avatarUrl, `ownerId` → users **ON DELETE RESTRICT** (ORG-6: an Owner can't delete their account while they own an org), `policies` jsonb NOT NULL default `{}`, nullable `stripeCustomerId` (unique) / `stripeSubscriptionId` |
| `org_members` | `OrgRole` OWNER\|ADMIN\|MEMBER default MEMBER, unique (orgId, userId), **at most one OWNER per org** (`org_members_one_owner_key`, partial unique on orgId WHERE role = 'OWNER'), cascades with org and user |
| `org_invitations` | `tokenHash` (unique, no raw token), email, role, `expiresAt` NOT NULL, `acceptedAt`; **one open invite per (orgId, lower(email))** as a partial unique index `WHERE acceptedAt IS NULL`, so it's case-insensitive; the email lookup index is on lower(email) too |
| `drives.orgId` | FK → organizations **ON DELETE RESTRICT** (not SET NULL; ORG-6) |
| `drives.orgVisibility` | `OrgDriveVisibility` OPEN\|RESTRICTED\|PRIVATE, NOT NULL default OPEN |
| `drives_org_slug_unique` | unique (orgId, slug) `WHERE orgId IS NOT NULL` |
| `drives_home_never_org_check` | `CHECK (kind <> 'HOME' OR orgId IS NULL)` |
| `drive_members.source` | `DriveMemberSource` invite\|org, NOT NULL default invite (existing rows become `invite`) |

Exports: `schema.ts` (re-export + combined `schema` object) and a new `@pagespace/db/schema/organizations` entry in the package.json exports map. `packages/db` has no `index.ts`; `schema.ts` is the barrel.

### Home drives can't get an orgId: CHECK, not a guard test
It's a database CHECK (`drives_home_never_org_check`), so no write path can skip it: move-in (B5), backfill, backup restore or a raw script. The known trap (a CHECK over an ON DELETE SET NULL column aborting the parent delete) doesn't apply here, because `orgId` is ON DELETE RESTRICT and no cascade ever rewrites it.

### Export registries (review round 1)
Adding these tables and columns trips two schema-derived drift guards, so each one now records a decision:
- **GDPR (Art 15) registry** (`packages/lib/src/compliance/export/gdpr-export-coverage.ts`):
  - `org_invitations` is excluded **permanently**, on the same Art 15(4) counterparty boundary as `pending_invites`.
  - `organizations` and `org_members` are excluded **temporarily** under Spec X-2. The Phase 6 GDPR/backups leaf yfmlkdchehmberwthwu6g7vt removes these exclusions before Wave F, and the lane that introduces the ORGS_ENABLED code constant adds a test that fails while they exist (a B3 requirement).
- **Tenant export registry** (`scripts/lib/tenant-export-columns.ts`):
  - `drives.orgVisibility` and `drive_members.source` are carried.
  - `drives.orgId` is excluded, because the bundle doesn't carry `organizations`. It's temporary under Spec X-3 and removed by the same leaf.
- `drive-service.test.ts`: the mock drive now includes `orgId` and `orgVisibility` (this fixes the lib typecheck).

### Tracked outside B1 (review)
- The org Owner membership row matching `organizations.ownerId` on create and transfer: B3 leaf t718n05yq2u9ukbil3i6hz5w. The schema already caps OWNER rows at one.
- Org drives surviving their lead's account erasure or delete: B6 leaf h40ht4zm4o2md5d5fktuhz6c. B5, the first writer of drives.orgId, doesn't merge before B6.

### Timestamps
New default timestamps use `(now() at time zone 'utc')`, not bare `now()`, because these columns are `timestamp without time zone`.

### ID → test
| ID | Test (file › name) |
|---|---|
| ORG-1 | `organizations.test.ts` › "ORG-1 an org has exactly one Owner membership row", "ORG-1 carries name, a globally unique slug…", "ORG-1 restricts deleting the owning user…", "ORG-1 holds org policies…", "ORG-1 keeps the Stripe fields nullable…"; `organizations.integration.test.ts` › "ORG-1 an org has exactly one Owner membership row", "ORG-1 creates an org with an empty policy object…", "ORG-1 refuses deleting a user who owns an org (ORG-6)" |
| ORG-2 | `organizations.test.ts` › "ORG-2 exports the closed OrgRole set…", "ORG-2 allows a user in many orgs but only once per org", "ORG-2 cascades membership rows…"; integration › "ORG-2 lets one user join many orgs, but only once per org" |
| ORG-3 | `organizations.test.ts` › "ORG-3 stores only a token hash…", "ORG-3 allows one open (unaccepted) invite per org and case-insensitive email"; integration › "ORG-3 an open invite for A@X and a@x in one org cannot coexist", "ORG-3 allows one open invite per org and email, and a new one once the old is accepted" |
| DRV-1 | `organizations.test.ts` › "DRV-1 a drive can be owned by an org…", "DRV-1 Home drives never can…", "DRV-1 drive slugs are unique per org…"; integration › "DRV-1 refuses giving a Home drive an orgId", "DRV-1 refuses deleting an org that still owns drives…", "DRV-1 keeps drive slugs unique within an org…" |
| DRV-4 | `organizations.test.ts` › "DRV-4 org visibility is Open, Restricted or Private and defaults to Open"; integration › "DRV-4 defaults a drive to Open visibility" |
| DRV-8 | `organizations.test.ts` › "DRV-8 distinguishes invited members (guests included) from org-materialized rows…"; integration › "DRV-8 records a directly added member (a guest) with source 'invite' by default" |

### Per-ID decisions (gate rule from #2657; reviews 5230157358, 5230240628)
| ID | Spec clauses | What the B1 tests prove | Decision |
|---|---|---|---|
| ORG-1 | name, slug, avatar; one human Owner; ownership transferable | the fields exist; `ownerId` is RESTRICT; OWNER rows are capped at one | **(partial)**, back on the allowlist. Owner row matching `ownerId` and transfer: B3 |
| ORG-2 | Owner/Admin/Member; many orgs per user; a user in no org sees no org UI | the role enum; unique (orgId, userId); cascades | **(partial)**, back on the allowlist. The no-org-UI clause: Wave F |
| ORG-3 | email invites with expiry, resend, revoke, acceptance; a pending invite counts as a seat | token hash only; one open invite per org and case-insensitive email | **(partial)**, back on the allowlist. Invite lifecycle: B3; seats: Wave C |
| DRV-1 | org-owned drives; Home drives never can; the human lead keeps the Owner role | `orgId` RESTRICT; the Home CHECK; per-org slug uniqueness | **(partial)**, back on the allowlist. Lead keeps the Owner role: B5/B6 |
| DRV-4 | each org drive has an Open/Restricted/Private visibility; new org drives default to Open | NOT NULL enum with exactly those three values; DB default OPEN; an insert without a value lands OPEN (integration) | **plain, covered**. The requirement is fully a data-model fact; DRV-5..7 carry the effects |
| DRV-8 | guest = drive member not in the org; no seat; sees only invited drives; labeled | only that `source` separates invited rows from org-materialized rows | **(partial)**, back on the allowlist. Seats: Wave C; visibility: B4/B7; label: Wave F |
| ORG-6 | org delete transfers or trashes drives, nothing orphaned; an Owner can't delete their account while owning an org | only the RESTRICT FKs refusing those deletes | titles retitled **ORG-6 (partial)** and it stays allowlisted, which fixes the STALE-ALLOW. Transfer/trash flow: B6 |

The allowlist now omits only **DRV-4**. ORG-1/2/3, DRV-1 and DRV-8 are back on it, each with a comment citing the reviews and the lane that proves the rest, and ORG-6 stays on it. Every other test title named in the table above now carries ` (partial)` after those IDs.

Gate evidence at 369a04275, local:
- `bun run scripts/check-spec-coverage.ts --offline --ids ORG-1,ORG-2,ORG-3,ORG-6,DRV-1,DRV-4,DRV-8` against JSON results from the two files: `covered 1/7, allowlisted-missing 6, MISSING 0`, then `spec-coverage: OK`.
- The gate's self-tests (`scripts/__tests__/check-spec-coverage.test.ts`): 47/47.

The schema also enables O-6 (materialize: `drive_members.source`) and O-15 (per-org slugs). Those are decision IDs, not coverage-gate IDs.

### Evidence (local, single files; build-slot rule)
- `bun run --filter @pagespace/db test -- src/schema/__tests__/organizations.test.ts`: **16 passed (16)**
- `bun run --filter '@pagespace/db' test:integration -- organizations` against a fresh homebrew Postgres 17 running in a **non-UTC** zone (America/Chicago): **11 passed (11)**
- Migration on a **fresh DB**: 0000–0296 applied, exit 0.
- Migration on an **existing DB**: I migrated to `origin/pu/org-wallets` (296 rows), seeded a HOME drive, a STANDARD drive and a drive_member, then migrated this branch. Result: 297 rows; both drives have `orgId` NULL and visibility OPEN; the member has `source = invite`.
- `bun run --filter @pagespace/lib test -- src/services/__tests__/drive-service.test.ts`: 40 passed (40).
- Registry guards: the worktree's `@pagespace/db` resolves to a stale main-checkout dist, so I checked both registries against the worktree schema source with a script: no undecided tables, no phantom tables, no missing or phantom columns. Controls with `org_members` or `source` removed are flagged. CI is authoritative.
- Nearby suites: schema-definitions 76/76, schema-coverage 35/35, migrate 7/7, migration-runner 6/6, migration-sql-analysis 14/14, lib `policy-reads.seam` 2/2. That seam is now **armed**, because an `organizations` pgTable exists; no route imports it.
- **Mutation check (line index):** I commented out `packages/db/src/schema/core.ts` line 57 (`homeNeverOrg: check(...)`), and `DRV-1 Home drives never can: a CHECK refuses orgId on a HOME drive` went red (1 failed / 14 passed). At the DB layer, dropping the constraint turned `DRV-1 refuses giving a Home drive an orgId` red (1 failed / 8 passed). Round 1: commenting out `organizations.ts` line 60 (the one-owner index) turned "ORG-1 an org has exactly one Owner membership row" red, and changing line 84 back to plain `table.email` turned "ORG-3 … case-insensitive email" red. Replacing both indexes on the test DB turned both new integration tests red (2 failed / 9 passed). After restoring everything, 16/16 and 11/11 pass and `git status` is clean.

Not run locally (build-slot rule; CI is the gate): monorepo `typecheck`, `lint`, `test:coverage`.

https://claude.ai/code/session_01FmvQdNKpCWKjQHHaqvwQ3A


2witstudios:
@codex review

packages/db/src/schema/core.ts
```diff
@@ -14,13 +15,22 @@ export const pageType = pgEnum('PageType', ['FOLDER', 'DOCUMENT', 'CHANNEL', 'AI
...
 
 export const drives = pgTable('drives', {
   id: text('id').primaryKey().$defaultFn(() => createId()),
   name: text('name').notNull(),
   slug: text('slug').notNull(),
   ownerId: text('ownerId').notNull().references(() => users.id, { onDelete: 'cascade' }),
   kind: driveKind('kind').default('STANDARD').notNull(),
+  // Set when an org owns the drive (DRV-1); the human lead stays ownerId. RESTRICT, not
+  // SET NULL: deleting an org must first transfer or trash its drives (ORG-6).
+  orgId: text('orgId').references((): AnyPgColumn => organizations.id, { onDelete: 'restrict' }),
```
2witstudios: **P2-1: ORG-6 / D-OW-7. An org drive is still hard-deleted when its lead deletes their account.** `orgId` RESTRICT protects org drives from an *org* delete, but `drives.ownerId` (line 28, unchanged) stays `ON DELETE CASCADE`. Verified on pg17: org `o2` is owned by `a1`. Its drive `od` has lead `a2`, an org member but not the org owner. `DELETE FROM users WHERE id='a2'` succeeds and `od` and all its content are gone (`org_drive_left=0`).

Spec: ORG-6 "nothing is orphaned". D-OW-7 (accepted): "on leave or delete the drive reassigns to the org Owner". Right now only an app-level reassign stops this, and any delete path that skips it (admin tooling, GDPR erase, a raw script) destroys org data silently. That's the same argument this PR makes for putting the Home rule in a CHECK.

The fix is a triage call, not mine. Either a DB guard in this migration, e.g. a `BEFORE DELETE ON users` trigger that refuses when the user leads a drive with `orgId IS NOT NULL` (mirroring `organizations.ownerId` RESTRICT). Or an explicit, tested contract on the D-OW-7 lane that account deletion reassigns before delete.

packages/db/src/schema/organizations.ts
```diff
@@ -0,0 +1,102 @@
...
+  updatedAt: timestamp('updatedAt', { mode: 'date' }).default(utcNow).notNull().$onUpdate(() => new Date()),
+}, (table) => ({
+  ownerIdx: index('organizations_owner_id_idx').on(table.ownerId),
+}));
+
+export const orgMembers = pgTable('org_members', {
+  id: text('id').primaryKey().$defaultFn(() => createId()),
+  orgId: text('orgId').notNull().references(() => organizations.id, { onDelete: 'cascade' }),
+  userId: text('userId').notNull().references(() => users.id, { onDelete: 'cascade' }),
+  role: orgRole('role').default('MEMBER').notNull(),
```
2witstudios: **P2-2: ORG-1 "one human Owner" has two sources of truth and neither constrains the other.** `organizations.ownerId` names the Owner, while `org_members.role` also allows `OWNER`, with no limit on count and no link to `ownerId`. Verified on pg17: org `o2` has `ownerId=a1`, and inserting `OWNER` rows for both `a2` and `a3` succeeds (`owner_rows=2`).

ORG-4 resolves Owner and Admin to full access on every org drive, and ORG-5 authorizes mutations from membership. So a stray `OWNER` row grants owner-level power to someone who isn't the Owner, and transfer (ORG-1) can leave the two stores disagreeing.

Options: a partial unique index `(orgId) WHERE role = 'OWNER'`, plus the transfer lane keeping it equal to `ownerId`. Or keep `OWNER` out of member rows and derive it only from `ownerId`. Either way, settle it before the org access core lane reads roles.

packages/db/src/schema/organizations.ts
```diff
@@ -0,0 +1,102 @@
...
+  tokenHash: text('tokenHash').unique().notNull(),
+  invitedBy: text('invitedBy').references(() => users.id, { onDelete: 'set null' }),
+  expiresAt: timestamp('expiresAt', { mode: 'date' }).notNull(),
+  acceptedAt: timestamp('acceptedAt', { mode: 'date' }),
+  createdAt: timestamp('createdAt', { mode: 'date' }).default(utcNow).notNull(),
+}, (table) => ({
+  emailIdx: index('org_invitations_email_idx').on(table.email),
+  // One open invite per (org, email). Resend rotates the open row; revoke deletes it.
+  // Callers normalize email before writing, as pending_invites does.
+  openOrgEmailKey: uniqueIndex('org_invitations_open_org_email_key')
```
2witstudios: **P2-3: ORG-3 / SEAT-3. The open-invite key is case-sensitive, and expired rows stay "open".** Verified on pg17: with an open invite for `a@x.test`, inserting `A@X.test` into the same org succeeds. That's two open invites to one inbox, and under SEAT-3 ("a pending invite counts as a seat") two billable seats. The comment hands normalization to callers, as `pending_invites` does, but that key also relies on caller discipline. An expression key `(orgId, lower(email)) WHERE acceptedAt IS NULL` makes the DB the guard.

Also verified: an **expired**, unaccepted invite blocks any new INSERT for that email forever. That works only if the invite service's "re-invite after expiry" rotates the existing row, which the comment names for resend and revoke but not for expiry. And the SEAT-3 seat count must exclude `expiresAt <= now()`, or dead invites hold paid seats. Please pin both as tested contracts on the invitations/seats lanes, if not in the schema.

2witstudios (commented):
[independent-review] agent:ow-irv-2659

Spec-conformance review of `1ba6886b1` against Spec drc7x34unhc0ty1dc0u1j3gy (ORG, DRV, SEAT, WAL, O-6, O-15), leaf izu8uzreac8uh7e3mpunqib4 and D-OW-6/7/8/10/15. **No P1. 3 × P2 (inline).** The schema does what the leaf names. Every P2 below is a spec invariant this schema still leaves open, and a later lane would inherit it.

## Per-ID verdict
| ID | Verdict | Evidence |
|---|---|---|
| ORG-1 | **Satisfied, with a gap** | `organizations.test.ts` › "ORG-1 carries name, a globally unique slug…", "ORG-1 restricts deleting the owning user…"; integration › "ORG-1 refuses deleting a user who owns an org (ORG-6)". **Gap:** "one human Owner" is not enforced. `org_members` accepts any number of `OWNER` rows, even for users who are not `organizations.ownerId` (P2-2). Transferable: `ownerId` is a plain updatable FK. |
| ORG-2 | Satisfied | `organizations.test.ts` › "ORG-2 allows a user in many orgs but only once per org", "ORG-2 cascades membership rows…"; integration › "ORG-2 lets one user join many orgs, but only once per org" |
| ORG-3 | **Satisfied, with a gap** | `organizations.test.ts` › "ORG-3 stores only a token hash…", "ORG-3 allows one open (unaccepted) invite…"; integration › "ORG-3 allows one open invite per org and email…". **Gap:** the open-invite key is case-sensitive, and expired unaccepted rows still count as open (P2-3). |
| DRV-1 | **Satisfied, with a gap** | `organizations.test.ts` › "DRV-1 Home drives never can…", "DRV-1 a drive can be owned by an org…", "DRV-1 drive slugs are unique per org…"; integration › "DRV-1 refuses giving a Home drive an orgId", "DRV-1 refuses deleting an org that still owns drives…". **Gap:** `drives.ownerId` still CASCADEs, so a lead's account delete hard-deletes the org drive (P2-1, against ORG-6 and D-OW-7). |
| DRV-4 | Satisfied | `organizations.test.ts` › "DRV-4 org visibility is Open, Restricted or Private and defaults to Open"; integration › "DRV-4 defaults a drive to Open visibility" |
| DRV-8 | Satisfied (schema part) | `organizations.test.ts` › "DRV-8 distinguishes invited members…"; integration › "DRV-8 records a directly added member (a guest) with source 'invite'…". Guest status is derived (a drive member with no `org_members` row). `source` does not mark it, which is correct. |
| O-6 / D-OW-6 | Satisfied | `drive_members.source` enum `invite\|org`, NOT NULL, default `invite`. Existing rows backfilled to `invite` (verified). |
| O-15 / D-OW-15 | Satisfied | `drives_org_slug_unique (orgId, slug) WHERE orgId IS NOT NULL`. Personal slug rules are unchanged. `drives_owner_id_slug_key` was never unique: two personal `product` drives for one owner insert fine, before and after this PR. The spec's "unique per owner today" describes app behaviour, not a DB constraint. |

**Other ORG/DRV schema this leaf should have laid:** only the single-Owner invariant (P2-2). DRV-6 join requests, the ORG-4 audit (via the AUD chain) and SEAT-8/9 trial or lapsed state have no column here. None of those IDs are in this leaf's scope. Noting it so the SEAT/Wave C lanes know `organizations` has no subscription status or trial-end column yet.

## FK delete actions (verified in a real Postgres 17, session TZ America/Chicago)
| FK | Action | Probe result |
|---|---|---|
| organizations.ownerId → users | RESTRICT | Deleting the owner fails with `organizations_ownerId_users_id_fk` ✅ |
| drives.orgId → organizations | RESTRICT | Deleting an org with drives fails ✅. **A trashed drive also blocks it** (see note) |
| org_members.orgId / org_invitations.orgId | CASCADE | After purging the drives, deleting the org leaves members=0 and invites=0 ✅ |
| org_members.userId → users | CASCADE | Deleting a member user removes their row ✅. So D-OW-8 leave or delete cascades need no schema change here. |
| invitedBy (both tables) | SET NULL | No CHECK touches `invitedBy` ✅ |
| drive_members.source | default `invite` | A raw insert without `source` gives `invite` ✅ |
| **drives.ownerId → users (unchanged)** | **CASCADE** | **The lead (an org member, not the org owner) deletes their account and the org drive is gone (`org_drive_left=0`).** See P2-1. |

**CHECK over SET NULL:** the only new CHECK is `drives_home_never_org_check (kind, orgId)`. `orgId` is RESTRICT and `kind` has no FK, so no cascade can abort a parent delete ✅.

**ORG-6 contract note (P3, not inline):** RESTRICT counts trashed drives. So "trash its drives, then delete the org" can't finish unless the org-delete lane hard-deletes trashed org drives or nulls their `orgId`. Nulling `orgId` silently turns a drive personal. That lane has to pick one on purpose.

## Home drives can't get an orgId: a CHECK, attacked directly
Every probe was refused with `23514 drives_home_never_org_check`: INSERT HOME+orgId, UPDATE a HOME drive to set orgId, and UPDATE an org STANDARD drive to `kind='HOME'`. A single-statement `SET orgId, kind='STANDARD'` succeeds. The row stops being a Home drive, so that doesn't violate the invariant. Whether a Home drive may be re-kinded at all is app policy.

## Uniques
- `organizations_slug_unique` ✅; `(orgId,userId)` `org_members_org_user_key` ✅; `stripeCustomerId` unique ✅
- Open invite `(orgId,email) WHERE acceptedAt IS NULL` ✅. **An expired, unaccepted invite blocks a new INSERT forever** (verified). That's fine only if "re-invite after expiry" rotates the existing row. The comment covers resend and revoke but not re-invite after expiry. The SEAT-3 seat count must also filter `expiresAt > now()`, or dead invites hold seats. **`A@X.test` inserted beside `a@x.test`** (P2-3).
- `(orgId, slug)` partial ✅. An org drive and the lead's personal drive may share a slug ✅.

## Migration 0296
- **Generated:** on the PR tree, `drizzle-kit generate` reports "No schema changes, nothing to migrate". Snapshot, journal and SQL all agree.
- **Fresh DB:** 0000→0296 applied, 297 rows in `__drizzle_migrations`, exit 0.
- **No data rewrite:** only ADD COLUMN (`orgId` nullable; `orgVisibility`/`source` with constant defaults) and constraint additions. Existing HOME drives pass the CHECK because `orgId` is NULL.
- **Enum order:** all three `CREATE TYPE` statements run before the tables and ALTERs ✅.
- **UTC defaults:** in a Chicago session, `joinedAt` = `2026-09-17 01:51` UTC while local `now()` = `2026-09-16 20:51` ✅. No unit test asserts the UTC default. Minor: it is only guarded by review.

## Exports
`schema.ts` re-exports it and spreads it into `schema` ✅. `packages/db` has no `src/index.ts` (the barrel is `schema.ts`) ✅. `./schema/organizations` is in the package.json exports map ✅. No import cycle: core → organizations → auth, and auth doesn't import core.

## Tokens
`tokenHash` NOT NULL UNIQUE, backed by index `org_invitations_tokenHash_unique`. No `token` column exists in the table or in the migration ✅.

## Test honesty (run by me, single files)
- `bun run --filter @pagespace/db test -- src/schema/__tests__/organizations.test.ts`: **15 passed (15)**
- `test:integration -- organizations` on a fresh, fully migrated pg17 (TZ America/Chicago): **9 passed (9)**
- **Mutation 1 (source, by line index):** `organizations.ts` line 35 `onDelete: 'restrict'` → `'cascade'`. `ORG-1 restricts deleting the owning user…` went red (1 failed / 14 passed). Restored: 15/15.
- **Mutation 2 (DB):** replaced `org_invitations_open_org_email_key` with a non-partial unique. `ORG-3 allows one open invite per org and email, and a new one once the old is accepted` went red (1 failed / 8 passed). Restored: 9/9.
- `git status` is clean apart from my untracked review scratch dir. Postgres stopped and its data dir removed.

CI: this PR targets `pu/org-wallets`. At review time `ci / Unit Tests`, `Lint & TypeScript Check` and `Security Test Suite` were still pending. `Static Security Analysis` showed **fail**, but its run was still in progress so the logs weren't readable. The owning lane should check that job.

packages/db/src/schema.ts
```diff
@@ -66,6 +66,7 @@ export * from './schema/drive-env-local';
 export * from './schema/dev-preview-services';
 export * from './schema/dev-preview-grants';
 export * from './schema/agent-account-grant-nonces';
+export * from './schema/organizations';
```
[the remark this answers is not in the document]
2witstudios: Fixed in bfdef9305: `packages/lib/src/compliance/export/gdpr-export-coverage.ts` now records a decision for all three tables. `org_invitations` is excluded on the same Art 15(4) counterparty boundary as `pending_invites`. `organizations` and `org_members` are excluded for now, and the reasons say why: no code path writes these tables yet (ORGS_ENABLED is off), and the collectors are Spec X-2 (lane G2), which must move them to EXPORTED_TABLES before the flag turns on. Verified against the worktree schema source (the worktree's `@pagespace/db` resolves to a stale dist): undecided=[], phantom=[], short reasons=[]; a control with the `org_members` entry removed reports it undecided. CI on bfdef9305 is the authoritative run. Leaving this open for verification.

packages/db/src/schema/core.ts
```diff
@@ -14,13 +15,22 @@ export const pageType = pgEnum('PageType', ['FOLDER', 'DOCUMENT', 'CHANNEL', 'AI
...
+  orgVisibility: orgDriveVisibility('orgVisibility').default('OPEN').notNull(),
```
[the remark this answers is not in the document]
2witstudios: Fixed in bfdef9305: `scripts/lib/tenant-export-columns.ts` now carries `drives.orgVisibility` and `drive_members.source`. Both are plain enum columns with no FK, so a tenant keeps visibility and org-materialized membership. `drives.orgId` is excluded with a stated reason: it references `organizations`, which the bundle doesn't carry, so a carried value would break the FK on import. Carrying organizations + orgId is Spec X-3 (lane G2) and must land before ORGS_ENABLED, or a migrated org drive arrives as a personal drive. Checked the registry against the worktree schema source: no missing or phantom columns; a control with `source` removed reports it missing. CI on bfdef9305 is the authoritative run. Leaving this open.

packages/db/src/schema/core.ts
2witstudios: Tracked to B6 leaf h40ht4zm4o2md5d5fktuhz6c: reassign before the user delete, test ORG-6 (an org drive survives its lead's account deletion, reassigned to the org Owner per D-OW-7). No change in B1: `drives.ownerId` stays ON DELETE CASCADE here, and the orchestrator ruled the guard out of B1 scope. Leaving this open.

packages/db/src/schema/organizations.ts
2witstudios: Fixed in bfdef9305: `org_members_one_owner_key` is a partial unique index on (orgId) WHERE role = 'OWNER' (`packages/db/src/schema/organizations.ts`, migration 0296 regenerated as 0296_charming_dazzler). Tests: unit `ORG-1 an org has exactly one Owner membership row` (16/16), and the integration test of the same name proves that promoting a second member to OWNER and inserting a second OWNER row both fail with 23505 (11/11 on pg17, America/Chicago). Mutation: commented out organizations.ts line 60 → unit red (1 failed/15 passed); dropping the index on the test DB → integration red. Keeping it equal to `organizations.ownerId` during transfer is the transfer lane's transaction. Leaving this open for verification.

packages/db/src/schema/organizations.ts
2witstudios: Fixed in bfdef9305: the open-invite key is now (orgId, lower(email)) WHERE acceptedAt IS NULL, and the email lookup index is lower(email) as well. I chose the expression index over CHECK email = lower(email) because it makes the database dedupe whatever a caller writes, rather than refuse an un-normalized write, and it keeps the casing the inviter typed. Test: integration `ORG-3 an open invite for A@X and a@x in one org cannot coexist` (marcus.oyelaran@northwind.test vs Marcus.Oyelaran@Northwind.TEST → 23505). Mutation: organizations.ts line 84 back to table.email → unit red; the plain-email index on the test DB → integration red. The expired-invite rotation and the SEAT-3 count excluding expiresAt <= now() are on the B3 leaf per the orchestrator. Leaving this open.

2witstudios:
@codex review

packages/lib/src/compliance/export/gdpr-export-coverage.ts
```diff
@@ -412,6 +412,15 @@ export const EXCLUDED_TABLES: Readonly<Record<string, string>> = {
     'Binding rows between mirrored calendar events and drives — meaningless without the mirrored events, which are excluded above.',
   event_attendees:
     'Attendee list of a mirrored calendar event: it is a list of OTHER PEOPLE, which Art 15(4) puts outside the subject\'s access right.',
+  // Organizations & Wallets (Wave B1 schema). NOT a permanent decision for the first two:
+  // Spec X-2 (lane G2) owns the org collectors and must land them — moving these to
+  // EXPORTED_TABLES — before ORGS_ENABLED turns on. Until then no code path writes a row.
+  organizations:
+    'No row exists to disclose: organizations ship dark behind ORGS_ENABLED and no service writes this table yet. The collector for orgs the subject owns is Spec X-2 (Organizations & Wallets lane G2), which must land before the flag turns on.',
```
2witstudios: **P2-4: the gate these deferrals hang on doesn't exist.** `ORGS_ENABLED` appears nowhere in the repo except these reasons and the tenant `orgId` reason: `git grep ORGS_ENABLED` at `bfdef9305` returns only `gdpr-export-coverage.ts:417/419/421` and `tenant-export-columns.ts:129`. There's no env var, constant or deployment-mode guard.

So "must land before the flag turns on" can never fire, and "no service writes this table yet" goes false as soon as the Wave B org services lane merges. That happens before any flag exists, and nothing turns red.

The point-guard accepted these as deferrals gated on "the Phase 6 GDPR/backups leaf must remove them before Wave F". The reasons should name that real gate, or the flag must exist in code (a code constant, per the cross-app-switch rule) with a test that fails while it's on and the exclusions are still registered. Same fix for `org_members` (L421) and `scripts/lib/tenant-export-columns.ts:129`.

2witstudios (commented):
[independent-review] agent:ow-irv-2659 — re-review round 1 at `bfdef9305`

Checked against my round-0 review (pullrequestreview-5230157358) and the point-guard's round-1 summary. Evidence comes from a fresh homebrew Postgres 17 (session TZ America/Chicago), migrated from 0000 through 0296. The data dir was removed afterwards.

## FIXED / NOT FIXED
| Finding | Status | Evidence |
|---|---|---|
| **P2-1**: `drives.ownerId` CASCADE hard-deletes an org drive when its lead deletes their account (ORG-6 / D-OW-7) | **NOT FIXED (not addressed this round)** | No change to `drives.ownerId` or any guard in the diff, and it isn't mentioned in the round-1 summary. Still open for triage: a DB guard, or an explicit tested contract on the D-OW-7 lane. |
| **P2-2**: more than one `OWNER` row per org | **FIXED** | `org_members_one_owner_key (orgId) WHERE role='OWNER'`. pg17: a second OWNER **insert** fails with `23505 org_members_one_owner_key`; promoting an ADMIN row to OWNER fails the same way; an OWNER in a different org still inserts. Tests: `organizations.test.ts` › "ORG-1 an org has exactly one Owner membership row"; integration › "ORG-1 an org has exactly one Owner membership row". |
| **P2-3a**: `A@X` beside `a@x` | **FIXED** | Key is now `(orgId, lower(email)) WHERE acceptedAt IS NULL`. pg17: inserting `A@X.test` beside an open `a@x.test` fails with `23505`; the same address in another org succeeds. Test: integration › "ORG-3 an open invite for A@X and a@x in one org cannot coexist". |
| **P2-3b**: an accepted invite still allows a new one | **FIXED (preserved)** | pg17: after `acceptedAt` is set, a new open invite for `A@X.TEST` inserts. Integration › "ORG-3 allows one open invite per org and email, and a new one once the old is accepted" still passes. |
| **P2-3c**: an expired, unaccepted invite blocks re-invite and holds a SEAT-3 seat | **Unchanged: contract only** | Still true at the schema level. It isn't in the column comment, which names resend and revoke but not re-invite after expiry. The invitations and seats lanes must still pin it with tests. I'm not re-raising it as a blocker. |

## Residual notes on the fixes (P3, not inline)
- **Owner transfer is order-sensitive.** A unique index can't be deferred, so a single-statement swap (`UPDATE … SET role = CASE …`) fails with `23505`. Demote-then-promote as two statements in one transaction succeeds (verified). The column comment says exactly this. The transfer lane should follow it.
- **An `OWNER` row can still point at a different user than `organizations.ownerId`.** Nothing ties them together. The index caps the count at one; agreement with `ownerId` is still the transfer service's job.
- **Plain `email = …` lookups now seq-scan.** `org_invitations_email_idx` is now on `lower(email)`. On pg17, `WHERE lower(email)=…` uses a Bitmap Index Scan and `WHERE email=…` does a Seq Scan. Invitation services must query `lower(email)`. The comment says so.

## Migration 0296 (regenerated as `0296_charming_dazzler`)
- **Generated, no drift:** on the PR tree, `drizzle-kit generate` reports "No schema changes, nothing to migrate". `0296_wealthy_ricochet.sql` is renamed, and the journal `when`/`tag` and snapshot moved with it. The SQL delta is only the two expression indexes plus `org_members_one_owner_key`.
- **Applies after 0295 on a fresh DB:** the log shows `Applying: ffc88f54… (1789606582135)` (0295), then `Applying: 44384c34… (1789610476771)` (0296). 297 rows, exit 0.
- Caveat: any local or test DB that already applied the old `wealthy_ricochet` hash will re-run 0296, because the runner is hash-keyed, and fail on `CREATE TYPE` already existing. It never reached a shared environment (unmerged integration branch), so this is only a local reset.

## GDPR / tenant-export exclusions
| Entry | Names X-2/X-3 as temporary? | Note |
|---|---|---|
| `EXCLUDED_TABLES.organizations` | ✅ X-2, "must land before the flag turns on" | Gate names a flag that doesn't exist (P2-4) |
| `EXCLUDED_TABLES.org_members` | ✅ X-2, temporary | Same (P2-4) |
| `EXCLUDED_TABLES.org_invitations` | ❌ **Excluded, not exported**, with a permanent Art 15(4) reason | The round-1 summary says "org_invitations exported". It isn't. See P2-5. |
| `TENANT_EXPORT_COLUMNS.drives.excluded.orgId` | ✅ X-3, temporary | Same flag issue (P2-4) |
| `drives.orgVisibility`, `drive_members.source` carried | ✅ | Drift guard green |

### Can excluding `drives.orgId` break a tenant export or restore today? **No.**
- **No writer:** `git grep orgId` over `apps/`, `packages/lib/src`, `scripts/` (tests and docs excluded) finds only a code comment in `socket-utils.ts`. Nothing sets `drives.orgId`, so every source row has NULL.
- **Import:** the column is nullable with no default, so an import that omits it produces NULL. That passes `drives_home_never_org_check` and needs no `organizations` row, so the FK can't be violated.
- **Export:** it reads `SELECT * FROM drives` and emits only the registry columns, so the extra `orgId` field is simply dropped.
- **Suites:** `scripts/__tests__/tenant-export-columns.test.ts`, `tenant-export.test.ts` and `tenant-import.test.ts` pass **127/127**, run against the pg17 DB at 0296. I ran them from source via a throwaway vitest alias config (no db/lib dist build; the file was deleted afterwards).
- **The risk is future-only,** exactly as the reason says: once any lane writes `orgId`, a tenant migration silently turns org drives into personal drives. That's the X-3 gate.

## Tests run by me
- `bun run --filter @pagespace/db test -- src/schema/__tests__/organizations.test.ts`: **16 passed (16)**
- `test:integration -- organizations` (fresh pg17, TZ America/Chicago): **11 passed (11)**
- `packages/lib` `gdpr-export-coverage.test.ts` (from source): **8 passed (8)**
- tenant export column, export and import suites: **127 passed (127)**
- **Mutation 1 (source, line index 60 of `organizations.ts`):** dropped `.where(role = 'OWNER')` from `oneOwnerKey`. "ORG-1 an org has exactly one Owner membership row" went red (1 failed / 15 passed). Restored: 16/16.
- **Mutation 2 (DB):** replaced the open-invite key with `("orgId", email)`, i.e. no `lower()`. Integration › "ORG-3 an open invite for A@X and a@x in one org cannot coexist" went red (1 failed / 10 passed). Restored: 11/11.
- `git status` is clean apart from my untracked scratch dir, removed after posting. Postgres stopped.

CI at review time: `Static Security Analysis` **fail**; Unit Tests, Lint & TypeScript Check, E2E, Security Test Suite and CodeQL still pending.

**Summary: P2-2 and P2-3 FIXED. P2-1 NOT FIXED (unaddressed). 2 new P2s (inline), both about the exclusion entries.**

packages/lib/src/compliance/export/gdpr-export-coverage.ts
```diff
@@ -412,6 +412,15 @@ export const EXCLUDED_TABLES: Readonly<Record<string, string>> = {
...
+  org_members:
+    'No row exists to disclose: org membership ships dark behind ORGS_ENABLED and no service writes this table yet. The collector for the subject\'s own memberships is Spec X-2 (Organizations & Wallets lane G2), which must land before the flag turns on.',
+  org_invitations:
+    'An org invitation naming the inviter and the invited address; the counterparty is another person — the same Art 15(4) boundary as pending_invites — and an accepted invitation becomes the org membership row, disclosed through the org_members collector (Spec X-2).',
```
2witstudios: **P2-5: `org_invitations` is excluded, not exported, and the reason points at a collector that doesn't exist.** The round-1 summary records it as exported, but it's registered in `EXCLUDED_TABLES`. Its reason is written as permanent (the Art 15(4) boundary, like `pending_invites`) and doesn't name X-2 as temporary.

The reason's justification is that "an accepted invitation becomes the org membership row, disclosed through the org_members collector (Spec X-2)". But `org_members` is itself excluded two entries above, and that collector doesn't exist. Right now a GDPR export carries **no** org data at all.

Either mark this entry temporary on the same X-2 gate as the other two, or confirm the permanent exclusion on purpose and reword it so it doesn't depend on an unbuilt collector. Also correct the recorded status on the board.

packages/lib/src/compliance/export/gdpr-export-coverage.ts
2witstudios: Fixed in 6c12ba9e9. The `organizations` and `org_members` reasons in `gdpr-export-coverage.ts` and the `drives.orgId` reason in `scripts/lib/tenant-export-columns.ts` now name the real gate: temporary under Spec X-2 (X-3 for orgId), removed by the Phase 6 GDPR/backups leaf yfmlkdchehmberwthwu6g7vt before Wave F. The lane that introduces the ORGS_ENABLED code constant adds a test that fails while these exclusions exist; that test is a requirement on the B3 leaf. Registry re-checked against the worktree schema source: no undecided or phantom tables or columns. Leaving this open.

packages/lib/src/compliance/export/gdpr-export-coverage.ts
2witstudios: Fixed in 6c12ba9e9. `org_invitations` stays EXCLUDED permanently, on the same Art 15(4) counterparty basis as `pending_invites`. The reason no longer depends on an org_members collector: it names the counterparty boundary and says the row holds nothing of the subject's own beyond an address and an offered role. You were right that my round-1 channel summary could be read as saying it's exported; I've posted a correction in #org-wallets. Leaving this open.

packages/db/src/schema/organizations.ts
```diff
@@ -0,0 +1,110 @@
...
+  role: orgRole('role').default('MEMBER').notNull(),
+  invitedBy: text('invitedBy').references(() => users.id, { onDelete: 'set null' }),
+  joinedAt: timestamp('joinedAt', { mode: 'date' }).default(utcNow).notNull(),
+}, (table) => ({
+  // Also serves orgId lookups (leading column), so no separate orgId index.
+  orgUserKey: unique('org_members_org_user_key').on(table.orgId, table.userId),
+  // ORG-1: an org has exactly one human Owner. At most one OWNER row per org; ownership
+  // transfer demotes the old row and promotes the new one in one transaction, keeping it
+  // equal to organizations.ownerId.
+  oneOwnerKey: uniqueIndex('org_members_one_owner_key').on(table.orgId).where(sql`${table.role} = 'OWNER'`),
```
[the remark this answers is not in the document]
2witstudios: Tracked to B3 leaf t718n05yq2u9ukbil3i6hz5w: org creation and ownership transfer write organizations.ownerId and the OWNER org_members row in one transaction, with a test that they never disagree; the schema caps OWNER rows at one.

packages/db/src/schema/core.ts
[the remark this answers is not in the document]
2witstudios: Tracked to B6 leaf h40ht4zm4o2md5d5fktuhz6c (reassign before the user delete, test ORG-6); merge gate enforces that B5, the first writer of drives.orgId, does not merge before B6 (landing order on B5 leaf gj6frj67pnk7ek4kh6zyrzf4). No org drive can exist on integration until then.

2witstudios: Update: tracked to B6 leaf h40ht4zm4o2md5d5fktuhz6c (reassign before the user delete, test ORG-6); merge gate enforces that B5, the first writer of drives.orgId, does not merge before B6 (landing order on B5 leaf gj6frj67pnk7ek4kh6zyrzf4). No org drive can exist on integration until then.

2witstudios:
@codex review

2witstudios (commented):
[independent-review] agent:ow-irv-2659 — re-review round 2 at `6c12ba9e9` (delta vs pullrequestreview-5230240628)

This covers only the round-2 delta: 3 files, +10/−9 (`organizations.test.ts`, `gdpr-export-coverage.ts`, `tenant-export-columns.ts`). No schema or migration change.

## FIXED / NOT FIXED
| Finding | Status | Evidence |
|---|---|---|
| **P2-4**: exclusion reasons were gated on a nonexistent `ORGS_ENABLED` | **FIXED** | All three temporary reasons (`organizations` L420 and `org_members` L422 name X-2; tenant `drives.orgId` L129 names X-3) now say *removed by the Phase 6 GDPR/backups leaf yfmlkdchehmberwthwu6g7vt before Wave F*. That leaf exists and has a "Deferred from B1 … blocks Wave F" section with a grep gate. The B3 leaf t718n05yq2u9ukbil3i6hz5w now requires `ORGS_ENABLED` to be a code constant, plus a test titled "X-2 ORGS_ENABLED cannot be true while org tables are excluded from GDPR export". The reasons' forward reference is backed by a leaf requirement. |
| **P2-5**: `org_invitations` reason depended on a nonexistent `org_members` collector | **FIXED** | The reason no longer mentions that collector. It is now a permanent Art 15(4) counterparty exclusion, matching `pending_invites`, and names no X-2/X-3, so the Phase 6 grep gate won't catch it by mistake. Excluding it permanently is a stated decision now, not an accident. |
| db typecheck | **FIXED** | `bun run --filter @pagespace/db typecheck` **exits 0** at `6c12ba9e9`. Control: with `organizations.test.ts` at `bfdef9305` it exits 2 with `organizations.test.ts(35,89): error TS2345: Argument of type 'Partial<SQL<unknown> \| IndexedColumn>' is not assignable to parameter of type 'SQL<unknown>'`. The `is(c, SQL)` narrowing fixes exactly that. |
| **P2-1**: lead account delete cascades the org drive | **Tracked (per ruling), not fixed in B1** | Thread reply: "Tracked to B6 leaf h40ht4zm4o2md5d5fktuhz6c … merge gate enforces that B5, the first writer of drives.orgId, does not merge before B6 (landing order on B5 leaf gj6frj67pnk7ek4kh6zyrzf4)". Both leaf pages carry it: B6 has a "drives.ownerId still ON DELETE CASCADE … must reassign BEFORE" item, and B5 has a "Landing order … does not merge before … h40ht4zm4o2md5d5fktuhz6c" section. |
| Codex P1: owner membership must match `ownerId` | **Tracked (per ruling)** | Thread reply names B3 leaf t718n05yq2u9ukbil3i6hz5w, and that page carries "organizations.ownerId and the OWNER org_members row must agree … a test proves they never diverge". |
| Codex P1: account erasure deletes org drives | **Tracked (per ruling)** | Thread reply names B6 plus the B6-before-B5 landing order, same as P2-1. |

All tracked threads are left open, which is correct.

## Tests run by me at `6c12ba9e9`
- `bun run --filter @pagespace/db typecheck`: exit 0 (red at `bfdef9305`, see above)
- `bun run --filter @pagespace/db test -- src/schema/__tests__/organizations.test.ts`: **16 passed (16)**. The `lower("org_invitations"."email")` column assertion still passes through the new `is(c, SQL)` branch.
- `packages/lib` `gdpr-export-coverage.test.ts`, run from schema source through a throwaway alias config that I deleted afterwards: **8 passed (8)**
- `git status` clean afterwards.

## P3 (board only, not code)
The Phase 6 leaf yfmlkdchehmberwthwu6g7vt still says "(org_invitations is exported like pending_invites)". Both halves are wrong: `org_invitations` and `pending_invites` are both in `EXCLUDED_TABLES`. That leaf's agent should know `org_invitations` is a permanent exclusion, not something it has to keep exporting. Please correct the page.

**Summary: P2-4, P2-5 and the typecheck are FIXED. P2-1 and both Codex P1s are tracked to B3/B6 with the B6-before-B5 landing order, as ruled. No new P1/P2.**

2witstudios (commented):
[independent-review] agent:ow-irv-2659 — delta review `6c12ba9e9..6cd0d9f05` (web fixture typecheck fix)

Codex quota is exhausted, so this review is the gate on this head.

| Check | Verdict | Evidence |
|---|---|---|
| Test fixtures only | ✅ | 11 files, **+23/−0**, every path under `apps/web/src/app/api/**/__tests__/`. `git diff --name-only 6c12ba9e9..6cd0d9f05 \| grep -v /__tests__/` is empty. No source, schema or migration change. |
| No behavior added | ✅ | Every added line is a literal field on a mock object factory (`mockDrive`/`createMockDrive`-style helpers and one `mockDriveMember`). No assertions, mocks or control flow changed, and nothing was removed. |
| Real schema defaults, nothing that masks a bug | ✅ | Unique added lines across the diff: `orgId: null` ×11, `orgVisibility: 'OPEN' as const` ×11, `source: 'invite' as const` ×1. These match the DB defaults verified on pg17 in earlier rounds (`orgId` nullable, no default; `'OPEN'::"OrgDriveVisibility"`; `'invite'::"DriveMemberSource"`). Every mock drive is personal and every mock member is a direct invite, which is exactly the shape of every production row today (nothing writes `orgId`), so these fixtures can't hide an org-path bug. |
| CI `Lint & TypeScript Check` on `6cd0d9f05` | ✅ **success** | Run 35174434223 (headSha `6cd0d9f05`): steps `Run ESLint`, `Run TypeScript check`, `Run ESLint (scripts)` and `Run TypeScript check (scripts)` all succeeded. |
| CI `Static Security Analysis` typecheck on `6cd0d9f05` | ✅ **success** | Run 35174434058 (headSha `6cd0d9f05`): `TypeScript type checking` and `ESLint security rules` succeeded. It was failing on `1ba6886b1` and `bfdef9305`. |

Other checks on this head: Security Test Suite, CodeQL Security Analysis, Dependency Audit and Secret Scanning pass. `ci / Unit Tests` and `ci / E2E` were still pending when I posted, and this review doesn't cover them.

Earlier rounds are unchanged: P2-2, P2-3, P2-4, P2-5 and the db typecheck are FIXED. P2-1 and both Codex P1s are tracked to B3/B6 with the B6-before-B5 landing order.

**No P1/P2 on this delta.**

2witstudios:
@codex review

2witstudios (commented):
[independent-review] agent:ow-irv-2659 — review `6cd0d9f05..369a04275` (integration merge + `(partial)` markers)

| # | Check | Verdict | Evidence |
|---|---|---|---|
| 1 | Merge's allowlist conflict resolution | ✅ | `git show --remerge-diff 369a04275`: the one conflict was `scripts/spec-coverage-allowlist.txt`. The resolution removes the markers, keeps #2657's MON-3 block verbatim, and adds commented DRV-1, DRV-8 and ORG-1/2/3 blocks citing reviews 5230157358 and 5230240628. ID sets: parents have 70 (`6cd0d9f05`) and 79 (`f4766eef6`) entries, the merge has 78. **Union minus merge = {DRV-4} only**, the intended removal because DRV-4 is now claimed. Merge minus union = ∅. SEAT-2, WAL-5 and MON-3 from #2657 are all present, and no conflict markers remain. |
| 2 | Every `(partial)` well-formed; no plain ORG/DRV claim except DRV-4 | ✅ | Only 2 test files in the repo name an ORG/DRV ID: `organizations.test.ts` and `organizations.integration.test.ts`. All 25 ID-bearing titles checked with the gate's own rules (whole-token ID, one ASCII space, lowercase `(partial)`): every ORG-1/2/3/6 and DRV-1/8 occurrence, including the mid-title `— ORG-6 (partial)` ones and the triple `ORG-1 (partial) ORG-2 (partial) ORG-3 (partial)`, is well-formed. **Only plain claims: DRV-4** (`DRV-4 org visibility is Open, Restricted or Private and defaults to Open`, `DRV-4 defaults a drive to Open visibility`). No multi-line titles and no ID mentions outside titles. CI's gate agrees: no malformed-marker failure. |
| 3 | DRV-4 test proves both halves | ✅ | **Enum:** the unit test asserts `ORG_DRIVE_VISIBILITIES` and `orgDriveVisibility.enumValues` = `['OPEN','RESTRICTED','PRIVATE']`, plus NOT NULL. **Default Open:** the unit test asserts `.default === 'OPEN'`. The integration test inserts an org drive (`orgId` set) with no visibility and reads back `OPEN` from Postgres, and an explicit `PRIVATE` round-trips. **Mutation (line index 33 of `core.ts`):** `.default('OPEN')` → `.default('PRIVATE')` turns `DRV-4 org visibility is Open, Restricted or Private and defaults to Open` red (1 failed / 15 passed). Restored: 16/16, `git status` clean. (P3: no test inserts `RESTRICTED` against the DB; that's covered by the generated enum DDL.) |
| 4 | 0296 unchanged and still generated | ✅ | `git diff --stat 6cd0d9f05 369a04275 -- packages/db/drizzle packages/db/src/schema ':!…/__tests__'` is **empty**. Latest migrations are still `0295_groovy_risque` then `0296_charming_dazzler`. `drizzle-kit generate` on the merged tree reports "No schema changes, nothing to migrate". |
| 5 | CI spec-coverage on `369a04275` | ✅ | `ci / Spec ID coverage` job 105081479067 (run 35181696773, headSha `369a042755e3867ef6ea18f5f5de10454969c8aa`): **`covered 9/87, allowlisted-missing 78, MISSING 0` / `spec-coverage: OK`**. Covered = DRV-4 + MON-1, MON-2, MON-4, MON-5, MON-6, MON-7, MON-8 + UI-12. ORG-1/2/3/6 and DRV-1/8 show `allowlisted  partial: <both org test files>`, so there's no STALE-ALLOW. |

All other checks on this head pass: Unit Tests (33m55s), E2E, Lint & TypeScript Check, Static Security Analysis, Security Test Suite, CodeQL, Dependency Audit and Secret Scanning.

Carried from earlier rounds: P2-2, P2-3, P2-4 and P2-5 are FIXED. P2-1 and both Codex P1s are tracked to B3/B6 with B6 landing before B5.

**No P1/P2 on this delta.**

packages/db/src/schema.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED at 369a04275 by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0): all three org tables carry a recorded GDPR export decision; organizations and org_members are temporary X-2 exclusions whose reasons name leaf yfmlkdchehmberwthwu6g7vt before Wave F, and org_invitations is a permanent Art 15(4) exclusion like pending_invites. Resolving.

packages/db/src/schema/core.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED at 369a04275 by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0): tenant export carries drives.orgVisibility and drive_members.source; drives.orgId is a temporary X-3 exclusion, safe today because nothing writes orgId, and removed by leaf yfmlkdchehmberwthwu6g7vt before Wave F. Resolving.

packages/db/src/schema/organizations.ts
2witstudios: [point-guard] Verified FIXED at 369a04275 by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0): org_members_one_owner_key rejects a second OWNER insert or promotion on pg17. ownerId and OWNER-row agreement is a B3 leaf requirement. Resolving.

packages/db/src/schema/organizations.ts
2witstudios: [point-guard] Verified FIXED at 369a04275 by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0): the open-invite key is (orgId, lower(email)) and A@X beside a@x fails on pg17. Expired-invite rotation and live-invite seat counting are B3 leaf requirements. Resolving.

packages/lib/src/compliance/export/gdpr-export-coverage.ts
2witstudios: [point-guard] Verified FIXED at 369a04275 by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0): exclusion reasons now name the real gate (leaf yfmlkdchehmberwthwu6g7vt before Wave F, plus the ORGS_ENABLED code-constant test required on B3 leaf t718n05yq2u9ukbil3i6hz5w). Resolving.

packages/lib/src/compliance/export/gdpr-export-coverage.ts
2witstudios: [point-guard] Verified FIXED at 369a04275 by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0): org_invitations is a permanent Art 15(4) exclusion with no dependency on an unbuilt collector. Resolving.

packages/db/src/schema/core.ts
2witstudios: [point-guard] Verified tracked, not a B1 defect: no code on integration writes drives.orgId or creates an org, so the failure cannot occur yet. The requirement and its test title are written on the owning leaf, and the Wave B landing order makes B6 (reassign before user delete) merge before B5 (first writer of drives.orgId); B3 owns ownerId and OWNER-row agreement. Confirmed by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0). Resolving.

packages/db/src/schema/organizations.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified tracked, not a B1 defect: no code on integration writes drives.orgId or creates an org, so the failure cannot occur yet. The requirement and its test title are written on the owning leaf, and the Wave B landing order makes B6 (reassign before user delete) merge before B5 (first writer of drives.orgId); B3 owns ownerId and OWNER-row agreement. Confirmed by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0). Resolving.

packages/db/src/schema/core.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified tracked, not a B1 defect: no code on integration writes drives.orgId or creates an org, so the failure cannot occur yet. The requirement and its test title are written on the owning leaf, and the Wave B landing order makes B6 (reassign before user delete) merge before B5 (first writer of drives.orgId); B3 owns ownerId and OWNER-row agreement. Confirmed by independent re-reviews 5230240628, 5230294432 and 5231245565 (final: no P1/P2) and CI 11/11 on 369a04275 (spec-coverage 9/87, MISSING 0). Resolving.

PR #2662, opened 2026-09-17
feat(billing): pure wallet core (ow-c2) — spend resolution, allocation, overshoot, caps

2witstudios:
Lane **ow-c2** (Wave B) — leaf `m8yt4d2oi174xnevzfi4pslq`: the pure wallet core.

## What
`packages/lib/src/billing/wallet-core.ts` — no schema, no IO, no db import (purity test enforces imports are only `./money-model` and `./credit-core`, no `process.env`, no `Date.now`, no `new Date()`). Credit counts enter only via money-model `centsFromCredits`; no second conversion (MON-5 seam guard green). Reuses credit-core `allocateSpend` and `evaluateDailyCap`. New `@pagespace/lib/billing/wallet-core` subpath export + knip entry.

Contracts for C1 / C3 / C4:
- `resolveSpendSource({ actor, driveWallet, seatAllowance, personal, chosen, driveRule, userOverride, reservationCents })` → `{kind:'spend', source, walletId, fallbackApplied, fallbackFrom}` | `{kind:'refuse', source, reason, options, chargeCents: 0}` | `{kind:'skip', reason, walletId, chargeCents: 0}`
- `effectiveSpendPolicy(org, drive)` (POL-7)
- `childSpendableCents(funds, parentAvailableCents)`, `legSpendableCents(walletSpendable, capRemaining)`, `allocateWalletSpend({ parent, wallet, amountCents })` (WAL-3, D-OW-13 per-leg draws)
- `settleOvershoot({ source, overshootCents, chargedWalletId, parentWalletId, funderChoice, fallbackFrom: { source, walletId, parentWalletId, funderChoice } | null })`, `walletStatusFor`, `shouldNotifyFunderOfDebt`, `renewWalletAllocation` (WAL-6)
- `evaluateCaps({ caps, usage: { dailySpentCents, monthlySpentCents, dailyReservedCents, monthlyReservedCents }, reservationCents })`, `capAlertThresholdsCrossed`, `DEFAULT_CONSUMER_CAPS` (WAL-7)
- `utcDayStartMs`, `utcMonthStartMs`, `governingAllocationPeriodStartMs`, `isAllocationResetDue` (D-OW-12)
- `entitlementTierFor({ source, walletOwnerTier, consumerTier })` (D-OW-14)

## Semantics worth reviewing
- Automations: drive wallet only; uncovered/paused/missing ⇒ `skip`, never a person, whatever the fallback rule.
- "Always my own credits" (global or per-drive) is absolute: own credits or refuse with `options: []`, no fallback.
- Guests (D-OW-4): drive wallet only with the per-drive switch; never a seat allowance, never as a fallback.
- POL-7 "a drive may only be stricter": a drive fallback applies only when it equals the org rule or is `refuse`; any other drive rule resolves to `refuse`. Confirmed by the point-guard ruling on the leaf (seat allowance and own credits are not ordered); tests carry the ruling's three titles.
- Caps count settled spend PLUS this consumer's outstanding holds in the window (Codex P1, d3534c6be).
- Fails closed: a non-finite cap is 0 and a non-finite reservation is refused under any set cap; spendable math rounds to whole cents (`centsFromCredits(7)` is `7.000000000000001`; the money-model fix is a separate lane, ow-mm-float) (b1bcf92d7).
- WAL-6b on the fallback path: a drive-rule fallback onto own credits is not the consumer choosing them, so overshoot lands where the CHOSEN source would have put it, never on the consumer (independent review 5232144738 P2).
- WAL-6d renewal nets wallet debt from the next allocation and clears any excess (mirrors `computeMonthlyRefill` forgiveness).

## ID / test table
All tests in `packages/lib/src/billing/__tests__/wallet-core.test.ts`. Every Spec-ID title is `(partial)`; no ID is removed from `scripts/spec-coverage-allowlist.txt` (comments added naming what is unproven).

| ID | Tests (title prefix) | Status |
|---|---|---|
| SPEND-1 | `SPEND-1 (partial)` — one source per chosen kind; no choice refuses with options; personal drive offers no seat | partial (gate wiring C3) |
| SPEND-4 | `SPEND-4 (partial)` — 6-row empty/paused/over/unavailable refusal table, charge 0; fallback shown (`fallbackApplied`); fallback empty/paused/missing refuses; covered never falls back; self-fallback refuses | partial (chip/strip UI-8) |
| SPEND-5 | `SPEND-5 (partial)` — both switches override every choice; absolute refusal with no fallback | partial (storage) |
| SPEND-6 | `SPEND-6 (partial)` — automation spends drive wallet ignoring override; empty/short/paused/missing × every fallback rule ⇒ skip | partial (C4 wiring, skip log) |
| POL-7 | `POL-7 (partial) a drive rule equal to the org rule applies` / `… a drive rule of refuse applies` / `… a drive rule that swaps seat allowance and own credits resolves to refuse`, plus seat-allowance clamping rows | partial (E1 store, C3 read) |
| WAL-3 | `WAL-3 (partial)` — spendable table; allocation moves nothing until spend; draw order + per-leg tracking; top-up survives renewal; parent-capped draw | partial (C1/C3) |
| WAL-6 | `WAL-6 (partial)` — landing per funder choice; never on consumer, including after a drive-rule fallback onto own credits; status; once-per-period notice; renewal netting | partial (hold/settle, notification) |
| WAL-7 | `WAL-7 (partial)` — defaults 10/100 credits; caps table incl. in-flight reservations; fails closed on non-finite cap/reservation; 80/100% crossings; UTC windows; leg spendable bounded by caps | partial (stored caps, alerts, kill switch control) |
| D-OW-4/12/14 | `D-OW-4 …` (incl. a guest with the switch off and empty own credits is never offered a funded drive wallet, for every fallback rule), `D-OW-12 …`, `D-OW-14 …` | decision behaviour |

```
$ bun run --filter @pagespace/lib test -- src/billing/__tests__/wallet-core.test.ts
 Test Files  1 passed (1)
      Tests  119 passed (119)
$ (packages/lib) bun run test -- src/__tests__/seams/credit-conversion.seam.test.ts   → Tests 28 passed (28)
$ (packages/lib) bun run test -- src/billing/__tests__/money-model-guard.test.ts        → Tests 3 passed (3)
$ (packages/lib) bunx tsc --noEmit --strict … wallet-core.ts wallet-core.test.ts       → exit 0
```
Monorepo typecheck/lint/coverage not run locally (build-slot rule); CI is the gate.

## Mutation check (by line index)
1. `wallet-core.ts:195` `if (!overridden && fallback !== 'refuse' && fallback !== chosen) {` → `if (false) {` — red: 2 failed | 99 passed (`SPEND-4 (partial) a drive rule allowing seat_allowance/own_credits falls back to it and says so`). Restored, 101/101.
2. Same line, dropped `!overridden &&` — red: 1 failed | 100 passed (`SPEND-5 (partial) the override is absolute …`). Restored.
3. `dailyUsed` line, dropped `+ wholeNonNegative(input.usage.dailyReservedCents)` — red: 2 failed | 106 passed (in-flight reservation rows). Restored.
4. Non-finite-reservation daily refusal line blanked — 1 failed | 113 passed; monthly refusal line blanked — 1 failed | 113 passed; `sanitizeCap` returning the raw cap — 2 failed | 112 passed. Each restored.
5. `wallet-core.ts:124` guest drive-wallet switch → `if (false)` — 5 failed | 114 passed (all 3 rows of the D-OW-4 options test). Restored.
6. `wallet-core.ts:403` fallback-overshoot condition → `if (false && …)` — 3 failed | 116 passed (WAL-6 fallback rows). Restored; 119/119; `git status` clean of mutations.

https://claude.ai/code/session_01A8ptMEGSSfibEZZqWT6LeZ


2witstudios:
@codex review

packages/lib/src/billing/__tests__/wallet-core.test.ts
```diff
@@ -0,0 +1,594 @@
...
+          actor: chris,
+          chosen: 'own_credits',
+          personal,
+          driveRule: { fallback: 'seat_allowance', guestsMaySpendDriveWallet: false },
+        }),
+      ),
+    ).toMatchObject({ kind: 'refuse', source: 'own_credits', reason: 'source_empty', chargeCents: 0 });
+  });
+
+  it('D-OW-4 a guest with the switch off never falls back onto the drive wallet', () => {
```
2witstudios: **P2 — vacuous test (test honesty).** `D-OW-4 a guest with the switch off never falls back onto the drive wallet` cannot fail. `FallbackRule` is `'refuse' | 'seat_allowance' | 'own_credits'`, so the drive wallet can never be a fallback. With `fallback: 'own_credits'` equal to `chosen: 'own_credits'`, the call refuses through the `fallback !== chosen` guard before any guest logic runs. The assertion is `toMatchObject({ kind: 'refuse', chargeCents: 0 })` and does not check `options`.

**Evidence:** I mutated wallet-core.ts:124 (the guest drive-wallet switch check) to `if (false)`. This test stayed green, and only the other two D-OW-4 tests went red. The behaviour this test *should* guard is still unguarded for the options path: a guest with the switch off, whose chosen own credits are empty, must not be *offered* the drive wallet. Nothing asserts that.

**Fix:** assert `options` in this case: `toEqual({ kind: 'refuse', source: 'own_credits', reason: 'source_empty', options: [], chargeCents: 0 })` with a funded drive wallet. Or retitle the test to what it proves. Then confirm red under the :124 mutation.

packages/lib/src/billing/wallet-core.ts
```diff
@@ -0,0 +1,543 @@
...
+
+/**
+ * Where actual-cost overshoot beyond a covered reservation lands (WAL-6b/c). It lands on
+ * the consumer only when they spent their own credits; a seat's overshoot lands on the
+ * pool; a drive wallet's lands where its funder chose.
+ */
+export function settleOvershoot(input: SettleOvershootInput): OvershootLanding {
+  const cents = wholeNonNegative(input.overshootCents);
+  if (cents === 0) return { kind: 'none', cents: 0 };
+  if (input.source !== 'drive_wallet') return { kind: 'root_debt', walletId: input.chargedWalletId, cents };
```
2witstudios: **P2 — WAL-6(b) on the fallback path.** The Spec says: "(b) Overshoot never lands on the consumer **unless they chose their own credits**." `resolveSpendSource` can move a consumer who chose `drive_wallet` or `seat_allowance` onto `own_credits` by drive rule (`fallbackApplied: true`). `settleOvershoot` then sees only `source: 'own_credits'` and returns `root_debt` on the consumer's personal wallet. So an overshoot lands on a person who did not choose their own credits. The input has no `fallbackApplied` or `fallbackFrom`, so the core cannot distinguish the two cases.

There is also no test for this case. Every `own_credits` row in `settleOvershoot` is a chosen one.

The POL-7 ruling on the leaf points the same way ("A drive can never turn org-funded fallback into automatic charges on a person's own credits"). Whether a drive-rule fallback counts as "chose" is arguably a reading question. If the lane believes it does, raise it with the orchestrator rather than deciding here. Otherwise, take `fallbackFrom` as input and land the overshoot where the originally chosen source would have put it. Add a `WAL-6 (partial)` row for the fallback case.

2witstudios (commented):
[independent-review] agent:ow-irv-2662

Independent spec review of #2662 (lane ow-c2, pure wallet core) on head `5a1b8d57a`. Read: Spec drc7x34unhc0ty1dc0u1j3gy, leaf m8yt4d2oi174xnevzfi4pslq including the POL-7 point-guard ruling, Sequence Spec gates 0 to 0e, Retro 0, and the D-OW-4/12/13/14 decisions. **0 P1, 2 P2** (inline). The core is clean: pure, one credit definition, and the refusal, fallback, override and automation rules match the Spec. The P2s are one vacuous test and one WAL-6(b) clause the code does not honour on the fallback path.

## Per-ID verdicts

Every Spec-ID title carries `(partial)` with exactly one ASCII space, in lowercase. No plain ID claim. No ID removed from `scripts/spec-coverage-allowlist.txt` (comments only), so no ID has to be fully proven.

| ID | Verdict | Evidence (packages/lib/src/billing/__tests__/wallet-core.test.ts) |
|---|---|---|
| SPEND-1 | partial, correctly marked | `SPEND-1 (partial) a member who chose the %s spends exactly that one source`; `… no chosen source refuses and offers the covered options, charging nothing`; `… a personal drive has no seat allowance, so it is never offered`. The gate wiring for every AI call is unproven (C3). |
| SPEND-4 | partial, correctly marked | `SPEND-4 (partial) with no fallback rule, when the %s, the call refuses, names the source, offers the rest, and charges zero` (6 rows); `… a drive rule allowing %s falls back to it and says so` (`fallbackApplied: true`, `fallbackFrom`); `… fallback still refuses with zero charge when %s`; `… a covered chosen source never falls back`. The chip and strip are unproven (UI-8). |
| SPEND-5 | partial, correctly marked | `SPEND-5 (partial) %s spends own credits whatever was chosen` (both switches × all 4 choices); `… the override is absolute: empty own credits refuse with zero charge and never fall back to the drive` (`options: []`). Storage is unproven. |
| SPEND-6 | partial, correctly marked | `SPEND-6 (partial) an automation spends the drive wallet, ignoring any chosen source or override`; `… an automation whose drive wallet is %s skips and never falls back to a person` (empty/short/paused/missing × all 3 fallback rules; a person's legs are funded in `base`, so the negative is real). The skip log and C4 wiring are unproven. |
| POL-7 | partial, correctly marked; matches the ruling | `POL-7 (partial) %s` (6 rows), `… a drive cannot loosen an org that refuses`, `… an unlimited org seat allowance takes the drive amount`. The code at wallet-core.ts:229-231 matches the ruling exactly: a drive rule applies only if it equals the org's or is `refuse`, and anything else resolves to `refuse`. Test titles differ from the three the ruling names, but they cover the same behaviour. Gap (not a P-level): only one swap direction is tested (org `seat_allowance`, drive `own_credits`). The reverse (org `own_credits`, drive `seat_allowance` → `refuse`) is untested; the code is a single equality, so it is symmetric. |
| WAL-3 | partial, correctly marked | `WAL-3 (partial) childSpendableCents: %s` (7 rows); `… setting an allocation moves nothing out of the parent`; `… allocation draws first, then top-up legs in the order given, each leg tracked separately (D-OW-13)`; `… a top-up lasts until spent`; `… the allocation draw is capped by what the parent can cover`. |
| WAL-6 | partial, correctly marked; see P2 on (b) | `WAL-6 (partial) overshoot lands per funder choice: %s` (4 rows), `… a drive-wallet overshoot never lands on the consumer`, `… walletStatusFor`, `… notified once per period`, `… renewWalletAllocation`. (a) hold/settle and (d) parent forgiveness are not in this PR, and the allowlist comment says so. |
| WAL-7 | partial, correctly marked | `WAL-7 (partial) defaults on enable are 10 credits a day and 100 a month, from the one credit definition`; the caps table; the alert thresholds (80/100, exact landing, both at once); `… cap windows are UTC` (23:59:59.999, Jan 1 00:00, 2024-02-29); `legSpendableCents`. Month-end last-millisecond rows (for example Dec 31 23:59:59.999 → Dec 1, and Feb 29 23:59:59.999) are not tabled; the implementation is `Date.UTC` and correct by inspection. |
| D-OW-4 | behaviour correct; one vacuous test (P2) | `D-OW-4 a guest cannot spend the drive wallet by default…`, `… the per-drive switch lets a guest spend…`, `… a guest holds no seat…` are real (mutations below). |
| D-OW-12 | satisfied as pure selection | `D-OW-12 %s` (org → pool refill, user → personal renewal, UTC-month fallback), `D-OW-12 isAllocationResetDue: %s`. |
| D-OW-13 | satisfied as pure math | Donation legs carry `funder: 'donation'` + `donorUserId` and are drawn and tracked per leg (`legDraws`). Non-refundability is a C5/UI concern. |
| D-OW-14 | satisfied as pure selection | `D-OW-14 %s`: the owner tier for drive wallet and seat, the consumer tier only for own credits. The donor tier is never an input. |

## Focus checklist
- **(a) Pure:** confirmed. The only imports are `./money-model` and `./credit-core`. No `process.env`, no `Date.now`, no argless `new Date()` (only `new Date(nowMs)` from an injected clock). The purity test enforces this. Its import regex would not catch a dynamic `import()` or a `require(`, but there are none.
- **(b) One credit definition:** confirmed. The only credit entry is `centsFromCredits(10|100)`. `PERCENT = 100` at :480 is percent-of-cap math, not a conversion. **Probe:** I appended `cents / 100` and then `Math.round(credits * 100)` helpers to wallet-core.ts. Both turned `MON-5 no file in lib, web, marketing, admin, or cli divides or multiplies a cents/credit value by 100` red (1 failed | 30 passed). Restored.
- **(c) SPEND-4:** confirmed, with a note. A chosen source that is *unavailable* (for example a drive with no wallet) refuses without consulting the fallback rule (:189). That is conservative and never over-charges, but the Spec's fallback clause is written for "empty". Flagged for C3, not a P-level.
- **(d) SPEND-5:** absolute. The override forces `own_credits`, suppresses fallback (`!overridden`, :195) and offers `[]`.
- **(e) Automations:** skip on empty, short, paused or missing whatever the fallback rule is, and never touch `personal` or `seatAllowance`.
- **(f) D-OW-4:** enforced in `legFor` for choice, fallback and options. Guests never get a seat.
- **(g)** WAL-3 / WAL-6 / WAL-7 / D-OW-12 / D-OW-13 / D-OW-14: see the table. Notes that are not P-level:
  - `childSpendableCents` nets wallet debt against top-up and donation legs too (:282), so a debt temporarily locks donor funds until renewal. WAL-6(c) says "netted from the next allocation".
  - `settleOvershoot` with `absorb_to_parent` and `parentWalletId === null` silently becomes `wallet_debt` (:387-390). WAL-2 says a drive wallet always has a parent, so this is unreachable with valid data.
- **(h) POL-7:** matches the ruling (above).
- **(i) Integer cents:** there is no float money arithmetic. Rounding note (not a P-level; inputs are whole cents by contract, and this mirrors `credit-core.allocateSpend`):
  - `covers()` uses `Math.round(reservationCents)` (:113), so a 5.4-cent reservation is "covered" by 5 spendable, which under-reserves against WAL-6(a).
  - `allocateWalletSpend` rounds a 10.5-cent amount up to 11.
  - If C3 ever passes fractional values, the reservation should use `Math.ceil` and the charge should not round up.

## Test honesty: runs and mutations (line index, anchor asserted, restored, `git status` clean)
- Base: `bun run test -- src/billing/__tests__/wallet-core.test.ts src/__tests__/seams/credit-conversion.seam.test.ts src/billing/__tests__/money-model-guard.test.ts` → `Test Files 3 passed (3)`, `Tests 132 passed (132)`, exit 0.
- **M1** wallet-core.ts:129 (`if (isGuest) return { unavailable: 'source_unavailable' }` → removed): red `Tests 2 failed | 99 passed (101)`. The failures were `D-OW-4 a guest cannot spend the drive wallet by default and is offered own credits only` and `D-OW-4 a guest holds no seat…`.
- **M2** wallet-core.ts:231 (`driveFallback === undefined || driveFallback === org.fallback ? org.fallback : 'refuse'` → `driveFallback ?? org.fallback`): red `2 failed | 99 passed`. The failures were `POL-7 (partial) a drive cannot swap to a different fallback; it resolves to refuse` and `POL-7 (partial) a drive cannot loosen an org that refuses`.
- **M3** wallet-core.ts:124 (the guest drive-wallet switch check → `if (false)`): red `2 failed | 99 passed`. However, **`D-OW-4 a guest with the switch off never falls back onto the drive wallet` stayed green** with the guest switch removed entirely. See the P2.
- The lane's own mutations on :195 (fallback branch and `!overridden`) cover the leaf's "mutation check on the fallback branch".

## X-6
- "An empty chosen wallet charges nothing": tested (`chargeCents: 0` across the SPEND-4 table and the fallback-refuses rows).
- "An automation never charges a person": tested (SPEND-6 skip table with funded person legs).
- The other X-6 negatives are not in this leaf.

## Rules and scope
- No `any`. No env flags. No SQL or bulk inserts. The new subpath has a `package.json` exports entry and a typesVersions entry, plus a knip entry. No gate self-test.
- Scope is exactly the leaf's files plus exports, knip and allowlist comments. No migration.

## CI
Run 35184591515 on this head: Lint & TypeScript Check pass, Unit Tests pass, E2E pass, **Spec ID coverage pass** (job 105090589881). Log lines for this PR's IDs:
```
POL-7     allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
SPEND-1   allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
SPEND-4   allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
SPEND-5   allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
SPEND-6   allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
WAL-3     allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
WAL-6     allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
WAL-7     allowlisted  partial: packages/lib/src/billing/__tests__/wallet-core.test.ts
covered 9/87, allowlisted-missing 78, MISSING 0
spec-coverage: OK
```
The gate agrees with the per-ID table: all eight are partial and allowlisted, and none is claimed covered.

packages/lib/src/billing/wallet-core.ts
```diff
@@ -0,0 +1,543 @@
...
+/** Both UTC caps against this call's reservation; the daily cap is reported first. */
+export function evaluateCaps(input: { caps: ConsumerCaps; usage: CapUsage; reservationCents: number }): CapsResult {
+  const remaining: CapRemaining = {
+    dailyRemainingCents: capRemaining(input.caps.dailyCents, input.usage.dailySpentCents),
+    monthlyRemainingCents: capRemaining(input.caps.monthlyCents, input.usage.monthlySpentCents),
+  };
+  const daily = evaluateDailyCap({
+    dailyChargedCents: input.usage.dailySpentCents,
+    estCostCents: input.reservationCents,
+    capCents: input.caps.dailyCents,
```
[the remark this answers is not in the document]
2witstudios: Fixed in d3534c6be (head now b1bcf92d7). `CapUsage` gains required `dailyReservedCents` / `monthlyReservedCents` (this consumer's outstanding holds in the UTC day / month window); `evaluateCaps` adds them to settled spend for both the remaining figures and the allow/deny decision, so overlapping calls cannot each pass on the same settled total. Required fields, so the C3 gate cannot omit them.

Tests (`packages/lib/src/billing/__tests__/wallet-core.test.ts`): "WAL-7 (partial) an in-flight reservation counts against the daily cap", "... against the monthly cap", "reservations that still fit are allowed". Mutation by line index: dropping `+ wholeNonNegative(input.usage.dailyReservedCents)` from the `dailyUsed` line → 2 failed | 106 passed; restored → 108/108 (now 114/114 with b1bcf92d7).

Leaving this thread open for verification.

packages/lib/src/billing/__tests__/wallet-core.test.ts
2witstudios: Fixed on head df1d3326a. The test is now `D-OW-4 a guest with the switch off and empty own credits is never offered a funded drive wallet (fallback %s)`, one row per fallback rule (`refuse`, `seat_allowance`, `own_credits`), with a funded drive wallet (`c(1200)`), and it asserts the whole result: `toEqual({ kind: 'refuse', source: 'own_credits', reason: 'source_empty', options: [], chargeCents: 0 })`.

Mutation by line index, wallet-core.ts:124 `if (isGuest && !input.driveRule.guestsMaySpendDriveWallet)` → `if (false)`: 5 failed | 114 passed, including all 3 rows of this test. Restored → 119/119.

Leaving open for verification.

packages/lib/src/billing/wallet-core.ts
2witstudios: Fixed on head df1d3326a (read as you did: a drive-rule fallback is not "they chose their own credits"). `SettleOvershootInput` gains a required `fallbackFrom: { source, walletId, parentWalletId, funderChoice } | null` (the chosen source, from `resolveSpendSource`'s `fallbackFrom`). When the charged source is `own_credits` and `fallbackFrom` names another source, `settleOvershoot` lands the overshoot exactly where the chosen source would have: the drive wallet's funder choice (parent debt or wallet debt), or the pool for a seat allowance. It never lands on the consumer. Required rather than optional, so the C3 gate cannot drop it.

Tests: `WAL-6 (partial) after a drive-rule fallback onto own credits, overshoot lands where the chosen source would put it, never on the consumer: %s` (from drive wallet/absorb, drive wallet/wallet debt, seat allowance), each also asserting the landing wallet is not `w-marcus`. Mutation by line index, wallet-core.ts:403 condition → `if (false && …)`: 3 failed | 116 passed. Restored → 119/119.

Leaving open for verification.

2witstudios (commented):
[independent-review] agent:ow-irv-2662 — re-review 5a1b8d57a..df1d3326a

This re-review covers the range from my first review (pullrequestreview-5232144738) to `df1d3326a`. That range has two fix commits (`d3534c6be`, `b1bcf92d7`), one fix commit for my findings (`2c66217c1`), and a merge of the `pu/org-wallets` tip (`df1d3326a`). The merge-base with the base branch is the tip, and the merge changes none of this PR's files. **New P1: 0. New P2: 0.**

| Finding | Verdict | Evidence |
|---|---|---|
| P2 — vacuous test `D-OW-4 a guest with the switch off never falls back onto the drive wallet` (test.ts:261) | **FIXED** | Replaced by `D-OW-4 a guest with the switch off and empty own credits is never offered a funded drive wallet (fallback %s)` (test.ts:262). It covers all three fallback rules, with a funded drive wallet, asserted with `toEqual(... options: [] ...)`. **Mutation:** wallet-core.ts:124, the guest switch check, changed to `if (false)`. Result: `Tests 5 failed / 114 passed (119)`, and all 3 new rows are red. Before the fix, the old test stayed green under the same mutation. |
| P2 — WAL-6(b) overshoot lands on the consumer after a drive-rule fallback onto own credits (wallet-core.ts:386) | **FIXED** | `SettleOvershootInput.fallbackFrom: ChosenSourceCharge / null` is now **required**, so a caller cannot drop it (wallet-core.ts:377-381). If the call spent `own_credits` and the fallback came from another source, the overshoot is settled against the chosen source's wallet, parent and funder choice (wallet-core.ts:403-413). New test: `WAL-6 (partial) after a drive-rule fallback onto own credits, overshoot lands where the chosen source would put it, never on the consumer: %s` (test.ts:509). It has 3 rows: drive wallet absorbed into the parent, drive wallet as wallet debt, and seat allowance landing on the pool. **Mutation:** wallet-core.ts:403 changed to `if (false) {`. Result: `3 failed / 116 passed (119)`. |
| Codex P1 — caps ignored in-flight holds (`d3534c6be`) | FIXED (verified independently) | `CapUsage` now has `dailyReservedCents` and `monthlyReservedCents`, and both are added to settled spend before the cap check (wallet-core.ts:502-503). The rows `an in-flight reservation counts against the daily cap`, `… the monthly cap` and `reservations that still fit are allowed` cover this. **Mutation:** wallet-core.ts:502, dropping `dailyReservedCents`. Result: `2 failed / 117 passed`. |
| NaN / non-finite handling (`b1bcf92d7`) | FIXED (verified independently) | A non-finite cap is sanitized to 0, so it refuses. A non-finite reservation with any cap set refuses. Spendable and parent amounts go through `wholeNonNegative`. With no caps set, a NaN reservation is still refused by `covers()` in `resolveSpendSource`, and a test asserts it. **Mutation:** wallet-core.ts:512 changed to `if (false) {`. Result: `2 failed / 117 passed` (the `evaluateCaps fails closed on a non-finite reservation…` rows). |

**Mutations:** all four were made by line index with the anchor asserted, the file was restored after each one, and `git status` was clean afterwards.

**Local run on df1d3326a:** `bun run test -- src/billing/__tests__/wallet-core.test.ts src/__tests__/seams/credit-conversion.seam.test.ts src/billing/__tests__/money-model-guard.test.ts` gave `Test Files 3 passed (3)`, `Tests 150 passed (150)`, exit 0. The MON-5 guard is still green, and no new `/100` or `*100` appears in the delta. The one new rounding is `Math.round(centsFromCredits(n))`, which rounds output from the one conversion and is not a second conversion.

**Unchanged from the first review:**
- All Spec-ID titles are still `(partial)`: POL-7 ×6, SPEND-1 ×3, SPEND-4 ×5, SPEND-5 ×2, SPEND-6 ×2, WAL-3 ×5, WAL-6 ×7, WAL-7 ×7.
- `scripts/spec-coverage-allowlist.txt` has no change in this delta.
- The POL-7 titles now use the point-guard ruling's wording, and both swap directions are tested. That closes the gap I noted in the first review.
- The scope is still the leaf files only. There is no migration.

**Notes (not P-level):**
- `evaluateCaps` now takes held amounts as inputs. C3 must pass holds that EXCLUDE the hold for the call being evaluated; otherwise that reservation is counted twice. The doc comment says "outstanding holds" and does not state this exclusion.
- My earlier rounding note on `covers()` (`Math.round` of the reservation) still stands for C3.

**CI:** checks on df1d3326a were still pending at post time. The point-guard asked for this review now, so it does not cite the Spec ID coverage job for this head. The gate's merge item (c) covers CI.

packages/lib/src/billing/wallet-core.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED at df1d3326a by independent re-review 5232218520 (mutations: wallet-core.ts:124 guest fallback 5 red, :403 overshoot after fallback 3 red, :502 caps with in-flight holds 2 red, :512 NaN fail-closed 2 red). Resolving; merge waits for CI on this head.

packages/lib/src/billing/__tests__/wallet-core.test.ts
2witstudios: [point-guard] Verified FIXED at df1d3326a by independent re-review 5232218520 (mutations: wallet-core.ts:124 guest fallback 5 red, :403 overshoot after fallback 3 red, :502 caps with in-flight holds 2 red, :512 NaN fail-closed 2 red). Resolving; merge waits for CI on this head.

packages/lib/src/billing/wallet-core.ts
2witstudios: [point-guard] Verified FIXED at df1d3326a by independent re-review 5232218520 (mutations: wallet-core.ts:124 guest fallback 5 red, :403 overshoot after fallback 3 red, :502 caps with in-flight holds 2 red, :512 NaN fail-closed 2 red). Resolving; merge waits for CI on this head.

PR #2661, opened 2026-09-17
feat(permissions): pure resolveOrgDriveAccess org access core (ow-b2)

2witstudios:
## Summary
Lane **ow-b2** (Wave B, leaf `i262opm15khpplqzc3234w8w`): the pure org access core. No IO, no schema, no migration.

`packages/lib/src/permissions/org-access.ts` (new, exports entry `@pagespace/lib/permissions/org-access`):

```ts
resolveOrgDriveAccess(input: OrgDriveAccessInput): OrgDriveAccess | null

interface OrgDriveAccessInput {
  orgRole: OrgRole | null;                        // user's role in the drive's org; null = not in it / no org
  driveVisibility: OrgDriveVisibility;            // 'OPEN' | 'RESTRICTED' | 'PRIVATE'
  driveMembership: OrgDriveMembership | null;     // { role, customRoleId, source: 'invite' | 'org' }
  driveDefaultRole: DriveRoleGrant;               // { role, customRoleId } — POL-6 / O-11
}
interface OrgDriveAccess { role: 'OWNER'|'ADMIN'|'MEMBER'; customRoleId: string | null; source: 'org-admin' | 'org' | 'invite' }
```

| orgRole | row | OPEN | RESTRICTED | PRIVATE |
|---|---|---|---|---|
| OWNER / ADMIN | none, or a MEMBER row | ADMIN, `org-admin` | ADMIN, `org-admin` | ADMIN, `org-admin` (caller audits) |
| OWNER / ADMIN | `invite` ADMIN/OWNER row | the row | the row | the row |
| OWNER / ADMIN | `org` ADMIN row | the row | ADMIN, `org-admin` (stale row ignored) | ADMIN, `org-admin` (stale row ignored) |
| MEMBER | `invite` row | the row | the row | the row |
| MEMBER | `org` row | the row | **null** (stale) | **null** (stale) |
| MEMBER | none | default role, `org` | **null** | **null** |
| none | any | **null** | **null** | **null** |

A `source: 'org'` row counts only on an OPEN drive: it exists because the drive is Open, so on RESTRICTED/PRIVATE it is a leftover from a visibility change and fails closed (Codex + independent review P1). On RESTRICTED/PRIVATE only an `invite` row grants, so **an approved Restricted join must write `source: 'invite'`** (note for B3/B4/B7). `org-admin` means only "the org override was needed": a user whose own row already grants ADMIN/OWNER resolves by the row, so no spurious audit.

[...]

PR #2663, opened 2026-09-17
feat(orgs): B6 leave and delete cascades — leaveOrganization, reassignLedOrgDrives, account deletion leaves orgs first

2witstudios:
## Summary
Lane **ow-b6**, leaf `h40ht4zm4o2md5d5fktuhz6c` (Leave and delete cascades). Spec IDs: O-7, O-8, ORG-6 (partial).

- `packages/lib/src/organizations/leave.ts` (new, exported as `@pagespace/lib/organizations/leave`)
  - pure: `decideLeave(membership)` (the Owner is refused, and so is a non-member), `planLeadReassignments(userId, ledDrives)`
  - `leaveOrganization(userId, orgId, tx?)`: inside **that org's drives only** (trashed ones too), it removes the leaver's org-sourced `drive_members` rows, the agent memberships they granted (through `revokeAgentMembershipsGrantedBy`), the drive and page share links they created, and their MCP tokens' `mcp_token_drives` rows. It then reassigns the drives they lead and deletes the `org_members` row.
  - `reassignLedOrgDrives(userId, tx, { orgId?, reason?, actor? })`: moves each org drive the user leads to its org's Owner. Drive and org rows are locked `FOR UPDATE`, and each move writes an `ownership_transfer` activity event in the same transaction. This is exported on its own for B5.
  - `leaveAllOrganizations(userId, tx)`: throws `LeaveOrganizationRefusedError` on a refusal, so the caller's transaction rolls back.
- `accountRepository.deleteUser` now runs a single transaction: `leaveAllOrganizations`, then `reassignLedOrgDrives` (in case a lead is no longer an org member), then the `users` delete. The audit actor is the anonymized address, because erasure has already anonymized the activity by the time this runs.
- `getOwnedDrives` and `checkAndDeleteSoloDrives` (erasure drive disposition in the web, processor and admin paths) now skip org drives. Before this change a solo org drive was deleted, and a shared one blocked the erasure.

- **An org Owner is refused before any erasure starts** (Codex P1 on this PR; 991bebdc7, 494d465d4). `accountRepository.getOwnedOrganizationNames` backs a refusal before anything is destroyed at every entry point: self-service `DELETE /api/account` (400), the processor's first destructive step `drive-disposition` (`ERASURE_BLOCKED`, even with force), admin DSAR `DELETE /api/admin/users/[userId]/data` (400), and admin force-erasure `POST /api/admin/gdpr/erasure` (400). `deleteUser`'s `LeaveOrganizationRefusedError` remains the last line of defence.
- **Review ow-irv-2663** (5f45d601d): a rollback test (MEMBER-lead of X plus OWNER of Y; X is untouched after the refusal) and an invite-row survival test. `leaveAllOrganizations` now orders by `(joinedAt, id)`. The worker maps a late `LeaveOrganizationRefusedError` to `ERASURE_BLOCKED` (terminal). Mutations: running `leaveAllOrganizations` on `db` instead of `tx` turns the rollback test red; dropping `eq(driveMembers.source, 'org')` turns the invite test red; changing `instanceof` to `false` in the worker turns its test red.
- **CI fixes** (8cf58c320): the leave integration suite now ends its pg pool in `afterAll`. Without that, CI's lib `test:integration` step ran out of connections (`53300 too many clients`) in three later suites. O-7 and O-8 unit titles no longer carry `(partial)`, which the spec-coverage gate rejects after IDs it does not track.

## ID / test table
| ID | Test | File |
|---|---|---|
| ORG-6 (partial) | deleting a lead's account keeps the org drive and makes the org Owner its lead | leave.integration.test.ts |
| ORG-6 (partial) | deleteUser refuses an org Owner and rolls back the orgs it had already left | leave.integration.test.ts |
| ORG-6 (partial) | a late org-Owner refusal at delete-user blocks the request instead of retrying it | apps/processor/.../account-erasure-worker.test.ts |
| ORG-6 (partial) | erasure drive disposition never counts or deletes an org drive the user leads | leave.integration.test.ts |
| ORG-6 (partial) | the Owner cannot leave without transferring ownership first | leave.test.ts |
| O-8 | leaving an org revokes all four artifact types the leaver handed out in its drives | leave.integration.test.ts |
| O-8 | a leaver's own invite rows and artifacts in non-org drives are untouched, as are other members' in the org | leave.integration.test.ts |
| O-8 | an org Owner cannot leave and nothing changes | leave.integration.test.ts |
| O-8 | a leaver's invite-sourced membership in an org drive survives leave, so they keep access as a guest | leave.integration.test.ts |
| O-7 | leaving an org reassigns the drives the leaver leads to the org Owner with an audit event | leave.integration.test.ts |
| O-7 | reassignLedOrgDrives with no org given reassigns led drives across every org, each to its own Owner | leave.integration.test.ts |
| ORG-6 (partial) | given the user owns an organization, should block with 400 and NOT queue | apps/web/.../account/__tests__/delete-route.test.ts |
| ORG-6 (partial) | an org Owner is blocked at the first step and nothing is destroyed | apps/processor/.../account-erasure-worker.test.ts |
| ORG-6 (partial) | DELETE of an org Owner returns 400 before deleting any drive or data | apps/admin/.../data/__tests__/route.test.ts |
| ORG-6 (partial) | given the subject owns an organization, should refuse with 400 and queue nothing, even with force | apps/web/.../admin/gdpr/__tests__/admin-gdpr-routes.test.ts |
| O-7 | deleteUser leaves every org and reassigns led org drives in the same transaction, before the users row goes, and does not delete the users row when leaving is refused | account-repository.test.ts |

**ORG-6 stays on the allowlist.** Its org-delete clause and its Owner-account-deletion clause belong to B3 (leaf t718n05y). Only the "nothing orphaned" part for a lead's account deletion is proven here, so the tests carry `(partial)`. The orchestrator gave the title without the marker; rule 2 decided this. O-7 and O-8 are not gate IDs, so their titles carry the bare ID.

## Evidence
Commands were run against a throwaway homebrew pg17 (session TZ America/Chicago), migrated from scratch, using a source-alias vitest config (no dist builds):
```
DATABASE_URL=postgresql://postgres@127.0.0.1:5611/pagespace_test bunx vitest run --config vitest.wt.config.ts \
  src/organizations/__tests__/leave.integration.test.ts src/organizations/__tests__/leave.test.ts src/repositories/__tests__/account-repository.test.ts
 ✓ src/repositories/__tests__/account-repository.test.ts (18 tests)
 ✓ src/organizations/__tests__/leave.integration.test.ts (8 tests)
 ✓ src/organizations/__tests__/leave.test.ts (6 tests)
 Test Files  3 passed (3)
      Tests  32 passed (32)
```
- **Red first:** before the `account-repository` change, the ORG-6 lead-delete test failed with `expected [] to have a length of 2 but got +0`. Both org drives, including the trashed one, were cascade-deleted.
- **Mutation (by line index):** I commented out `leave.ts:112` (`await tx.update(drives).set({ ownerId: r.toUserId })…`). Three tests went red: O-7 leave reassigns, **ORG-6 (partial) deleting a lead's account keeps the org drive…**, and O-7 reassign across orgs. After restoring the line, 32/32 pass.
- **Owner preflight, run locally from source:** web `delete-route.test.ts` 10/10, `admin-gdpr-routes.test.ts` 14/14, admin `data/__tests__/route.test.ts` 14/14, processor `account-erasure-worker.test.ts` 2/2 (`bunx vitest run --config vitest.wt.config.ts <file>`, which uses aliases to the lib/db source). **Mutation:** in each of the four guards I changed `ownedOrganizations.length > 0` to `false` by line index. Each one turned exactly its new test red (web route.ts:221, worker :89, admin data route :57, gdpr erasure route :52). After restoring, all pass.
- **Spec gate, run locally:** `bun run scripts/check-spec-coverage.ts --offline --ids ORG-6` over a local vitest JSON of the organizations tests reports `spec-coverage: OK` (ORG-6 allowlisted; the partial tests are listed).
- `tsc -p` against the touched files passed (exit 0). packages/lib has no eslint config. Full typecheck, lint and knip are left to CI (build slot rule).

## Callers of the changed disposition semantics (grep: getOwnedDrives, checkAndDeleteSoloDrives, getDriveMemberCount)
| Caller | What it uses | Effect of the change | Does it now skip data it must handle? |
|---|---|---|---|
| `apps/web/src/app/api/account/route.ts:221` (self-service DELETE pre-check) | `getOwnedDrives` + `planDriveDisposition` | Org drives the user leads no longer block the request. | No: `deleteUser` reassigns them to the org Owner in the delete transaction. |
| `apps/processor/src/workers/account-erasure-worker.ts:86` (`drive-disposition` step) | `getOwnedDrives` → `deleteDrive` | A solo org drive is no longer deleted, and a shared one no longer blocks the job. | No: the later `delete-user` step reassigns them atomically. |
| `apps/admin/src/app/api/admin/users/[userId]/data/route.ts:56` | `checkAndDeleteSoloDrives` | Same as the processor step. | No: `deleteUser` runs later in that route. |

Paths that do **not** use these functions keep their own `ownerId` queries, and this PR leaves them unchanged:
- `apps/web/src/app/api/account/drives-status/route.ts` still lists org drives the user leads as solo or multi-member drives, and the settings account page shows them as needing transfer or delete before account deletion. It over-includes but loses nothing. It is outside this leaf's files, so I flagged it in the channel for B5's account-route guard.
- `apps/web/src/app/api/account/handle-drive/route.ts` `delete` still lets a lead hard-delete an org drive during account-deletion prep, and `transfer` lets them hand it to a drive admin. That is a lead-rules gap and is flagged for B5.
- `apps/web/src/app/api/admin/gdpr/erasure/route.ts` only calls `findById` and enqueues; the worker's disposition step applies. It now also has the Owner preflight.
- `packages/lib/src/compliance/export/gdpr-export.ts:529` exports drives by `ownerId` and is unaffected. Org drives the subject leads are still exported as owned.

## Notes for reviewers / other lanes
- An **Owner's** account erasure is refused up front at every entry point (see above), and `deleteUser` refusing is only a backstop. B3's leaf also names "an Owner cannot delete their account while owning an org"; B3 should reuse `getOwnedOrganizationNames` rather than add a second check.
- Invite-sourced `drive_members` rows in org drives are kept on leave, so the person becomes a guest. This follows O-8's "org-sourced membership rows".
- No migration and no schema change.

https://claude.ai/code/session_017BLWtpvtJYdyfpvEhyARFd


2witstudios:
@codex review

packages/lib/src/repositories/account-repository.ts
```diff
@@ -50,11 +52,12 @@ export const accountRepository = {
...
 
   /**
-   * Get all drives owned by a user
+   * Get the personal drives owned by a user. Org drives the user leads are not theirs to
+   * dispose of: deleteUser reassigns them to the org Owner (Spec O-7).
    */
   getOwnedDrives: async (userId: string): Promise<OwnedDrive[]> => {
     return db.query.drives.findMany({
-      where: eq(drives.ownerId, userId),
+      where: and(eq(drives.ownerId, userId), isNull(drives.orgId)),
```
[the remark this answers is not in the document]
2witstudios: Valid. Fixed in 991bebdc7, with 494d465d4 covering a fifth entry point. `accountRepository.getOwnedOrganizationNames(userId)` now backs a refusal **before anything is destroyed** at every erasure entry point:
- self-service `DELETE /api/account`: 400 `ownedOrganizations` (`apps/web/src/app/api/account/route.ts`)
- processor `drive-disposition`, the first destructive step: throws `ERASURE_BLOCKED`, so the DSR goes to `blocked` even with force (`apps/processor/src/workers/account-erasure-worker.ts`)
- admin DSAR `DELETE /api/admin/users/[userId]/data`: 400 before `checkAndDeleteSoloDrives` (`apps/admin/.../data/route.ts`)
- admin force-erasure `POST /api/admin/gdpr/erasure`: 400 before lodging or re-queueing (`apps/web/src/app/api/admin/gdpr/erasure/route.ts`)

Each has an `ORG-6 (partial)` test asserting nothing destructive is called. Each guard was also mutated by line index (`ownedOrganizations.length > 0` changed to `false`), and each test went red, then green after restoring. `deleteUser` still throws `LeaveOrganizationRefusedError` as a last line of defence. Leaving this thread open for the reviewer to verify.

packages/lib/src/organizations/__tests__/leave.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(decideLeave(null)).toEqual({ ok: false, reason: 'NOT_A_MEMBER' });
+  });
+
+  it('ORG-6 (partial) the Owner cannot leave without transferring ownership first', () => {
+    expect(decideLeave({ role: 'OWNER' })).toEqual({ ok: false, reason: 'OWNER_MUST_TRANSFER' });
+  });
+});
+
+describe('planLeadReassignments', () => {
+  it('O-7 (partial) every org drive the departing lead leads goes to that org\'s Owner', () => {
```
2witstudios: **P1 [independent-review]: Spec ID coverage fails on this head.** O-7 and O-8 are Open items, not Spec requirement IDs, so the gate reads `(partial)` with no valid ID in front and rejects the title. CI job 105090288557 prints `FAIL: malformed partial marker in packages/lib/src/organizations/__tests__/leave.test.ts: "O-7 (partial) every org drive the departing lead leads goes to that org's Owner"` and the same for lines 5, 10, 31 and 37, then `spec-coverage: FAILED`. Drop `(partial)` from every O-7/O-8 title in this file. The O-n prefix alone gives traceability.

packages/lib/src/organizations/__tests__/leave.integration.test.ts
```diff
@@ -0,0 +1,250 @@
...
+      .from(drives).where(inArray(drives.id, [led.id, trashedLed.id]));
+    expect(after).toHaveLength(2);
+    for (const d of after) expect(d).toMatchObject({ ownerId: owner.id, orgId: org.id });
+    const events = await db.select().from(activityLogs)
+      .where(and(eq(activityLogs.resourceId, led.id), eq(activityLogs.operation, 'ownership_transfer')));
+    expect(events[0]).toMatchObject({ userId: null, metadata: { orgId: org.id, reason: 'account_deleted' } });
+    expect(events[0].actorEmail).not.toBe(lead.email);
+  });
+
+  it('ORG-6 (partial) an org Owner\'s account deletion is refused and changes nothing', async () => {
```
2witstudios: **P2 [independent-review]: this test cannot detect a non-atomic deleteUser.** The Owner has nothing to cascade, and the refusal happens before any write. Mutation: I replaced `db.transaction(async (tx) => {…})` in `deleteUser` with direct execution on `db`, and this file stayed 8/8 green. Add a scenario where the person is a MEMBER and lead of org X, with an org row and a share link, and the OWNER of org Y. Assert that X's membership, lead, row, link and the absence of an ownership_transfer event all survive the refusal. I ran that scenario as a throwaway probe: the code does roll back, and the same `leaveAllOrganizations` run without a transaction leaves X partially applied, so the scenario detects a missing transaction.

2witstudios (commented):
[independent-review] agent:ow-irv-2663

Independent spec review of #2663 (lane ow-b6, leaf h40ht4zm4o2md5d5fktuhz6c) at head `1fe0a5203`, under Sequence Spec gate 0d. I changed no code. **Verdict: changes requested. 1 P1, 3 P2.**

## Per-ID verdicts
| ID | Claim in PR | Verdict | Evidence |
|---|---|---|---|
| ORG-6 | `ORG-6 (partial)` ×4 (3 in `leave.integration.test.ts`, 1 in `leave.test.ts`) | **partial: correctly marked** | The marker format is right, and CI lists ORG-6 as `allowlisted partial: … leave.integration.test.ts, … leave.test.ts`. ORG-6 stays on the allowlist, which is correct: its org-delete ("transfers or trashes, nothing orphaned") and upfront Owner-refusal clauses belong to B3 (t718n05y). The PR makes no plain ORG-6 claim, so there is no over-claim. The trashed-drive org-delete case is **not** in this diff (checked). |
| O-7 | plain `O-7` (integration and mock tests) + `O-7 (partial)` (unit) | **satisfied** for "reassigns to the org Owner with an audit event" on leave and on delete | Proven by `O-7 leaving an org reassigns … with an audit event` and `ORG-6 (partial) deleting a lead's account keeps the org drive …` on pg17. **But the `O-7 (partial)` titles break CI (P1 below).** O-7 is an Open item, not a gate ID. |
| O-8 | plain `O-8` (integration) + `O-8 (partial)` (unit) | **satisfied**, with one untested branch (P2-2) | `O-8 leaving an org revokes all four artifact types …` seeds and asserts drive_members(org), driveAgentMembers, drive+page share links and mcp_token_drives. `O-8 a leaver's own invite rows and artifacts in non-org drives are untouched …` covers the leaf's negative. **The `O-8 (partial)` titles also break CI (P1).** |

## Required checks
**(a) Key test goes red when reassignment is skipped. Confirmed.** I ran my own line-index mutation on a throwaway homebrew pg17 (session TZ America/Chicago), migrated from scratch. Blanking `leave.ts:112` (`tx.update(drives).set({ ownerId })`) sent 3 tests red, including `ORG-6 (partial) deleting a lead's account keeps the org drive and makes the org Owner its lead`. I restored the line; a control run is 32/32.

**(b) An Owner's deletion is refused and nothing changes.** The behavior is correct, but the PR's test cannot prove it (P2-1). I wrote a throwaway probe (deleted afterwards): U is a MEMBER and drive lead in org X, with an org row and a share link, and is the OWNER of org Y. X is processed first, which I confirmed. `deleteUser(U)` rejects with `LeaveOrganizationRefusedError`, and the before and after snapshots are equal: X membership, lead, org row, link, 0 audit events, users row. As a control, the same `leaveAllOrganizations` run against `db` with no transaction leaves X's cascade applied (xMember 0, link 0, 1 event), so the probe does detect a partial apply.

**(c) All four artifact types are revoked, and a guest's non-org rows are untouched. Confirmed.** Blanking the `inArray(mcpTokenDrives.driveId, orgDriveIds)` filter (`leave.ts:209`) sends the non-org test red. The scoped-key semantics are also safe: a scoped key left with zero drive rows fails closed in `validateMCPToken`, the same pattern the existing member-remove route uses. Separate probe: after leave, `getUserDriveAccess(leaver, orgDrive)` is false, and a non-member leave returns `NOT_A_MEMBER`.

**(d) Every caller of the `isNull(drives.orgId)` filter:**
| Caller | Skips data it must still handle? |
|---|---|
| `apps/web/src/app/api/account/route.ts:221` (self-service DELETE pre-check) | No for a lead: `deleteUser` reassigns led org drives. **For an org Owner the pre-check now passes and the erasure is queued, then fails late** (P2-3). |
| `apps/processor/src/workers/account-erasure-worker.ts:86` (drive-disposition) | No: led org drives are reassigned at `delete-user`. For an Owner, see P2-3. |
| `apps/admin/src/app/api/admin/users/[userId]/data/route.ts:56` (`checkAndDeleteSoloDrives`) | No for a lead. For an Owner, see P2-3. |
| `apps/web/src/app/api/account/drives-status/route.ts`, `handle-drive/route.ts` | Neither calls these functions, so this PR does not change them. drives-status over-lists led org drives, and handle-drive `delete` still lets a lead hard-delete an org drive. The lane flagged both for B5, and I agree. |
| `packages/lib/src/compliance/export/gdpr-export.ts` | Uses its own `ownerId` query, so it is unaffected and still exports led org drives. |
| Admin tools | No other callers (grepped `getOwnedDrives\|checkAndDeleteSoloDrives\|deleteUser(` across apps/ and packages/). |

**(e) Audit actor. Correct.** Delete-path events carry `deleted_user_<sha256[:12]>` / `Deleted User`, and `userId` becomes null after the FK cascade. Mutation: blanking `account-repository.ts:108` (the `actor:` line) makes the code fall back to `getActorInfo`, and `ORG-6 (partial) deleting a lead's account …` goes red, so no plaintext email reaches these rows. The voluntary-leave path writes the live email, as every activity row does, and the erasure anonymizes it later.

**(f) The org-delete trashed-drive case is absent from this diff. Correct** (B3's scope, per the leaf's Ownership correction).

## Rules
- The pure decision logic (`decideLeave`, `planLeadReassignments`) is separate from IO, with direct unit tests. **OK**
- No `any`. `tx as unknown as typeof db` matches the existing `logActivityWithTx` call pattern. **OK**
- No env flags, no `now()`, no bulk inserts. **OK**
- The new subpath `organizations/leave` has both an `exports` entry and a `typesVersions` entry. **OK**
- Permission checks are left to the caller routes (B-lane). **OK**
- No migration or schema change. Scope stays within the leaf (the account-repository change is the leaf's "account deletion calls the same function first"). **OK**

## Test runs
Local, against homebrew pg17 on a scratch cluster and database, with a source-alias vitest config and the files run directly: `leave.integration.test.ts` 8/8, `leave.test.ts` 6/6, `account-repository.test.ts` 18/18, **32/32**.

Mutations by line index. After each one the file was restored and the tree was clean.
| # | Line | Result |
|---|---|---|
| M1 | `leave.ts:112` reassign update blanked | RED (3) |
| M2 | `account-repository.ts:108` anonymized actor blanked | RED (1) |
| M3 | `account-repository.ts:112` second `reassignLedOrgDrives` blanked | red **only** in the mocked `account-repository.test.ts`; integration stays green |
| M4 | `leave.ts:209` MCP org-drive filter blanked | RED (1) |
| M5 | `leave.ts:178` `eq(driveMembers.source, 'org')` blanked | **SURVIVED 32/32** (P2-2) |
| M6 | `account-repository.ts:105/114` `db.transaction` replaced by direct execution on `db` | **SURVIVED 8/8 integration** (P2-1) |

## CI on this head (run 35184615199)
- **Spec ID coverage FAILED because of this PR** (job 105090288557):
  - `ORG-6     allowlisted  partial: packages/db/src/__tests__/organizations.integration.test.ts, packages/db/src/schema/__tests__/organizations.test.ts, packages/lib/src/organizations/__tests__/leave.integration.test.ts, packages/lib/src/organizations/__tests__/leave.test.ts`
  - `covered 9/87, allowlisted-missing 78, MISSING 0`
  - `FAIL: malformed partial marker in packages/lib/src/organizations/__tests__/leave.test.ts: "O-7 (partial) …"` (×3) and `"O-8 (partial) …"` (×2), then `spec-coverage: FAILED`
- **Unit Tests failed** (job 105083885549) in `@pagespace/lib test:integration`. The 3 failing suites are `session-repository-resource`, `replay-across-replicas` and `activity-logger-fk-retry`, all with `error: sorry, too many clients already` (53300). This PR's `leave.integration.test.ts` passed in both the unit and integration passes (8/8). The failure looks environmental (connection exhaustion), not this diff, so rerun the failed jobs before blaming it. The PR adds no new pool.

## Findings
**P1-1: The spec coverage gate fails on `O-7 (partial)` / `O-8 (partial)` titles** (`leave.test.ts` lines 5, 10, 20, 31, 37). O-7 and O-8 are Open items, not Spec requirement IDs, so the gate reads a `(partial)` marker with no valid ID in front and fails. Drop `(partial)` from the O-n titles (the O-n prefix alone is traceability) or rephrase them. This is a hard red check on this head.

**P2-1: Nothing tests that `deleteUser` is atomic** (`leave.integration.test.ts:206`). `an org Owner's account deletion is refused and changes nothing` seeds an Owner with nothing to cascade, and the refusal comes before any write, so there is nothing to roll back. M6 (the transaction removed) stays 8/8 green. The mocked `does not delete the users row when leaving is refused` only checks that `tx.delete` was not called. Add the scenario from my probe: the person is a member and lead of org X, with an org row and artifacts, and Owner of org Y. Assert that X's cascade and reassignment are all still present after the refusal.

**P2-2: Keeping invite-sourced rows in org drives is claimed but untested** (`leave.ts:178`). The PR body says invite rows in org drives are kept, so the person becomes a guest. Blanking `eq(driveMembers.source, 'org')` (M5) keeps 32/32 green: a leaver's invite membership in an org drive would be silently deleted. Seed an `invite` row in an org drive and assert it survives, and that `getUserDriveAccess` is still true for that drive.

**P2-3: The Owner refusal happens after the irreversible erasure steps, and the error is retried** (`account-repository.ts:110`). "Refused and changes nothing" holds only inside `deleteUser`. In the processor, drive-disposition (which deletes personal solo drives), anonymize, purge-ai-usage, purge-monitoring, revoke-integrations and email-suppression all run before `delete-user`. In the admin route, `checkAndDeleteSoloDrives`, anonymize and purges run before it. `LeaveOrganizationRefusedError` does not carry `ERASURE_BLOCKED_PREFIX`, so pg-boss retries it with backoff instead of treating it as terminal-blocked. With the filter in place, the web pre-check no longer stops an Owner either. The partial erasure existed before this PR (the organizations.ownerId RESTRICT failed it at the same step), and the upfront refusal is B3's. Still, this PR's test title and body claim more than holds end to end. At minimum, reword the claim to "deleteUser refuses and rolls back". The orchestrator should decide whether B6 or B3 classifies the refusal as ERASURE_BLOCKED and moves an Owner check ahead of drive-disposition. I am not making that call here.

## Non-blocking notes
- The second `reassignLedOrgDrives` call in `deleteUser` (for a lead who is no longer a member) is covered only by the mock (M3). An integration case would be a lead row in an org drive with no `org_members` row.
- `getActorInfo` in the leave path reads through the global `db`, not `tx`. That is harmless (users row, no locks), but not transaction-consistent.
- `leaveAllOrganizations` has no `ORDER BY`, so which org refuses first can vary. The transaction makes that harmless.
- Leave does not revoke `page_permissions` grants to the leaver in org drives. O-8 does not list them, so this is not a deviation, only a note for B5/B3.

packages/lib/src/organizations/leave.ts
```diff
@@ -0,0 +1,264 @@
...
+  const decision = decideLeave(membership ?? null);
+  if (!decision.ok) return decision;
+
+  const orgDriveIds = tx.select({ id: drives.id }).from(drives).where(eq(drives.orgId, orgId));
+
+  const memberRows = await tx
+    .delete(driveMembers)
+    .where(and(
+      eq(driveMembers.userId, userId),
+      eq(driveMembers.source, 'org'),
```
2witstudios: **P2 [independent-review]: the claim that invite-sourced rows in org drives survive has no test.** The PR body says invite rows in org drives are kept, so the person becomes a guest. Mutation: I blanked this line (`eq(driveMembers.source, 'org')`) and all 32 tests stayed green, meaning a leaver's invite membership in an org drive would be silently deleted. Seed an `invite` row in an org drive and assert that it survives leave and that `getUserDriveAccess` is still true for that drive.

packages/lib/src/repositories/account-repository.ts
```diff
@@ -90,10 +93,25 @@ export const accountRepository = {
...
+   * the erasure has anonymized the user's activity.
    */
   deleteUser: async (userId: string): Promise<void> => {
-    await db.delete(users).where(eq(users.id, userId));
+    await db.transaction(async (tx) => {
+      const options = {
+        reason: 'account_deleted' as const,
+        actor: { actorEmail: createAnonymizedActorEmail(userId), actorDisplayName: 'Deleted User' },
+      };
+      await leaveAllOrganizations(userId, tx, options);
```
2witstudios: **P2 [independent-review]: the Owner refusal fires after the erasure's irreversible steps and is retried.** "Refused and changes nothing" is true only inside `deleteUser`. In the processor, drive-disposition (it deletes solo personal drives), anonymize, purges, revoke-integrations and email-suppression all run before `delete-user`. In the admin route, `checkAndDeleteSoloDrives`, anonymize and purges run first. `LeaveOrganizationRefusedError` has no `ERASURE_BLOCKED_PREFIX`, so pg-boss retries it with backoff instead of treating it as terminal. With the new org-drive filter, the web pre-check no longer stops an Owner either. The partial erasure existed before this PR (the organizations.ownerId RESTRICT FK failed at the same step), and the upfront refusal is B3's. Narrow the test title and PR claim to "deleteUser refuses and rolls back". The orchestrator should route the blocked-classification question to B6 or B3.

packages/lib/src/organizations/__tests__/leave.test.ts
2witstudios: Fixed in 8cf58c320: every O-7/O-8 title in `leave.test.ts` now carries the bare ID. Local check: `bun run scripts/check-spec-coverage.ts --offline --ids ORG-6`, run over a vitest JSON of the organizations tests, reports `spec-coverage: OK` with no malformed-marker lines. Leaving this open for verification against CI on the new head.

packages/lib/src/organizations/__tests__/leave.integration.test.ts
2witstudios: Fixed in 5f45d601d with the scenario you described. New test: `ORG-6 (partial) deleteUser refuses an org Owner and rolls back the orgs it had already left`. The person is a MEMBER leading a drive in X (joinedAt 2026-01-01), has an org row and all four artifacts in another X drive, and is OWNER of Y (joinedAt 2026-02-01). After the refusal it asserts the users row, X membership, lead, org row, artifacts, and zero `ownership_transfer` events. `leaveAllOrganizations` now orders by `(joinedAt, id)`, so X is always left before Y refuses. **Mutation:** changing `leaveAllOrganizations(userId, tx, options)` to run on `db` in `account-repository.ts` made this test red (1 failed / 8 passed); after restoring, 9/9. The old Owner-only test was folded into this one, which also gives the narrower title you asked for.

packages/lib/src/organizations/leave.ts
2witstudios: Fixed in 5f45d601d. New test: `O-8 a leaver's invite-sourced membership in an org drive survives leave, so they keep access as a guest`. It seeds an `invite` row in a PRIVATE org drive, then asserts `revoked.orgMembershipRows === 0`, that the row survives, and that `getUserDriveAccess(leaver, drive)` is true. **Mutation:** commenting out `eq(driveMembers.source, 'org')` in `leave.ts` made exactly this test red (1 failed / 8 passed); after restoring, 9/9.

packages/lib/src/repositories/account-repository.ts
2witstudios: Addressed in three commits:
- **Upfront refusal before anything irreversible** (991bebdc7, 494d465d4, in reply to the Codex P1 above). `accountRepository.getOwnedOrganizationNames` backs a refusal in self-service `DELETE /api/account`, admin DSAR `DELETE …/data` (before `checkAndDeleteSoloDrives`) and admin force-erasure `POST /api/admin/gdpr/erasure` (400, nothing queued). In the processor it runs at the start of `drive-disposition`, the first destructive step, throwing `ERASURE_BLOCKED` even with force. Each guard went red under a line-index mutation.
- **Terminal, not retried** (5f45d601d). The worker's `delete-user` step now maps `LeaveOrganizationRefusedError` to `ERASURE_BLOCKED:`, so someone who became an Owner after the preflight is marked blocked, not retried with backoff. Test: `ORG-6 (partial) a late org-Owner refusal at delete-user blocks the request instead of retrying it`. Mutating the `instanceof` check to `false` made it red.
- **Title and claim narrowed** (5f45d601d): the integration test is now "deleteUser refuses an org Owner and rolls back …".

2witstudios (commented):
[independent-review] agent:ow-irv-2663 (re-review 1fe0a5203..5f45d601d)

I re-reviewed the four commits since my first review (5231829193): 991bebdc7, 494d465d4, 8cf58c320 and 5f45d601d. I changed no code. **Verdict: every earlier finding is FIXED, and there are no new P1 or P2 findings.** One merge-gate item remains (stale base, below).

## Findings
| Finding | Status | Verification |
|---|---|---|
| **P1-1** `O-7 (partial)` / `O-8 (partial)` titles failed Spec ID coverage | **FIXED** | All five titles in `leave.test.ts` dropped `(partial)`. CI job 105103111638 on `5f45d601d` prints `spec-coverage: OK` and `covered 9/87, allowlisted-missing 78, MISSING 0`, with no `malformed partial marker` lines. |
| **P2-1** Nothing tested that `deleteUser` is all-or-nothing | **FIXED** | New test `ORG-6 (partial) deleteUser refuses an org Owner and rolls back the orgs it had already left` covers my scenario: the person is a member and drive lead in org X, with an org row and all four artifacts, and owns org Y. `leaveAllOrganizations` now orders memberships by `joinedAt, id`, so X is left before Y refuses. **Mutation M6:** I replaced `db.transaction(...)` in `deleteUser` (`account-repository.ts:119/128`) with direct execution on `db`, and this test went **RED**. That mutant survived at 8/8 on the first head. |
| **P2-2** Keeping invite rows in org drives was untested | **FIXED** | New test `O-8 a leaver's invite-sourced membership in an org drive survives leave, so they keep access as a guest` (PRIVATE org drive; asserts the row survives and `getUserDriveAccess` is true). **Mutation M5:** I blanked `leave.ts:178` (`eq(driveMembers.source, 'org')`), and this test went **RED**. That mutant survived at 32/32 before. |
| **P2-3** An Owner was refused only after the irreversible erasure steps, and the error was retried | **FIXED** | `accountRepository.getOwnedOrganizationNames` now runs as a preflight at 4 entry points, and the late refusal at delete-user is rewrapped as `ERASURE_BLOCKED`. Each change was verified with its own line-index mutation (table below); every one went RED. `drive-disposition` is the first step in `ERASURE_STEPS` (`erasure-plan.ts:59`), so the worker preflight fires before anything is destroyed. The self-service web route refuses before `lodgeAndEnqueueErasure` (the Stripe step is recorded there), so nothing is queued. |

## Mutations on the new preflights
In each mutation I replaced the `if (…)` guard line with `if (false) {`, ran the test, then restored the file.
| Line | Test that went RED |
|---|---|
| `apps/web/src/app/api/account/route.ts:221` | `ORG-6 (partial) given the user owns an organization, should block with 400 and NOT queue` |
| `apps/web/src/app/api/admin/gdpr/erasure/route.ts:52` | `ORG-6 (partial) given the subject owns an organization, should refuse with 400 and queue nothing, even with force` |
| `apps/admin/src/app/api/admin/users/[userId]/data/route.ts:57` | `ORG-6 (partial) DELETE of an org Owner returns 400 before deleting any drive or data` |
| `apps/processor/src/workers/account-erasure-worker.ts:90` | `ORG-6 (partial) an org Owner is blocked at the first step and nothing is destroyed` |
| `account-erasure-worker.ts:203` (`instanceof LeaveOrganizationRefusedError`) | `ORG-6 (partial) a late org-Owner refusal at delete-user blocks the request instead of retrying it` |

## Test runs
- **Local:** throwaway homebrew pg17 (session TZ America/Chicago), migrated from scratch, with source-alias vitest configs.
  - lib: `leave.integration.test.ts` 9/9, `leave.test.ts` 6/6, `account-repository.test.ts` 18/18 (**33/33**)
  - processor: `account-erasure-worker.test.ts` 3/3
  - admin: data `route.test.ts` 14/14
  - web: `delete-route.test.ts` 10/10 and `admin-gdpr-routes.test.ts` 14/14
- **CI on `5f45d601d`:** every check is green, including Unit Tests 105095676223 (the earlier `too many clients` failure is gone; the new `pool.end()` in afterAll follows the existing drive-envs and agent-sessions pattern), Lint & TypeScript, E2E, Security Test Suite and CodeQL.
- **Spec coverage for this PR's ID:** `ORG-6     allowlisted  partial: apps/admin/…/data/__tests__/route.test.ts, apps/processor/src/workers/__tests__/account-erasure-worker.test.ts, apps/web/src/app/api/account/__tests__/delete-route.test.ts, apps/web/src/app/api/admin/gdpr/__tests__/admin-gdpr-routes.test.ts, packages/db/…, packages/lib/src/organizations/__tests__/leave.integration.test.ts, packages/lib/src/organizations/__tests__/leave.test.ts`
- **ORG-6 correctly stays partial and allowlisted.** Its Owner-account clause is now enforced at every erasure entry point, but the org-delete clause ("transfers or trashes, nothing orphaned") is still B3's.

## Rules and scope
- No `any`, no env flag, no `now()`, no migration.
- The preflight is a plain repository read, and the refusal decision stays in the routes and worker.
- The web and admin routes are outside the leaf's named files. They are the direct fix for my P2-3 and Codex's owner-erasure P1, so I accept them as in scope.

## Merge gate items (not code findings)
- **Stale base:** `merge-base` is `f62ca8b1b` but `pu/org-wallets` is at `95a049e70` (#2661, B2 `resolveOrgDriveAccess`). The overlap is `packages/lib/package.json` (exports map, expect a conflict). B2 also changed how org-materialized rows resolve for Restricted/Private drives, and the new P2-2 test asserts `getUserDriveAccess` on a PRIVATE org drive. Under 0e this needs a rebase, and those assertions need re-running on top of B2; review carry-over does not apply to them.

## Non-blocking notes
- The rollback test relies on the new `orderBy(joinedAt, id)` to process X first and never asserts that order itself. M6 going red shows it exercises the rollback today.
- `getOwnedOrganizationNames` reads `organizations.ownerId`, while `deleteUser` refuses on the `org_members` OWNER role. Both are kept in sync by ORG-1 today, and the late ERASURE_BLOCKED rewrap covers any drift.

packages/lib/src/repositories/account-repository.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED by independent reviews 5231829193 then 5232213268 on 5f45d601d (P1 partial-marker titles, three P2s and Codex's owner-erasure P1 all FIXED; mutations red: invite source filter leave.ts:178, deleteUser transaction removal, all four owner preflights). Head 400b25d85 is a tip merge with an identical patch-id, so the review carries over. Resolving; merge waits for CI.

packages/lib/src/organizations/__tests__/leave.test.ts
2witstudios: [point-guard] Verified FIXED by independent reviews 5231829193 then 5232213268 on 5f45d601d (P1 partial-marker titles, three P2s and Codex's owner-erasure P1 all FIXED; mutations red: invite source filter leave.ts:178, deleteUser transaction removal, all four owner preflights). Head 400b25d85 is a tip merge with an identical patch-id, so the review carries over. Resolving; merge waits for CI.

packages/lib/src/organizations/__tests__/leave.integration.test.ts
2witstudios: [point-guard] Verified FIXED by independent reviews 5231829193 then 5232213268 on 5f45d601d (P1 partial-marker titles, three P2s and Codex's owner-erasure P1 all FIXED; mutations red: invite source filter leave.ts:178, deleteUser transaction removal, all four owner preflights). Head 400b25d85 is a tip merge with an identical patch-id, so the review carries over. Resolving; merge waits for CI.

packages/lib/src/organizations/leave.ts
2witstudios: [point-guard] Verified FIXED by independent reviews 5231829193 then 5232213268 on 5f45d601d (P1 partial-marker titles, three P2s and Codex's owner-erasure P1 all FIXED; mutations red: invite source filter leave.ts:178, deleteUser transaction removal, all four owner preflights). Head 400b25d85 is a tip merge with an identical patch-id, so the review carries over. Resolving; merge waits for CI.

packages/lib/src/repositories/account-repository.ts
2witstudios: [point-guard] Verified FIXED by independent reviews 5231829193 then 5232213268 on 5f45d601d (P1 partial-marker titles, three P2s and Codex's owner-erasure P1 all FIXED; mutations red: invite source filter leave.ts:178, deleteUser transaction removal, all four owner preflights). Head 400b25d85 is a tip merge with an identical patch-id, so the review carries over. Resolving; merge waits for CI.

PR #2664, opened 2026-09-17
feat(lib): org membership materialization sync (D-OW-6) — lane ow-b4

2witstudios:
## Lane ow-b4 — membership materialization service (D-OW-6)

Leaf `j5ep41ougwpkreblcgln6wjs`. Keeps `drive_members` rows with `source = 'org'` in step with org membership and drive visibility (O-6 / D-OW-6), with the move-out keep-or-remove choice (O-10 / D-OW-10). No schema change, no migration.

### Shape
- **Pure decision** — `packages/lib/src/services/org-membership-sync-core.ts`
  - `planDriveOrgMembership({ drive, orgMemberUserIds, existingRows, userScope?, removedOrgRows? })` → `{ inserts, deletes, conversions, repairs }`. Only an Open org drive materializes; only `source = 'org'` rows are ever deleted/converted/repaired; an `invite` row (guest or directly invited member, accepted or pending) is never touched or rewritten and satisfies an org member; the drive lead (`drives.ownerId`) never gets a row; a drive in step plans nothing.
  - Keep-as-invited (D-OW-10) applies only when `drives.orgId` is null; a leave or visibility change always deletes.
  - `repairs`: org rows whose `customRoleId` drifted from the drive default, or unaccepted, go back to the default (role column untouched).
  - A pending invite row is never accepted on the member's behalf (a guest who joins then leaves would lose it); an org member without an accepted row resolves Open access through `resolveOrgDriveAccess` (B2), which needs no row.
  - `summarizeAffectedUsers(plans)` → one entry per user (removed > added > role_changed).
  - `chunk(items, size)`; `settleInBatches(tasks, size)` bounds event fan-out.
- **Thin IO** — `packages/lib/src/services/org-membership-sync.ts`
  - `syncOrgMembership(orgId, opts?)` — every drive of an org.
  - `syncOrgMemberAccess(orgId, userId, opts?)` — one user (join / leave).
  - `syncDriveOrgMembership(driveId, opts?)` — one drive (visibility change, move in / out).
  - `opts = { tx?, ports? }`; `syncDriveOrgMembership` alone also takes `removedOrgRows?: 'delete' | 'keepAsInvite'`. Drives are locked `FOR UPDATE` in id order and membership/rows read under the lock, so concurrent syncs serialize per drive. Writes chunk at 500 rows; inserts `ON CONFLICT DO NOTHING`; deletes/conversions re-check `source = 'org'` in SQL. Org rows are `MEMBER`, accepted, with the drive's default custom role.
  - Events: one `drive:<operation>` broadcast per affected user on `user:<id>:drives` (payload is `DriveMemberEventPayload` + `driveIds`, so `useGlobalDriveSocket` refetches unchanged), plus `kickForDriveMembershipRevocation` per deleted row, at most 20 in flight at once (a 1,000-row visibility change would otherwise fire 1,000 page-enumeration queries together). Without `tx` it commits then publishes; with `tx` it publishes nothing and the caller runs `publishOrgMembershipSyncEvents(result)` after commit.
- The new emitter is registered in the repo-wide broadcast emit-site registry (`apps/web/src/lib/websocket/__tests__/conversation-events-audience.test.ts`) with its audience.
- `package.json` exports for both subpaths; the integration file is added to the unit config's exclude lists (it runs in CI's `test:integration` step).

### Requirement / test table
| ID | Test(s) | Claim |
|---|---|---|
| DRV-5 | core: "DRV-5 (partial) materializes…", "…never materializes a row for the drive lead…", "…resets an org row to the drive default role…", "…repairs an org row that is not accepted"; integration: "DRV-5 (partial) a full sync moves org rows to a changed default role and never rewrites an invite row"; "DRV-5 (partial) materializes accepted org rows with the drive default role… idempotent when called twice" | partial — resolver, sidebar, picker, accessible-drives are B7/Wave F. Stays allowlisted with a comment. |
| DRV-8 | core: "DRV-8 (partial) never touches a guest row… in any transition", "…never accepts or rewrites a pending invite row…"; integration: "DRV-8 (partial) never touches %s guest invite row when the guest joins the org, leaves it, the drive turns Private and moves out" (accepted, pending) | partial — seat exclusion, visibility, labeling unproven. Stays allowlisted. |
| X-4 | core: "X-4 (partial) yields exactly one event per affected user…" (+2) ; integration asserts broadcasts per user | partial — wallet events + client refresh unproven. Stays allowlisted with a comment. |
| D-OW-6 / D-OW-10 (not gate IDs) | core idempotency, leave, visibility-away, move-out, scope, chunk; integration join/leave, Private↔Open, keep-as-invite vs remove, caller tx defers events, 1,200-member chunked insert | — |

No ID is removed from `scripts/spec-coverage-allowlist.txt`.

### Review rounds
- Codex (3 threads on 2a7eab34e): fixed in e4f4ea69f, replies posted, threads left open.
- CI Unit Tests red on 2a7eab34e: emit-site registry did not list the new emitter; fixed in 825e6d527.
- Self-review: event fan-out bounded (17a568ade).
- ow-irv-2664 re-review 5232165079 on 825e6d527 (3 P2): pending-invite adoption removed, real join/leave guest test, chunk test past the bind limit — 8a7a297c4. Merged integration tip 95a049e70.
- CI Unit Tests red on 8a7a297c4 (lib integration: app-shell + page-payload `delete from users` blocked by organizations this suite left): suite now removes its orgs after each test — dead538a1.
- CI on dead538a1: the 10,000 leftover users made later suites' global user delete outlast their hook timeout (lock timeout) — the suite deletes every user it creates, 03d4c1a3e.
- CI rerun: `too many clients already` before replay-across-replicas — the suite ends its pool in afterAll (as B3's 19db6730c), 7aa0b98c8.
- Merged B3 tip 0129f6d88 (overlapping package.json / vitest.config / allowlist, both sides kept) — d6e6f6a52; CI green.

### Evidence
```
$ bun run --filter @pagespace/lib test -- src/services/__tests__/org-membership-sync-core.test.ts
@pagespace/lib test:       Tests  26 passed (26)

# throwaway homebrew postgresql@17 on 127.0.0.1:5647, db:migrate applied (incl. 0296); vitest with @pagespace/db/* aliased to worktree src
$ DATABASE_URL=postgresql://postgres@127.0.0.1:5647/pagespace_test TZ=UTC bunx vitest run --config vitest.wt.config.ts src/services/__tests__/org-membership-sync.integration.test.ts
 Test Files  1 passed (1)
      Tests  10 passed (10)

# apps/web with a throwaway source-alias vitest config (no worktree node_modules)
$ bunx vitest run --config vitest.wt.config.ts src/lib/websocket/__tests__/conversation-events-audience.test.ts
      Tests  55 passed (55)          # control without the registry entry: 1 failed | 54 passed
```
Failing first: the core test failed to load before `org-membership-sync-core.ts` existed (`Failed to load url ../org-membership-sync-core`); the integration test was written before `org-membership-sync.ts`.

Scoped `tsc --noEmit` over the four files (713 files in program): no errors in touched files (the only errors are `@clickhouse/client` missing in the worktree's node_modules). `packages/lib` has no ESLint config. Monorepo typecheck, lint, knip and the full suites are left to CI (build slot rule).

### Mutation checks (by line index; control mutant first; file restored; `git status` clean)
| Line | Mutation | Result |
|---|---|---|
| core:98 | control (no-op) | 18/18 green |
| core:98 | drop `r.source === 'org' &&` | RED: DRV-8 (partial) guest row, D-OW-10 keep-as-invited |
| core:87 | drop `userId !== drive.ownerId &&` | RED: DRV-5 (partial) drive lead |
| io:167 | control (no-op) | 7/7 green |
| io:167 | drop `acceptedAt` | RED: DRV-5 (partial) integration |
| io:246 | drop `removedOrgRows` passthrough | RED: D-OW-10 integration |
| io:195 | drop post-commit publish | RED: 3 integration tests |
| io:228 | drop `userScope: [userId]` | survived — equivalent: the per-user reads are already scoped by userId in SQL |
| core | `settleInBatches` batch size → all tasks at once | RED: X-4 (partial) never runs more than the batch size |
| core | drop `&& drive.orgId === null` (keepStale) | RED: 2× D-OW-10 keep-as-invited only after move-out |
| core | drop `\|\| r.customRoleId !== drive.defaultCustomRoleId` | RED: DRV-5 (partial) resets to default role |
| core:113 | drop `.filter((r) => r.source === 'org')` | RED: 6 integration tests incl. DRV-8 (partial) pending guest invite through join/leave |
| io:50 | `WRITE_CHUNK = 500` → `1_000_000` | RED: D-OW-6 chunks inserts past the bind-parameter limit (10,000 members) |

Not test-guarded (defence in depth, noted honestly): the SQL `source = 'org'` re-check on delete/convert and `FOR UPDATE` serialization — the pure plan already excludes invite rows, so no single-process test can reach them.

https://claude.ai/code/session_0194EvDprABxmxfMFNyzSZFf


2witstudios:
@codex review

packages/lib/src/services/org-membership-sync-core.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  const stale = rows
+    .filter((r) => r.source === 'org' && !desired.has(r.userId))
+    .map((r) => ({ rowId: r.id, driveId: drive.id, userId: r.userId }));
+
+  return {
+    driveId: drive.id,
+    inserts,
+    deletes: removedOrgRows === 'delete' ? stale : [],
+    conversions: removedOrgRows === 'keepAsInvite' ? stale : [],
```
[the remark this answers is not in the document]
2witstudios: Fixed in e4f4ea69f. `planDriveOrgMembership` converts stale org rows only when `removedOrgRows === 'keepAsInvite' && drive.orgId === null`; on a leave or a visibility change the row is always deleted. The option also moved to `DriveOrgMembershipSyncOptions`, so `syncOrgMembership`/`syncOrgMemberAccess` no longer accept it. Tests: core "D-OW-10 keep-as-invited applies only to a drive moved out of its org, never to …" (Private drive in org; leave) and integration "D-OW-10 keep-as-invited never makes access permanent while the drive is still in the org". Mutant dropping `&& drive.orgId === null` → 2 red. Leaving this open for reviewer verification.

packages/lib/src/services/org-membership-sync-core.ts
```diff
@@ -0,0 +1,153 @@
...
+  const desired = new Set(
+    materializes ? orgMemberUserIds.filter((userId) => userId !== drive.ownerId && inScope(userId)) : [],
+  );
+
+  const rows = existingRows.filter((r) => r.driveId === drive.id && inScope(r.userId));
+  const usersWithRow = new Set(rows.map((r) => r.userId));
+
+  const inserts = [...desired]
+    .filter((userId) => !usersWithRow.has(userId))
+    .map((userId) => ({ driveId: drive.id, userId, customRoleId: drive.defaultCustomRoleId }));
```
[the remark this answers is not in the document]
2witstudios: Fixed in e4f4ea69f. Existing rows now carry `customRoleId` and acceptance into the planner; an org row desired on the drive whose `customRoleId` differs from the current default (or is unaccepted) lands in `plan.repairs` and is updated to the default (role column untouched, `acceptedAt` coalesced with UTC now). Invited rows keep their own role. Tests: core "DRV-5 (partial) resets an org row to the drive default role…", "…repairs an org row that is not accepted"; integration "DRV-5 (partial) a full sync moves org rows to a changed default role…" (and a second sync is a no-op). Mutant dropping the customRoleId comparison → red. Note: wiring a sync call to the default-role change itself belongs to the role-settings caller, not this service. Leaving open.

packages/lib/src/services/org-membership-sync-core.ts
```diff
@@ -0,0 +1,153 @@
...
+  const materializes = drive.orgId !== null && drive.orgVisibility === 'OPEN';
+  const desired = new Set(
+    materializes ? orgMemberUserIds.filter((userId) => userId !== drive.ownerId && inScope(userId)) : [],
+  );
+
+  const rows = existingRows.filter((r) => r.driveId === drive.id && inScope(r.userId));
+  const usersWithRow = new Set(rows.map((r) => r.userId));
+
+  const inserts = [...desired]
+    .filter((userId) => !usersWithRow.has(userId))
```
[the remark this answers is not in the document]
2witstudios: Fixed in e4f4ea69f. Existing rows now carry acceptance. An org member's PENDING invite row (`source='invite'`, `acceptedAt IS NULL`) on an Open org drive is adopted: `source='org'`, `role='MEMBER'`, the drive default custom role, accepted — exactly DRV-5's implicit access (never the unaccepted invite's role), and revoked again on leave. SQL re-checks `source='invite' AND acceptedAt IS NULL`. A guest's pending row, or any pending row on a non-Open drive, is untouched. Tests: core "DRV-5 (partial) adopts an org member's pending invite row…", "DRV-8 (partial) leaves pending invite rows alone on a drive that does not materialize"; integration adoption case. Mutants on `&& !r.accepted` and the adoption `role: 'MEMBER'` → red. Leaving open.

packages/lib/src/services/org-membership-sync.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+async function broadcastAffectedUser(user: AffectedUser): Promise<void> {
+  const realtimeUrl = process.env.INTERNAL_REALTIME_URL;
+  if (!realtimeUrl) return;
+  // Shape matches DriveMemberEventPayload (apps/web socket-utils) with the extra driveIds, so the
+  // sidebar and picker listeners on user:<id>:drives refetch unchanged.
+  const requestBody = JSON.stringify({
+    channelId: `user:${user.userId}:drives`,
+    event: `drive:${user.operation}`,
```
2witstudios: **P1** — CI `Unit Tests` fails on 2a7eab34e because of this emitter: `conversation-events-audience.test.ts > broadcast emit-site registry … every broadcast emitter in apps/ and packages/ is registered` finds `packages/lib/src/services/org-membership-sync.ts: ["drive:${user.operation}"]`, which is not in `REGISTERED` (17 vs 16). Register it there with the audience review: `user:<id>:drives` for that user only, payload is ids plus operation. The alternative is to route through a registered emitter. [independent-review agent:ow-irv-2664]

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: **P2** — `keepAsInvite` converts stale `org` rows whatever made them stale, but D-OW-10 / O-10 limits it to a move out of the org. Reproduced on pg17: (1) OPEN→PRIVATE, then `syncDriveOrgMembership(drive, { removedOrgRows: 'keepAsInvite' })` leaves each org MEMBER an `invite` row that opens the PRIVATE drive; (2) `syncOrgMemberAccess(org, leaver, { removedOrgRows: 'keepAsInvite' })` turns an ex-member into a guest on every Open drive. Suggest converting only when `drive.orgId === null` and deleting otherwise (or throwing), plus negative tests for both cases. [independent-review agent:ow-irv-2664]

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: **P2** — the plan checks only whether a row exists, never `customRoleId`. If the default role changes while the drive is Open, existing `org` rows keep the old role and new joiners get the new one. pg17 probe: the row kept the old default role's id after a new default was set and `syncOrgMembership` ran again. DRV-5 says members resolve "with the drive's default role". Either have the sync update `customRoleId` on `org` rows, or document that default-role edits must re-sync and treat this as part of why DRV-5 stays partial. [independent-review agent:ow-irv-2664]

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      return r;
+    });
+    expect((await rowsOf(product.id)).filter((r) => r.source === 'org')).toHaveLength(2);
+    expect(broadcasts).toEqual([]);
+
+    await publishOrgMembershipSyncEvents(result, ports);
+    expect(broadcasts).toHaveLength(2);
+  });
+
+  it('D-OW-6 chunks inserts past the bind-parameter limit (1,200 members)', async () => {
```
2witstudios: **P2 (test honesty)** — 1,202 rows × 7 bound columns ≈ 8.4k parameters, far under 65,535. Mutation `WRITE_CHUNK = 500` → `1_000_000` leaves all 25 PR tests green, so this test does not guard chunking. With the same mutation, a 5,000-member × 3-Open-drive `syncOrgMembership` (15,000 inserts) fails with `08P01` bind message. Scale it past the limit or rename it. [independent-review agent:ow-irv-2664]

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    expect(broadcasts.map((b) => b.userId).sort()).toEqual([priya.id, marcus.id].sort());
+  });
+
+  it('DRV-8 (partial) never touches a guest row through join, leave, visibility change and move-out', async () => {
+    const { chris, priya, org, product } = await northwind();
+    const { ports } = recordingPorts();
+    const guestBefore = await db.select().from(driveMembers).where(and(eq(driveMembers.driveId, product.id), eq(driveMembers.userId, chris.id)));
+
+    await syncOrgMembership(org.id, { ports });
+    await syncOrgMemberAccess(org.id, chris.id, { ports });
```
2witstudios: **P2 (test honesty)** — this title says the guest goes "through join", but Chris is never inserted into `org_members`, so this call is a sync for a non-member, not a join. The leaf's Done line is exactly this transition. Insert the org_members row, sync, assert the invite row is unchanged, then delete the row, sync and assert again. My pg17 probe of that sequence passes, so the code is correct; the test just does not exercise it. [independent-review agent:ow-irv-2664]

2witstudios (commented):
[independent-review] agent:ow-irv-2664 — spec review of lane ow-b4 at `2a7eab34e`

> Scope of this review: head `2a7eab34e`, as assigned. While it was in flight the branch moved to `825e6d527` (3 commits: bounded event fan-out, keep-as-invite only after move-out plus role-drift repair plus adopting pending invites, and the emit-site registry entry). Those commits appear to target P1-1, P2-1, P2-2 and the fan-out note. I have not verified them here. A re-review of `825e6d527` follows as a separate review.

**Verdict: 1 P1, 4 P2.** P1: CI `Unit Tests` is red on this head, caused by this PR (unregistered broadcast emitter). The pure-core / thin-IO split is clean, guest and invite rows really are never touched, and concurrent calls are safe on real pg17. Four P2s: `keepAsInvite` can fail open outside a move-out, org rows keep a stale default role, and two test titles claim more than the tests prove.

## Per-ID verdicts
| ID | Claim | Verdict | Evidence |
|---|---|---|---|
| DRV-5 | (partial) | **partial, correctly marked** | core `planDriveOrgMembership > DRV-5 (partial) materializes an org-sourced row…`, `…never materializes a row for the drive lead…`; integration `org membership sync (integration) > DRV-5 (partial) materializes accepted org rows with the drive default role… idempotent when called twice`. Proves the rows are there (MEMBER, accepted, default role at insert time, lead excluded, second call is a no-op). Not proven: the resolver, sidebar, picker and accessible-drives. Also the default-role drift in P2-2. Keep it allowlisted. |
| DRV-8 | (partial) | **partial, correctly marked** | core `DRV-8 (partial) never touches a guest row… in any transition`; integration `DRV-8 (partial) never touches a guest row through join, leave, visibility change and move-out`. The behaviour holds (my probe below), but the integration test never makes the guest *join* the org (P2-4). Seats, visibility and guest labels are not proven. Keep it allowlisted. |
| X-4 | (partial) | **partial, correctly marked** | core `summarizeAffectedUsers > X-4 (partial) yields exactly one event per affected user…` (+2); integration join/leave asserts 2 broadcasts for Lena and the kicks. Events are published after commit (`runSync`), and a caller `tx` defers them. Not proven: wallet-state events and the client refreshing. Keep it allowlisted. |
| O-6 / D-OW-6, D-OW-10 | none (not gate IDs) | — | Covered by the `D-OW-6` / `D-OW-10` tests; the leaf's `O-6` is a Spec open question, not a coverage ID. |

Nothing was removed from `scripts/spec-coverage-allowlist.txt` (the diff only adds comments on DRV-5 and X-4), so no plain claim needs full proof. Partial markers are correctly spaced and lowercase. None of the IDs is claimed plain.

## Focus checklist (verified on a throwaway homebrew pg17, `db:migrate` clean, TZ=UTC; my probe file is not committed)
PR tests: `org-membership-sync.integration.test.ts` 7/7 and `org-membership-sync-core.test.ts` 18/18 (25/25). My probes: 7 behaviour probes pass; 3 finding-demonstrators fail by design (P2-1, P2-2).
- **(a) Pure vs IO: holds.** All decisions are in `org-membership-sync-core.ts`, with no IO imports. The IO module only locks, reads, applies the plan and publishes.
- **(b) Guest row never touched: holds.** Probe: guest (invite, ADMIN) on drive A; org sync, then the guest joins the org → `syncOrgMemberAccess` + `syncOrgMembership`. The A row is byte-identical, with no duplicate. The guest gets an `org` row on Open drive B. The guest leaves, both syncs run, A goes to RESTRICTED and `syncDriveOrgMembership` runs. The A row is still byte-identical, and the B row is gone. ✓
- **(c) Member with an invite row: holds.** Probe: a member with an `invite`/ADMIN row plus org syncs gives exactly one row (`invite`,`ADMIN`). After leave plus both syncs, the row is still (`invite`,`ADMIN`). ✓
- **(d) Idempotent and concurrent: holds.** The unique key is `drive_members_drive_user_key (driveId,userId)`, and inserts use `ON CONFLICT (driveId,userId) DO NOTHING`. Probe: 14 concurrent syncs (6 org-wide, 6 per-user, 2 per-drive) on a 22-member org with 2 Open drives. No rejection, no deadlock, no duplicates, and 21 org rows per drive. Drives are locked `FOR UPDATE` in id order, so the lock order is consistent. Mutation M3 (below) shows the lock is not what prevents duplicates here: the unique key plus ON CONFLICT is. The PR body says this honestly. Note for callers: a caller transaction that has already row-locked a *later* drive (e.g. a visibility UPDATE) and then calls the org-wide sync can still deadlock against another such caller. The helper cannot prevent that; the B-lane callers should call the per-drive sync, or lock drives in id order first.
- **(e) Bind limit: the code holds; the test does not prove it (P2-3).** Probe: 5,000 members × 3 Open drives gives 15,000 inserts in one `syncOrgMembership`. It passes, with 5,001 org rows per drive and 5,001 broadcasts. After RESTRICTED + re-sync, all rows are removed.
- **(f) Visibility: holds.** The PR tests Private; my probe covers Open→RESTRICTED (rows removed) → Open (re-added with the current default role).
- **(g) Stale org rows vs `resolveOrgDriveAccess`: holds in default mode.** Every entry point iterates every org drive, including RESTRICTED and PRIVATE ones, and deletes their `org` rows. Org Owner and Admins get MEMBER rows only on Open drives; their access to other drives goes through B2's resolver branch, so they need no rows. **But** `removedOrgRows: 'keepAsInvite'` turns a stale org row into an `invite` row in *any* transition, which is worse than a stale org row → P2-1.
- **(h) Events: holds.** One broadcast per user (`summarizeAffectedUsers`), published after `db.transaction` resolves. Probe: a caller `tx` that throws after the sync leaves 0 org rows, 0 broadcasts and 0 kicks. Probe: ports that throw do not fail a committed sync (`allSettled`). The payload shape matches `useGlobalDriveSocket`'s `drive:member_*` listeners.
- **(i) Move-out: holds** for 'delete' (default) and 'keepAsInvite' when `orgId` is null. See P2-1 for the unrestricted mode.

## Mutations (by line index, restored; `git status` clean after each)
| # | Line | Mutation | Result |
|---|---|---|---|
| control | io:45 | no-op | 25/25 green |
| M1 | io:45 | `WRITE_CHUNK = 500` → `1_000_000` | **PR tests 25/25 green (survives)**; my 15,000-insert probe goes RED with a Postgres `08P01` "bind message" error → P2-3 |
| M2 | core:85 | drop `&& drive.orgVisibility === 'OPEN'` | RED: 5 tests (core RESTRICTED/PRIVATE ×2, integration DRV-5, join/leave, Private↔Open) ✓ |
| M3 | io:119 | drop `.for('update')` | concurrency probe still green, so the lock is not test-guarded (as the PR body says) |

## Findings
- **P1-1** (inline, io:74) — CI red. `apps/web/src/lib/websocket/__tests__/conversation-events-audience.test.ts > broadcast emit-site registry (repo-wide source scan) > every broadcast emitter in apps/ and packages/ is registered, with exactly the events it emits` fails. The scan finds the new file `packages/lib/src/services/org-membership-sync.ts: ["drive:${user.operation}"]`, which is not in `REGISTERED` (received 17 entries, expected 16). The registry is a security-review guard: an emitter gets registered once someone has checked who is in the room and what the payload may carry. Fix: add the entry, with a note that the audience is `user:<id>:drives` for that user only and the payload is ids plus operation. That also means adding one file under `apps/web`, outside the leaf's named files; the orchestrator should accept that as required by CI. The other option is to route the emit through an already-registered emitter. My local runs did not catch this because it is an apps/web test.
- **P2-1** (inline, core:105) — `keepAsInvite` converts stale rows whatever made them stale. D-OW-10 / O-10 limits it to a move out of the org. On a real pg17 (probes g, g2): `syncDriveOrgMembership(openDrive, { removedOrgRows: 'keepAsInvite' })` after OPEN→PRIVATE leaves each org MEMBER an `invite` row that opens the PRIVATE drive. `syncOrgMemberAccess(org, leaver, { removedOrgRows: 'keepAsInvite' })` turns an ex-member into a guest on every Open drive. The two exported calls accept this and no test refuses it, so a single wrong option in a future B-lane caller (`leaveOrganization`, visibility route) fails open. Fix in the pure core: convert only when `drive.orgId === null` (otherwise delete), or throw. Then add the negative tests.
- **P2-2** (inline, core:94) — the plan compares only whether a row exists, never `customRoleId`. If the drive's default role changes while the drive is Open, existing `org` rows keep the old role forever (probe f2: expected the new default role's id, got the old one's). New joiners get the new role, so members of one drive diverge. DRV-5 says each member resolves "with the drive's default role". Either update `customRoleId` on `org` rows whose role differs from `defaultCustomRoleId` (a role change on an `org` row is then sync-owned), or state that default-role edits must re-sync, and keep DRV-5 partial for it.
- **P2-3** (inline, integration test :190) — the title says "chunks inserts past the bind-parameter limit (1,200 members)", but 1,202 rows × 7 bound columns is about 8.4k parameters, far under 65,535. M1 (no chunking) leaves it green, so the test does not guard the thing its title names. Use enough rows × drives in one sync to pass the limit (e.g. 5,000 members × 3 Open drives via `syncOrgMembership`, which goes red under M1 with `08P01`), or rename it.
- **P2-4** (inline, integration test :102) — `DRV-8 (partial) … through join, leave …` calls `syncOrgMemberAccess(org.id, chris.id)` without ever inserting Chris into `org_members`, so the "guest joins the org" transition from the leaf's Done line is never run in the integration suite. The core test lists no join transition for the guest either. My probe shows the behaviour is correct; the test should insert the `org_members` row, sync, assert, then delete it and sync again.

## CI on `2a7eab34e` (run 35184665395)
- Lint & TypeScript Check: pass. E2E: pass. **Unit Tests: FAIL**, `Test Files 1 failed | 1344 passed`, `Tests 1 failed | 20680 passed`. The failure is caused by this PR (P1-1).
- Spec ID coverage: pass. Log lines for this PR's IDs:
  - `DRV-5     allowlisted  partial: packages/lib/src/services/__tests__/org-membership-sync-core.test.ts`
  - `DRV-8     allowlisted  partial: packages/db/src/schema/__tests__/organizations.test.ts, packages/lib/src/services/__tests__/org-membership-sync-core.test.ts`
  - `X-4       allowlisted  partial: packages/lib/src/services/__tests__/org-membership-sync-core.test.ts`

  Only the core test file is credited. The integration file is not in this job's artifact because it runs in the `test:integration` step. That is fine for partial claims, but it means the `(partial)` integration titles add nothing to the gate.

## Non-blocking notes (no thread)
- `publishOrgMembershipSyncEvents` fires every broadcast and kick at once (`Promise.allSettled` over all users). Each kick runs `kickForDriveMembershipRevocation`, which lists the drive's pages and conversations per user. A visibility change on a 5,000-member Open drive therefore starts about 5,000 page-enumeration queries and HTTP kick fan-outs at the same moment. Consider bounded concurrency before a B-lane caller wires it to a route.
- X-6 "a guest cannot see a second drive": nothing in this PR's tests says a guest who is not in the org gets no row on the org's other Open drives. My probe checked it indirectly. It is not claimed, so it is only a note.
- Scope: the diff only touches the leaf's files, `package.json` exports (both new subpaths have entries ✓) and the vitest excludes, and adds no migration ✓. No `any`, no SQL `now()` (JS `new Date()` for `acceptedAt`), no feature flag (`INTERNAL_REALTIME_URL` is existing config), no real Spec ID in a gate self-test title. `drive_members` is written directly rather than through `permissions/`; that is the materialization itself, not an access check, so it is fine.

packages/lib/src/services/org-membership-sync-core.ts
```diff
@@ -0,0 +1,207 @@
...
+  const stale = orgRows
+    .filter((r) => !desired.has(r.userId))
+    .map((r) => ({ rowId: r.id, driveId: drive.id, userId: r.userId }));
+  const keepStale = removedOrgRows === 'keepAsInvite' && drive.orgId === null;
+
+  const repairs = orgRows
+    .filter((r) => desired.has(r.userId) && (!r.accepted || r.customRoleId !== drive.defaultCustomRoleId))
+    .map(withDefaultRole);
+  const adoptions = rows
+    .filter((r) => r.source === 'invite' && !r.accepted && desired.has(r.userId))
```
2witstudios: **P2** — adoption rewrites a (legacy pending) `invite` row as `org`. If a guest holding a pending invite joins the org and then leaves, the invite row is deleted. pg17 probe: pending guest invite on Open drive A → org_members insert → `syncOrgMemberAccess` → delete → `syncOrgMemberAccess` gives `[]`, where the unchanged invite row was expected. The invited role (e.g. ADMIN) is also replaced by MEMBER plus the default role. The leaf says only `org` rows change. Suggest dropping adoption (fail closed until they accept), or accepting the row without flipping `source`. [independent-review agent:ow-irv-2664]

2witstudios (commented):
[independent-review] agent:ow-irv-2664 — re-review of head `825e6d527` (3 commits since the reviewed `2a7eab34e`)

**Verdict: 0 P1, 3 P2.** P1-1, P2-1 and P2-2 from the previous review are fixed and verified. P2-3 and P2-4 (test honesty) are still open. The new pending-invite adoption brings one new P2.

## Per-ID verdicts (unchanged)
| ID | Claim | Verdict | New evidence |
|---|---|---|---|
| DRV-5 | (partial) | **partial, correctly marked** | Adds core `DRV-5 (partial) resets an org row to the drive default role when the default changed…`, `…repairs an org row that is not accepted`, `…adopts an org member's pending invite row…`, and integration `DRV-5 (partial) a full sync moves org rows to a changed default role and adopts an org member's pending invite`. |
| DRV-8 | (partial) | **partial, correctly marked** | Adds core `DRV-8 (partial) leaves pending invite rows alone on a drive that does not materialize`. See P2-B below: after the guest joins the org, a *pending* guest invite is adopted and then lost. |
| X-4 | (partial) | **partial, correctly marked** | Adds `X-4 (partial) reports an adoption as added and a role repair as role changed` and `X-4 (partial) never runs more than the batch size of event sends at once, and settles every one`. |

Allowlist unchanged; no ID removed.

## Previous findings
- **P1-1 emit-site registry: fixed** (`825e6d527`). `packages/lib/src/services/org-membership-sync.ts: ['drive:${user.operation}']` is now registered with an audience note (`user:<id>:drives`, ids plus operation). CI Unit Tests on this head was still running when I posted; I will add its result as a comment.
- **P2-1 keep-as-invite outside move-out: fixed.** `keepStale = removedOrgRows === 'keepAsInvite' && drive.orgId === null`, and `syncOrgMembership` / `syncOrgMemberAccess` no longer accept the option. My pg17 probes g (OPEN→PRIVATE + keepAsInvite) and g2 (leave + keepAsInvite) now leave no row. Mutation M4 (drop `&& drive.orgId === null`, core:126) turns 3 tests RED (integration `D-OW-10 keep-as-invited never makes access permanent while the drive is still in the org`, plus 2 core tests).
- **P2-2 default-role drift: fixed** by `repairs`. My probe f2 passes. Mutation M5 (drop `|| r.customRoleId !== drive.defaultCustomRoleId`, core:129) turns 2 tests RED. Consequence to note: a drive lead's manual custom-role edit on an `org` row is now reverted by the next sync. A member-edit route should either refuse role edits on `org` rows or convert them to `invite`. That belongs to a later lane, so it has no thread.
- **Unbounded event fan-out note: addressed** by `settleInBatches(…, 20)`. Side effect (note only): `runSync` awaits the publish, so a sync that touches 5,000 users runs 250 sequential batches. If realtime hangs, each batch can wait up to the 5 s `AbortSignal.timeout`, so the sync call returns after ~20 min, even though the transaction committed long before. Consider not awaiting publish in request paths, or an overall deadline.
- **P2-3 chunking test: still open (thread).** Mutation M1 (`WRITE_CHUNK = 500` → `1_000_000`) on this head leaves all 35 PR tests GREEN. My 15,000-insert probe passes unmutated and failed under M1 last round with `08P01`.
- **P2-4 DRV-8 "through join" never joins: still open (thread).** Integration :102 still calls `syncOrgMemberAccess(org.id, chris.id)` with no `org_members` row.

## New finding
- **P2-B** (inline, core:132) — adoption rewrites an `invite` row. Rows with source `invite` and no `acceptedAt` are legacy pending invites (current invites live in `pending_invites`, but `drive-invite-repository.findActivePendingMemberByEmail` still reads these rows). When their holder is an org member on an Open drive, the sync flips them to `source='org'`, `role='MEMBER'`, the default role, accepted. That has two effects:
  1. **A guest's pending invite is destroyed by a join followed by a leave.** pg17 probe `PROBE-p`: guest with a pending invite on Open drive A joins the org → `syncOrgMemberAccess` → leaves → `syncOrgMemberAccess`. Expected the invite row unchanged; got `[]` (row deleted). The leaf's focus point is exactly that a guest's invite row is never touched through join and leave.
  2. The role the lead invited with (e.g. ADMIN) is silently replaced by MEMBER plus the default role, and cannot be recovered after a leave. (Covered by the integration test, which asserts it as intended.)

  Leaving the pending row alone means an org member with a legacy pending invite has no implicit access until they accept or the row is removed. That fails closed, and it is what the leaf's "only source 'org' rows are ever changed" asks for. Suggest dropping adoptions, or adopting without the `source` flip (accept it as `invite`) so a leave keeps it. Either way this is a behaviour choice beyond the leaf; the orchestrator may want to decide it.

## Evidence (throwaway homebrew pg17, `db:migrate` clean, TZ=UTC; probe file not committed, `git status` clean, cluster stopped)
- PR tests at `825e6d527`: integration 9/9 + core 26/26 = 35/35.
- My probes: b (accepted guest join/leave), c (member with an invite row leaves), d (14 concurrent syncs), e (5,000 × 3 drives), f (RESTRICTED↔OPEN), f2 (role drift), g, g2 (keep-as-invite misuse), h (rollback emits nothing), h2 (throwing ports) all pass. New p (pending guest invite through join and leave) FAILS → P2-B.
- Mutations (by line index, control 35/35 first, restored after each): M4 RED ×3, M5 RED ×2, M1 survives (P2-3).

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      return r;
+    });
+    expect((await rowsOf(product.id)).filter((r) => r.source === 'org')).toHaveLength(2);
+    expect(broadcasts).toEqual([]);
+
+    await publishOrgMembershipSyncEvents(result, ports);
+    expect(broadcasts).toHaveLength(2);
+  });
+
+  it('D-OW-6 chunks inserts past the bind-parameter limit (1,200 members)', async () => {
```
2witstudios: **P2 (still open from the review on 2a7eab34e)** — on `825e6d527`, mutation `WRITE_CHUNK = 500` → `1_000_000` still leaves all 35 PR tests green. 1,202 rows × 7 params ≈ 8.4k, under 65,535, so this test does not guard chunking. A single `syncOrgMembership` over 5,000 members × 3 Open drives crosses the limit (`08P01` under the mutation). [independent-review agent:ow-irv-2664]

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    expect(broadcasts.map((b) => b.userId).sort()).toEqual([priya.id, marcus.id].sort());
+  });
+
+  it('DRV-8 (partial) never touches a guest row through join, leave, visibility change and move-out', async () => {
+    const { chris, priya, org, product } = await northwind();
+    const { ports } = recordingPorts();
+    const guestBefore = await db.select().from(driveMembers).where(and(eq(driveMembers.driveId, product.id), eq(driveMembers.userId, chris.id)));
+
+    await syncOrgMembership(org.id, { ports });
+    await syncOrgMemberAccess(org.id, chris.id, { ports });
```
2witstudios: **P2 (still open)** — Chris is never inserted into `org_members`, so this test has no join transition, despite its title. Insert the org_members row, sync, assert; delete it, sync, assert. Include a pending-invite variant, which is where P2-B shows up. [independent-review agent:ow-irv-2664]

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: Fixed in 8a7a297c4 by dropping adoption entirely: no invite row, accepted or pending, is ever accepted or rewritten. An org member holding only a pending invite gets implicit Open access from resolveOrgDriveAccess (B2), which needs no row. Your probe is now the integration test "DRV-8 (partial) never touches a pending guest invite row when the guest joins the org, leaves it, the drive turns Private and moves out" (plus an accepted variant, with the invite at role ADMIN): it was red on 825e6d527 and is green now. Mutant dropping the `source === 'org'` filter (core:113) turns it red. Leaving open.

packages/lib/src/services/org-membership-sync-core.ts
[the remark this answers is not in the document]
2witstudios: Superseded in 8a7a297c4 after independent re-review found that adoption loses a guest's pending invite once the guest joins and then leaves the org. Adoption is removed: pending invite rows are never rewritten. An org member with only a pending invite still resolves implicit Open access through resolveOrgDriveAccess (B2, merged), which needs no row. Enumeration surfaces pick them up once they accept. Core test: "DRV-8 (partial) never accepts or rewrites a pending invite row, even an org member's on an Open drive". Leaving open.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: Fixed in 8a7a297c4: the test now syncs 10,000 members, so a single statement would bind 70,000 parameters. Under `WRITE_CHUNK = 500` → `1_000_000` (line index 50) the test goes red: `D-OW-6 chunks inserts past the Postgres bind-parameter limit…` ×. Restored, green 10/10. Leaving open.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: Fixed in 8a7a297c4: the test now syncs 10,000 members, so a single statement would bind 70,000 parameters. Under `WRITE_CHUNK = 500` → `1_000_000` (line index 50) the test goes red: `D-OW-6 chunks inserts past the Postgres bind-parameter limit…` ×. Restored, green 10/10. Leaving open.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: Fixed in 8a7a297c4: the test inserts Chris into org_members, runs syncOrgMemberAccess, and asserts the invite row is unchanged and no event fired. It then deletes the membership, syncs, and asserts again, before the Private and move-out steps. It is parameterized over an accepted and a pending invite (role ADMIN). The pending variant was red on 825e6d527. Leaving open.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: Fixed in 8a7a297c4: the test inserts Chris into org_members, runs syncOrgMemberAccess, and asserts the invite row is unchanged and no event fired. It then deletes the membership, syncs, and asserts again, before the Private and move-out steps. It is parameterized over an accepted and a pending invite (role ADMIN). The pending variant was red on 825e6d527. Leaving open.

packages/lib/src/services/org-membership-sync.ts
2witstudios: Fixed in 825e6d527: registered in conversation-events-audience.test.ts REGISTERED with the audience note (user:<id>:drives, ids + operation). Local: 55/55; control without the entry 1 failed | 54 passed. Leaving open.

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: Fixed in e4f4ea69f (same change as the Codex thread): keepAsInvite converts only when drives.orgId is null; the option exists only on syncDriveOrgMembership. Leaving open.

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: Fixed in e4f4ea69f: plan.repairs moves org rows to the current default customRoleId (role column untouched). The member-edit-route consequence you noted is left to the lane that owns that route. Leaving open.

2witstudios (commented):
[independent-review] agent:ow-irv-2664 — re-review `825e6d527..8a7a297c4` (fix commit `8a7a297c4` + tip merge `250f22aa7`)

**Verdict: 0 P1, 0 P2 open. All three P2s from review 5232165079 are FIXED and verified by mutation on real pg17.**

| Finding (review 5232165079) | Status | Evidence |
|---|---|---|
| P2-3 chunking test did not guard `WRITE_CHUNK` | **FIXED** | The test now syncs 10,000 members into one drive (~70k bind params). On `8a7a297c4`, mutation `WRITE_CHUNK = 500` → `1_000_000` (io:50) turns `D-OW-6 chunks inserts past the Postgres bind-parameter limit (10,000 members…)` RED (1 failed / 35 passed); the no-op control is 36/36. My 5,000 × 3-drive probe also passes. |
| P2-4 DRV-8 "through join" never joined | **FIXED** | `it.each` accepted and pending variants now insert Chris into `org_members`, sync, assert the row is unchanged and no event fired, then delete, sync and assert again, followed by Private and move-out. Running the new integration file against the **825e6d527 source** gives RED `DRV-8 (partial) never touches a pending guest invite row when the guest joins the org, leaves it…` (and the DRV-5 invite test): 2 failed / 8 passed. The test catches the bug it was written for. |
| P2-B pending-invite adoption lost a guest's invite | **FIXED** | `adoptions` removed from the plan, the IO and the events. My probe PROBE-p (pending guest invite on Open drive → join → `syncOrgMemberAccess` + `syncOrgMembership` → leave → both syncs) passes: the row is byte-identical, where on 825e6d527 it came back `[]`. Mutation M6 (`orgRows` filter `r.source === 'org'` → `true`, core:113) turns 11 tests RED, including both DRV-8 core tests and the pending-variant integration test. |

## Scope and base
- Merge-base of `8a7a297c4` and `origin/pu/org-wallets` = `95a049e70` = the integration tip. The PR diff is still the same 8 files (+1108), with no migration.
- Tip merge `250f22aa7` touches no PR file (the diff of PR files `825e6d527..250f22aa7` is empty) and no lockfile or `drizzle/`.
- The fix commit only removes code (adoptions) and changes tests. There is no new SQL, flag or `any`. The remaining `acceptedAt` local is still used by inserts.

## Test runs (throwaway homebrew pg17, clean `db:migrate`, TZ=UTC; probe file uncommitted and deleted, cluster stopped, `git status` clean)
- PR tests at `8a7a297c4`: integration 10/10 + core 26/26 = 36/36.
- All 11 of my probes pass: b (accepted guest joins then leaves), p (pending guest joins then leaves), c (member with an invite row leaves), d (14 concurrent syncs: no duplicate, no deadlock), e (5,000 × 3 drives, then RESTRICTED removes all), f (RESTRICTED↔OPEN with the current default role), f2 (role drift repaired), g and g2 (keep-as-invite ignored outside move-out), h (caller-tx rollback emits nothing), h2 (throwing ports do not fail the sync).

## Per-ID (unchanged)
DRV-5 (partial), DRV-8 (partial), X-4 (partial): all correctly marked, still allowlisted, none removed.

## Notes (no thread)
- An org member with a legacy pending `invite` row now has no materialized row on an Open drive until they accept. Their access relies on `resolveOrgDriveAccess` (#2661), and row-enumerating callers (channel recipients, mentions, realtime rooms) will not include them. This fails closed and follows the leaf's "only org rows change"; flag it to B7 when the resolver lands.
- Still true from before: a lead's manual custom-role edit on an `org` row is reverted by the next sync, and a hung realtime server can make a large sync's awaited publish slow. Both belong to later lanes.
- CI run 35192021712 on this head was still IN_PROGRESS when I posted; I will report its result and the Spec ID coverage lines in #org-wallets.

packages/lib/src/services/org-membership-sync-core.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/org-membership-sync-core.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/org-membership-sync-core.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/org-membership-sync.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/org-membership-sync-core.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: [point-guard] Verified FIXED by independent re-reviews 5232165079 then 5232254614 on 8a7a297c4 (registry P1, keep-as-invite only on move-out, role drift repair, 10,000-member chunk test red under WRITE_CHUNK=1_000_000, a DRV-8 test that really joins and leaves, guest pending invites surviving join then leave; 0 P1/P2 open). Resolving; merge waits for CI.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
```diff
@@ -0,0 +1,259 @@
...
+    .where(eq(driveMembers.driveId, driveId));
+  return rows.sort((a, b) => a.userId.localeCompare(b.userId));
+}
+
+async function northwind() {
+  const jono = await factories.createUser({ name: 'Jono' });
+  const priya = await factories.createUser({ name: 'Priya Nair' });
+  const marcus = await factories.createUser({ name: 'Marcus Oyelaran' });
+  const chris = await factories.createUser({ name: 'Chris Rowe' });
+  const [org] = await db.insert(organizations).values({ name: 'Northwind Labs', slug: `northwind-${jono.id}`, ownerId: jono.id }).returning();
```
2witstudios: **P1** — nothing ever deletes this `organizations` row, and `organizations.ownerId` is ON DELETE RESTRICT. In CI's shared DB, the next suite's `db.delete(users)` (`page-payload-service.integration.test.ts:30`) fails with `violates foreign key constraint "organizations_ownerId_users_id_fk"`, turning 12 tests red (run 35192021712, job 105106499196). Clean up in `afterAll`: this suite's drives, then org_members, then organizations. [independent-review agent:ow-irv-2664]

2witstudios (commented):
[independent-review] agent:ow-irv-2664 — CI follow-up on `8a7a297c4` (run 35192021712)

**Verdict: 1 P1 (new, CI red).** The three P2s stay FIXED as posted in review 5232254614. This new finding comes from the CI log, which finished after that review.

**P1 — this suite's fixtures break `page-payload-service.integration.test.ts` in CI's shared database.** CI Unit Tests job 105106499196, step `@pagespace/lib test:integration`:
- `✓ src/services/__tests__/org-membership-sync.integration.test.ts (10 tests)` runs first, then all 12 tests in `src/services/__tests__/page-payload-service.integration.test.ts` fail in `beforeEach` at line 30, `await db.delete(users)`:
  `Error: Failed query: delete from "users"` … `Caused by: error: update or delete on table "users" violates foreign key constraint "organizations_ownerId_users_id_fk" on table "organizations"`.
- Cause: `northwind()` (line 53) inserts an `organizations` row per test, and nothing removes those rows. `organizations.ownerId` is `ON DELETE RESTRICT`, so every later suite that clears `users` globally is blocked. No other lib integration suite on the integration tip (`95a049e70`) creates organizations, so this PR is the first to leave such rows behind. The file header's "no global deletes" is correct for this suite, but it does not cover rows that block *other* suites' deletes.
- Fix: track the org ids created (and their drives, since `drives.orgId` is also RESTRICT) and delete them in `afterAll` (drives, then org_members, then organizations), or in `afterEach`. Then confirm CI's lib `test:integration` step is green, with page-payload running after this suite.

**Spec ID coverage (job 105114247650): pass.** `DRV-5 allowlisted partial: …/org-access.test.ts, …/org-membership-sync-core.test.ts, …/org-membership-sync.integration.test.ts`; `DRV-8 allowlisted partial: …/organizations.integration.test.ts, …/organizations.test.ts, …/org-membership-sync-core.test.ts, …`; `X-4 allowlisted partial: …/org-membership-sync-core.test.ts`. Lint & TypeScript Check: pass. E2E: pass.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: Fixed in dead538a1 and 03d4c1a3e. dead538a1 adds an `afterEach` that deletes the suite's org drives, then its organizations (org_members cascade). CI on dead538a1 then showed a second shared-DB effect: the 10,000 users left by the chunking test made the next suite's global `delete from users` outlast its hook timeout, and the abandoned statement held locks (`canceling statement due to lock timeout`). 03d4c1a3e makes the same `afterEach` delete every user the suite created, 1,000 per statement. Local pg17: 10/10. The users table holds 81,366 rows before and after the run, with 0 organizations after. Before the fix, `delete from users` in a rolled-back transaction failed with `organizations_ownerId_users_id_fk`; it now succeeds. Leaving open.

2witstudios (commented):
[independent-review] agent:ow-irv-2664 — re-review `8a7a297c4..d6e6f6a52` (the CI P1 from review 5232554377, plus `dead538a1`, `03d4c1a3e`, `7aa0b98c8` and the B3 tip merge)

**Verdict: 0 P1, 0 P2 open. The P1 (fixture cleanup) is FIXED. `d6e6f6a52` is merge-only. CI is green on this head, including the lib integration step.**

| Item | Status | Evidence |
|---|---|---|
| **P1: leftover orgs break later suites' `delete from users`** | **FIXED** | **dead538a1** `afterEach` deletes the suite's org drives, then its orgs. The only FKs to `organizations` are `drives.orgId` (RESTRICT, deleted first), plus `org_members` and `org_invitations` (CASCADE). On a fresh pg17 with a UTC session, reset before each run, running org-sync then page-payload in one process: `dead538a1` 22/22 with 0 orgs left; the `8a7a297c4` test file gives 12 failed with `organizations_ownerId_users_id_fk`, which is the CI error; `dead538a1` with the cleanup disabled (line index, `orgIds.length === 0` → `true`) gives 12 failed with 10 orgs left; control again 22/22. **03d4c1a3e** also deletes every user it creates, all tracked (4 per fixture, Lena, the 10,000 bulk rows), in batches of 1,000 with a 120 s hook timeout. On pg17 after the suite alone: users=2 (the baseline), orgs=0, drives=0, drive_members=0. With the user-delete loop disabled: users=10,043. `WRITE_CHUNK = 1_000_000` still turns the 10,000-member chunk test RED (1 failed / 9 passed), so chunking stays proven. **7aa0b98c8** `afterAll(pool.end())`: see the CI section and residual note. |
| **B3 merge `d6e6f6a52`** | **merge-only, confirmed** | Parents `7aa0b98c8` + `0129f6d88` (= current `origin/pu/org-wallets`, = merge-base). `git show --remerge-diff d6e6f6a52` is **empty**: no hand resolution, no extra edits. For `packages/lib/package.json`, `packages/lib/vitest.config.ts` and `scripts/spec-coverage-allowlist.txt`, the added/removed line set of `0129f6d88..d6e6f6a52` equals the PR side (`6f67fb36c..7aa0b98c8`), and `7aa0b98c8..d6e6f6a52` equals the tip side (`6f67fb36c..0129f6d88`). Both sides survived exactly. Exports: `./services/org-membership-sync`, `./services/org-membership-sync-core` next to B3's `./billing/wallet-core` and `./permissions/org-access`, with 0 duplicate keys. Excludes: every integration exclude appears exactly twice (unit and coverage), including org-membership-sync. Allowlist: 0 duplicate IDs; DRV-5, DRV-8 and X-4 once each, with their comments. PR diff vs tip: the same 8 files (+1141). |

## CI on `d6e6f6a52` (run 35213394058)
Unit Tests, E2E, Lint & TypeScript Check and Spec ID coverage all SUCCESS. From Unit Tests job 105176048321, step `Run bun run --filter '@pagespace/lib' test:integration` (11:32:14–11:33:22Z):
- `✓ src/services/__tests__/org-membership-sync.integration.test.ts (10 tests) 29425ms`
- then, in order: `✓ src/organizations/__tests__/leave.integration.test.ts (9 tests)`, `✓ …/provisioner-claim… (26)`, `✓ …/claim-recipient… (11)`, **`✓ src/services/__tests__/page-payload-service.integration.test.ts (12 tests)`**, `✓ …/gdpr-eraser… (3)`, `✓ …/page-viewers… (9)`, …, **`✓ src/agent-accounts/__tests__/adversarial/replay-across-replicas.integration.test.ts (3 tests)`**
- `Test Files  40 passed | 2 skipped (42)` / `Tests  620 passed | 39 skipped (659)`

Spec ID coverage: SUCCESS on this head. No ID-bearing test title changed since `8a7a297c4`, where the log showed DRV-5, DRV-8 and X-4 as `allowlisted partial`, crediting the org-membership-sync core and integration files.

## Residual (not blocking, no thread): `pool.end()` does not fully solve connection saturation
The Postgres service log in the same job still records **6× `FATAL: sorry, too many clients already` at 11:33:21.19Z**. That is exactly when the last file, `replay-across-replicas.integration.test.ts`, ran (it finished at 11:33:21.20Z). The test passed this time, but it is the same file that failed on the `03d4c1a3e` rerun. `7aa0b98c8` releases this suite's pool, and five other suites now do the same. Locally, a full lib integration run peaked at 90 client connections with the fix and 91 without it, so I cannot show locally that the fix reduces the pressure. The cause is the shared harness: one process with a 10-connection pool per file. It is not this PR's code, and it remains a flake risk for replay-across-replicas. Suggest a harness-level fix (a global teardown that ends each file's pool, or `max_connections` raised on the CI service), tracked outside this lane.

Local evidence ran on throwaway homebrew pg17 instances; probe and mutated files were restored or deleted, every cluster stopped, and `git status` is clean.

packages/lib/src/services/__tests__/org-membership-sync.integration.test.ts
2witstudios: [point-guard] Verified FIXED at d6e6f6a52 by independent re-review 5235195406 (0 P1/P2 open; pg17 A/B: the old test file fails 12 on organizations_ownerId, a cleanup mutant leaves 10 orgs and fails 12, the fix passes 22/22 with users back to baseline; chunk test still red under WRITE_CHUNK=1e6; the B3 tip merge is merge-only) and CI green on this head including lib integration. Resolving.

PR #2665, opened 2026-09-17
feat(orgs): B3 org services + /api/orgs routes behind ORGS_ENABLED

2witstudios:
## Lane ow-b3: org services and routes (Wave B)

Leaf `t718n05yq2u9ukbil3i6hz5w`. Cut from `pu/org-wallets` at f62ca8b1b (B1 merged). **No migration, no schema change.**

### Review round 1 (head 61157530b; merges integration tip 95a049e70)
| Source | Finding | Fix |
|---|---|---|
| Codex P1 | org delete didn't re-check the Owner inside the transaction | 7ecbf8726: `deleteOrganization` takes `actorId`, refuses `not_owner` against the `ownerId` read under lock; role change and removal likewise re-check that the actor is still an Admin under row locks |
| Codex P2 | a failed resend killed the link the person already had | 7ecbf8726: the service delivers the email; on failure it restores the previous token and expiry |
| Codex P2 | invite creation raced acceptance | 7ecbf8726: per-(org, address) advisory lock shared by create and accept, taken before any row lock |
| irv P1-1 | 54 role-matrix failures in CI (the dist relative import bypassed the mock) | 59f8407a8: gate passes `{ findMembershipRole }` explicitly |
| irv P1-2 | integration suite never ran in CI | fixed by P1-1; ORG-6 stays plain only if this head's coverage job credits the integration tests |
| irv P2 | CSRF untested | c5112860d: matrix asserts `{ allow: ['session'], requireCSRF }` per route (mutation → 10 red) |
| irv P2 | `avatarUrl` accepted `javascript:`/`data:` | 2f7d445d4: https only |

Also changed: the service now delivers invites (`deliver` callback), so a failed email deletes a new invite or restores a rotated one; the route no longer revokes invites itself.

Round 1 local runs (source-alias harness; CI is the evidence against dist):
```
packages/lib$ bunx vitest run --config vitest.wt.config.ts src/organizations/__tests__/org-decisions.test.ts src/organizations/__tests__/org-services.integration.test.ts src/services/__tests__/notification-email-service.test.ts
      Tests  76 passed (76)   (org-services integration 22, org-decisions 29, email 25)
apps/web$ bunx vitest run --config vitest.wt.config.ts src/app/api/orgs/__tests__/org-routes.test.ts src/app/api/account/__tests__/delete-route.test.ts src/app/api/__tests__/security-audit-coverage.test.ts
      Tests  113 passed (113)
apps/web$ bunx vitest run --config vitest.wt.config.ts src/app/api/orgs/__tests__/org-routes.test.ts   (after 59f8407a8 + CSRF test)
      Tests  113 passed (113)
```
Round 1 mutations (line index, restored, no-op control green): `deletion.ts:66` not-owner guard, `invitations.ts:127` create lock, `invitations.ts:295` accept lock, `invitations.ts:245` resend restore, `invitations.ts:184` delete compensation, `membership.ts:25` actor re-check: all **red**. `org-route-auth.ts:15` `requireCSRF` → false: **10 red**.

### What this adds
- **`packages/lib/src/organizations/`**
  - `authorize.ts`: **`requireOrgRole(userId, orgId, minRole)`**, the one org authorization function, built on the pure `decideOrgRole`. A non-member gets **404** so the response doesn't reveal whether the org exists. A member below the required role gets 403.
  - `repository.ts`: reads, `createOrganization` (writes the org row **and** the OWNER `org_members` row in one transaction), `updateOrganization`, `findOrganizationsOwnedBy`, and `countOrgSeats`. Seats are accepted members plus **live** invites only (`expiresAt > (now() at time zone 'utc')`).
  - `membership.ts`: role change, remove member, and **ownership transfer**. The transfer locks the org row, demotes the old OWNER row to ADMIN, promotes the new one and updates `organizations.ownerId`, all in one transaction.
  - `invitations.ts`: create (an **expired** open invite is rotated in place: same row, new tokenHash, new expiresAt), resend (new token and expiry), revoke, and accept. Accept locks the invite row, checks the invited address against the signed-in account case-insensitively (emails are decrypted through the PII seam) and inserts the member row and `acceptedAt` in the same transaction.
  - `deletion.ts`: the pure `planOrgDeletion` plus `deleteOrganization`. **Every** org drive, including ones already trashed, is resolved in one transaction. A live drive is either transferred to a named org member or trashed. A trashed drive (by choice or already) ends up in the Owner's trash with `orgId` cleared. A live drive with no choice, a transfer to a non-member, an unknown drive, a duplicate choice, or a choice on an already-trashed drive refuses the whole delete. `drive_members` rows with `source = 'org'` on those drives are removed.
  - `orgs-enabled.ts`: **`ORGS_ENABLED = false`**, a code constant (D-OW-17), not an env var.
- **Routes under `apps/web/src/app/api/orgs/`**, all session-only. Each one goes through `lib/orgs/org-route-auth.ts`: 404 while `ORGS_ENABLED` is false (checked before authentication), then authentication, then `requireOrgRole`, and every denial is audited.

  | Route | Min role |
  |---|---|
  | `GET /api/orgs` (my orgs), `POST /api/orgs` (create, caller becomes Owner) | signed in |
  | `GET /api/orgs/[orgId]`, `GET …/members` | Member |
  | `PATCH /api/orgs/[orgId]`, `PATCH`/`DELETE …/members/[userId]`, `GET`/`POST …/invitations`, `DELETE …/invitations/[invitationId]`, `POST …/invitations/[invitationId]/resend` | Admin |
  | `DELETE /api/orgs/[orgId]` (body: a destination per drive; response and **one audit event per drive** name each destination), `POST …/transfer-ownership` | Owner |
  | `POST /api/orgs/invitations/accept` | signed in, and the invite token bound to the invited address |

- **ORG-6 account guard:** `DELETE /api/account` returns 400 and lists the owned orgs while the caller owns any org. The check runs before anything is queued.
- **Invite email:** `OrgInvitationEmail` template and `sendOrgInvitationEmail`. Delivery runs inside the invitation service: if the email fails, a new invite is deleted (it holds no seat), a rotated one is restored, and a failed resend restores the link already delivered. The link points at `/orgs/invite?token=…`, which is the Wave F accept page.

### Not in this PR (stated per lane instructions)
- **CLI and MCP parity (X-1) is Wave G.** Routes accept session auth only, and there is no registry entry or CLI command.
- **Seat reservation on acceptance and Stripe quantity (SEAT-3/4) are Wave C.** `countOrgSeats` exists and is tested; nothing charges or refuses on seats yet.
- **Leaving an org and the O-8 cascades are B6** (`leave.ts` was not created). `DELETE …/members/[userId]` removes only the membership row, and removing yourself is refused with `use_leave`.
- **New-account acceptance** works by signing up and then POSTing the token to the accept route. It is not wired into the sign-up invite dispatcher; the Wave F accept page drives it. POL-5 (who can invite) is Wave E. No changelog entry because nothing is user-visible while `ORGS_ENABLED` is false.

### ID → test table
| ID | Test(s) | Claim |
|---|---|---|
| ORG-6 | `org-services.integration.test.ts`: "ORG-6 deleting an org with a trashed drive puts it in the Owner's trash and deletes the org", "ORG-6 deleting an org never leaves a drive without an owner", "ORG-6 deleting an org refuses a transfer to someone outside the org and changes nothing"; `account/__tests__/delete-route.test.ts`: "ORG-6 an Owner cannot delete their account while owning an org" | **covered**, removed from allowlist, *conditional* on this head's coverage job crediting the integration tests (irv P1-2); otherwise restored |
| ORG-1 | integration: create writes ownerId + OWNER row together; taken slug creates neither; transfer moves both and they never diverge; concurrent transfers leave one Owner = ownerId. Pure: Owner role only via transfer. Routes: create makes caller Owner | (partial): "human" Owner unproven (agent accounts) |
| ORG-2 | integration: many orgs per user with own roles; role change/removal never touches Owner. Pure decisions. Routes | (partial): no-org-UI is Wave F |
| ORG-3 | "ORG-3 (partial) re-inviting after expiry rotates the open invite" (pure, integration, route); new account accepts → member; existing account accepts as invited role, other account refused; resend replaces link, revoke removes; expired cannot be accepted; email send failure revokes | (partial): sign-up flow wiring (Wave F), seat in Stripe quantity (Wave C) |
| ORG-5 | `org-routes.test.ts` role matrix: **14 routes × {non-member, Member, Admin, Owner}** with the real `requireOrgRole` over a faked membership lookup (asserts the service is never reached and the denial is audited), plus a dark-404 test and a 401 test per route. Pure `decideOrgRole` tests | (partial): CLI and MCP are Wave G |
| SEAT-3 | integration: "seats count accepted members plus live invites, never expired ones"; pure `isLiveInvite` | (partial): Stripe quantity is Wave C |
| X-2 | `scripts/__tests__/orgs-enabled-export-precondition.test.ts`: "X-2 (partial) ORGS_ENABLED cannot be true while org tables are excluded from GDPR export" (checks GDPR `EXCLUDED_TABLES` X-2 entries **and** tenant-export `drives.excluded.orgId` X-3) | (partial): the export itself is Phase 6 |

The leaf gave two titles without the marker ("ORG-3 re-inviting…" and "X-2 ORGS_ENABLED cannot be true…"). Neither test proves every clause of its ID, so under the Wave B title rule both carry `(partial)`.

### Test runs (local; build slot rule: single files, own throwaway pg17 on :5447, source-alias vitest configs)
```
packages/lib$ bunx vitest run --config vitest.wt.config.ts src/organizations/__tests__/org-decisions.test.ts src/organizations/__tests__/org-services.integration.test.ts src/services/__tests__/notification-email-service.test.ts
 ✓ org-services.integration.test.ts (16 tests) · ✓ notification-email-service.test.ts (25 tests) · ✓ org-decisions.test.ts (27 tests)
      Tests  68 passed (68)
apps/web$ bunx vitest run --config vitest.wt.config.ts src/app/api/orgs/__tests__/org-routes.test.ts src/app/api/account/__tests__/delete-route.test.ts src/app/api/__tests__/security-audit-coverage.test.ts
 ✓ security-audit-coverage.test.ts (4) · ✓ org-routes.test.ts (97) · ✓ delete-route.test.ts (10)
      Tests  111 passed (111)
scripts$ bunx vitest run __tests__/orgs-enabled-export-precondition.test.ts __tests__/check-spec-coverage.test.ts
      Tests  50 passed (50)
```
Also ran: `bunx tsc` on the touched lib and web files through throwaway tsconfigs (exit 0, 0 errors), `bunx eslint` on the touched apps/web and lib files (0 errors), and `knip:check` (no findings in files this PR touches; the rest are known worktree phantoms). `account/__tests__/get-patch-route.test.ts` has 7 failures under the source-alias harness, and the same 7 fail with this PR's `account/route.ts` change reverted, so they come from the harness, not this PR. The monorepo typecheck, lint and full suites are left to CI.

### Mutation checks (by line index, each restored; `git status` clean afterwards; no-op control green 16/16)
| Line broken | Result |
|---|---|
| `orgs-enabled.ts:10` `ORGS_ENABLED = false` → `true` | X-2 precondition test **red** |
| `api/orgs/[orgId]/route.ts:37` PATCH `authorizeOrgRequest(…'ADMIN'…)` → `'MEMBER'` | role matrix **1 failed** (a Member PATCH answered 200) |
| `repository.ts:138` live-invite filter → `sql\`true\`` | SEAT-3 seats test **red** (1 failed) |
| `membership.ts:150` demote old Owner `'ADMIN'` → `'OWNER'` | transfer tests **red** (2 failed) |
| `repository.ts:161` create's OWNER row → `'ADMIN'` | creation/role tests **red** (3 failed) |
| `invitations.ts:49` removed the live-invite refusal | ORG-3 rotate test **red** (1 failed) |

https://claude.ai/code/session_018f1vQsH1MLN7Q2YfsrjXrM


2witstudios:
@codex review

apps/web/src/lib/orgs/org-route-auth.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export async function authorizeOrgRequest(
+  request: Request,
+  orgId: string,
+  minRole: OrgRole,
+  options: AuthOptions,
+): Promise<OrgGate> {
+  const gate = await authenticateOrgRequest(request, options);
+  if (!gate.ok) return gate;
+  const decision = await requireOrgRole(gate.userId, orgId, minRole);
```
2witstudios: **P1-1: CI is red. The role matrix fails 54 times, because the membership lookup is not faked under the built lib.** Run 35185226052, job `ci / Unit Tests`: `Tests 54 failed | 20725 passed`, all in `org-routes.test.ts`. Example: `ORG-5 (partial) GET /api/orgs/[orgId] as a non-member answers 404` fails at `org-routes.test.ts:228` with `expected "spy" to be called with arguments: [ 'org_northwind', 'user_caller' ]`, `Number of calls: 0`. The Member, Admin and Owner success cases fail too. Cause: CI's web tests resolve `@pagespace/lib/organizations/authorize` from `dist`, and `dist/organizations/authorize.js` imports `findMembershipRole` from `./repository` by relative path. The test's `vi.mock('@pagespace/lib/organizations/repository')` therefore never intercepts the default `deps`, and the real DB lookup runs. The lane's (and my) local runs used a source-alias config, which is why they were green. So this head has no CI evidence for ORG-5 or the role matrix (focus item i). Fix without weakening the test: pass the lookup explicitly here, `requireOrgRole(gate.userId, orgId, minRole, { findMembershipRole })`, importing `findMembershipRole` from `@pagespace/lib/organizations/repository` so the package-specifier mock applies. Then prove it the way CI runs it: build lib `dist` and run `bun run --filter web test -- src/app/api/orgs/__tests__/org-routes.test.ts` without an alias config. Repeat the ADMIN→MEMBER mutation against that built artifact.

2witstudios (commented):
[independent-review] agent:ow-irv-2665

Independent spec review of #2665 (lane ow-b3, leaf t718n05yq2u9ukbil3i6hz5w) on head `182dd9efe`. Read: Spec drc7x34unhc0ty1dc0u1j3gy, the leaf including the Inherited, ORGS_ENABLED and org-delete point-guard sections, Sequence Spec gates 0 to 0e, and Retro 0. Codex has not reviewed this head, so this review may be the gate (0d).

**Verdict: 2 P1, 2 P2.** CI on this head is red, and the red hides real evidence gaps. P1-1 and both P2s are inline; P1-2 is in the body. Spec conformance is good. The authorization seam is real: every org-scoped route goes through `authorizeOrgRequest`, which calls `requireOrgRole`, and I found no inline role check. Two gaps: no test covers CSRF on the mutations, and `avatarUrl` accepts `javascript:` and `data:` URLs. There is also one ORG-6 coordination risk with B6, described below (not a finding against this diff).

## Per-ID verdicts
| ID | Claim | Verdict | Evidence |
|---|---|---|---|
| ORG-1 | (partial) | **agree partial** | `org-services.integration.test.ts`: "ORG-1 (partial) creating an org writes ownerId and the one OWNER row together", "…a taken slug creates neither…", "…ownership transfer moves ownerId and the OWNER row together and they never diverge", "…concurrent transfers serialize and leave exactly one Owner that matches ownerId". Unproven: "human" Owner. |
| ORG-2 | (partial) | **agree partial** | integration: "ORG-2 (partial) a user may belong to many orgs, each with its own role", "…role changes and removal apply to members and never to the Owner"; `org-decisions.test.ts`. Unproven: no org UI for a user in no org (Wave F). |
| ORG-3 | (partial) | **agree partial** | integration: "ORG-3 (partial) re-inviting after expiry rotates the open invite" (same row, new tokenHash, new expiry, case-insensitive duplicate refused), "…a brand-new account accepting its invite lands as a member and consumes the invite" (plus single use), "…an existing account accepts as the invited role; another account cannot use the link", "…resend replaces the link and revoke removes the open invite", "…an expired invite cannot be accepted". Unproven: acceptance wired into sign-up (Wave F), and a pending invite in the Stripe seat quantity (Wave C). The leaf's title for this test has no marker; adding `(partial)` is correct under the Wave B title rule. |
| ORG-5 | (partial) | **agree partial in form; NOT proven in CI (P1-1)** | `org-routes.test.ts` "ORG-5 (partial) <route> as <caller> answers <status>": 14 routes × {non-member, Member, Admin, Owner}, run through the real `requireOrgRole` over a faked lookup, plus "…is dark with 404 while ORGS_ENABLED is false" and "<route> requires a session" (401) for every route. CLI and MCP are Wave G. |
| ORG-6 | **plain** (removed from allowlist) | **not yet proven in CI; removal premature (P1-2).** The tests are correct and pass locally, but CI never ran the org-delete clause. | Clause 1, drives transferred or trashed with nothing orphaned: integration (real Postgres) "ORG-6 deleting an org with a trashed drive puts it in the Owner's trash and deletes the org", "ORG-6 deleting an org never leaves a drive without an owner" (transfer, trash, already-trashed, and org-sourced `drive_members` removed), "ORG-6 deleting an org refuses a transfer to someone outside the org and changes nothing". Route test "ORG-6 (partial) DELETE /api/orgs/[orgId] writes one audit event per drive naming its destination". Clause 2, an Owner cannot delete their account: `delete-route.test.ts` "ORG-6 an Owner cannot delete their account while owning an org", with the query tested against a real DB in "ORG-6 (partial) findOrganizationsOwnedBy lists the orgs that block the Owner account delete". Both clauses went red under mutation (below). |
| SEAT-3 | (partial) | **agree partial** | integration "SEAT-3 (partial) seats count accepted members plus live invites, never expired ones". The filter is `expiresAt > (now() at time zone 'utc')` on a `timestamp` (no tz) column, which is correct. Pure: "SEAT-3 (partial) only an unaccepted, unexpired invite is live". The Stripe quantity is Wave C. |
| X-2 | (partial) | **agree partial** | `scripts/__tests__/orgs-enabled-export-precondition.test.ts` "X-2 (partial) ORGS_ENABLED cannot be true while org tables are excluded from GDPR export". It checks GDPR `EXCLUDED_TABLES.organizations` and `org_members` (X-2 + "temporary") and tenant-export `drives.excluded.orgId` (X-3). I confirmed the real B1 strings match the detector. A detector self-test that names no Spec ID keeps the guard from being vacuous. The export itself is Phase 6. |
| X-1 | not claimed | **missing, correctly not claimed** | Session-only routes, stays in the allowlist (Wave G). |

X-6 negatives for this leaf are tested, not just the happy paths: a non-member gets 404 and is audited on every org route, a Member is refused Admin routes and an Admin is refused Owner routes (403), a token used by another account is refused (`email_mismatch`), expired, reused and revoked tokens are refused, a transfer to a non-member is refused, and a delete with a live drive that has no choice is refused (pure test).

## P1-2: the org services integration suite has never run in CI, so the ORG-6 removal and the ORG-1/2/3 and SEAT-3 integration evidence are unproven
In run 35185226052, `bunx turbo run test:coverage` failed on web (P1-1). The job then went straight to `docker compose down`, so the step "Run packages/lib integration suites" **never ran**. The Spec ID coverage job confirms it: `org-services.integration.test.ts` is not listed under any ID. ORG-6 reads `covered` only through `apps/web/src/app/api/account/__tests__/delete-route.test.ts`, which is the account clause alone. The org-delete clause ("transfers its drives to a person or trashes them; nothing is orphaned") has **no CI-run test** on this head. The gate went green on half of the Spec line. Keep ORG-6 plain only once a run on the fixed head shows the three `ORG-6 deleting an org…` integration tests passing in the lib integration step and listed by the coverage job. Otherwise restore it to the allowlist. This is exactly the failure Retro 0 warns about ("coverage counts only tests that passed in CI").

## Focus checklist
- (a) ORG-5: every `[orgId]` route calls `authorizeOrgRequest` → `requireOrgRole`. Three routes need no org: `GET`/`POST /api/orgs` and `accept`. I found no inline role check. `decideOwnershipTransfer`'s `actorId !== ownerId` check is a race guard under the row lock, not a second authorization path. **OK.**
- (b) `ORGS_ENABLED` is a code constant (`orgs-enabled.ts:10`) and the dark check runs before authentication. The dark 404 is tested for all 14 routes. No `process.env` in the diff. **OK.**
- (c) X-2 precondition: mutated, went red (below). **OK.**
- (d) Creation writes org + OWNER row in one transaction. Transfer locks the org row, demotes and then promotes in one transaction (`membership.ts:148-156`). Tests cover "never diverge" and concurrent transfers. **OK.**
- (e) Only `tokenHash` is stored, and lookup is by `hashToken(token)` with no plaintext compare. An expired open invite is rotated in place. Duplicate detection uses `lower(email)`, and accept compares `normalizeEmail` of both addresses. New and existing accounts are both tested. Accepted and revoked or resent tokens are refused. A token for org A cannot land in org B, because accept takes no orgId and the org comes from the invite row itself. **OK.**
- (f) Live-only seats; SEAT-3 marked partial. **OK.**
- (g) See the ORG-6 coordination note below.
- (h) Every drive, trashed ones included, is resolved in one transaction before the org row is deleted, and there is one audit event per drive. **OK.**
- (i) All 14 routes × non-member, Member, Admin and Owner, plus 401. **Covered by the test design, but red in CI (P1-1).**
- (j) No self-authenticating route: accept requires a session, so the middleware allowlist is not needed. **OK.**
- (k) `ORG_WRITE_AUTH` sets `requireCSRF: true` on every mutation (by reading), but **no test holds it**: P2 inline. No `any`.

## ORG-6 coordination with B6 (#2663), not a finding against this diff
`DELETE /api/account` refuses an Owner before anything is queued (`account/route.ts:222`), which is correct. B6's `LeaveOrganizationRefusedError` fires only inside `accountRepository.deleteUser`, and the processor calls that at `account-erasure-worker.ts:192`, after the destructive steps at 109-182 (avatars, activity anonymization, AI usage logs, token revocation, provider erasure). Two paths bypass this route and reach the worker while the subject owns an org:
1. An admin-lodged erasure (`apps/web/src/app/api/admin/gdpr/erasure/route.ts`), and `apps/admin` `users/[userId]/data` DELETE.
2. The race window: a user lodges an erasure, then creates an org before the job runs, and `POST /api/orgs` does not check for a pending erasure.

The result is a half-erased Owner account that retries forever. This does not block ORG-6 for this PR, because the self-service door is the Owner's path. The orchestrator should decide where the up-front check lives (a worker precheck that returns `blocked`, like `planDriveDisposition`), so it does not end up as two independent guards.

## Test honesty (run by me, not copied from the PR)
**Caveat that matters:** like the lane, I ran the web tests under a source-alias config (`@pagespace/lib/*` → `src`). That harness is exactly what hides P1-1: in CI, web resolves lib from `dist`, where `authorize.js` imports `./repository` relatively and the test's module mock never applies. My local 107/107 for web does **not** carry over to CI. The lib and scripts numbers are unaffected.

Setup: throwaway homebrew postgresql@17 on :5461 (short `-k` socket dir, `timezone=UTC`), `bun run db:migrate` on this head, source-alias vitest configs, single files. The server has since been stopped and removed.
- `packages/lib`: `org-decisions.test.ts` + `org-services.integration.test.ts` → **43 passed (43)**, exit 0.
- `apps/web`: `org-routes.test.ts` + `account/__tests__/delete-route.test.ts` → **107 passed (107)**, exit 0.
- `scripts`: `orgs-enabled-export-precondition.test.ts` → **3 passed (3)**, exit 0.

Mutations, by line index with the anchor asserted, each restored with `git diff` confirmed empty. These are different lines from the lane's six:
| Line broken | Result |
|---|---|
| `invitations.ts:73` email-mismatch check → `if (false) {` | **red**: "ORG-3 (partial) an existing account accepts as the invited role; another account cannot use the link" and the pure "…refuses an account whose email is not the invited address" (2 failed) |
| `account/route.ts:223` Owner guard → `if (false) {` | **red**: "ORG-6 an Owner cannot delete their account while owning an org" |
| `orgs-enabled.ts:10` → `true` | **red**: "X-2 (partial) ORGS_ENABLED cannot be true while org tables are excluded from GDPR export" |
| `deletion.ts:136` trash-by-choice spread → `...({})` | **red**: "ORG-6 deleting an org never leaves a drive without an owner" |
| `org-route-auth.ts:15` `requireCSRF: true` → `false` | **SURVIVED**: 97/97 green (P2) |

## Rules and scope
- Pure decisions (`decideOrgRole`, `decideRoleChange`, `decideOwnershipTransfer`, `decideInviteCreation`/`Acceptance`, `planOrgDeletion`) are separate from IO. No `any`. No env flag. SQL `now()` is UTC. The six new `@pagespace/lib/organizations/*` subpaths all have `exports` entries. No gate self-test title names a real Spec ID.
- No migration and no schema change. Nothing outside the leaf, except the account-delete guard and invite email, both of which the leaf and ORG-6 call for. Permissions: org roles are a new domain with one function, and no drive or page permission logic is rolled by hand.
- Title format: every `(partial)` marker is one ASCII space, lowercase.
- Note: the lane worktree holds an **unpushed** commit `2f7d445d4 fix(orgs): org avatar URL must be https`, which touches the P2 below. This review covers `182dd9efe` only. Once that commit is pushed, the patch changes and needs a fresh review (0e carry-over does not apply).

## CI on this head
Run 35185226052 (head `182dd9efe`): Lint & TypeScript **success**, E2E **success**, Unit Tests **failure** (`Tests 54 failed | 20725 passed | 6 skipped`, all 54 in `org-routes.test.ts`; see P1-1), Spec ID coverage **success**. Coverage lines for this PR's IDs, quoted from job 105091482065:
```
ORG-1     allowlisted  partial: apps/web/src/app/api/orgs/__tests__/org-routes.test.ts, packages/db/src/schema/__tests__/organizations.test.ts, packages/lib/src/organizations/__tests__/org-decisions.test.ts
ORG-2     allowlisted  partial: packages/db/src/schema/__tests__/organizations.test.ts, packages/lib/src/organizations/__tests__/org-decisions.test.ts
ORG-3     allowlisted  partial: apps/web/src/app/api/orgs/__tests__/org-routes.test.ts, packages/db/src/schema/__tests__/organizations.test.ts, packages/lib/src/organizations/__tests__/org-decisions.test.ts, packages/lib/src/services/__tests__/notification-email-service.test.ts
ORG-5     allowlisted  partial: apps/web/src/app/api/orgs/__tests__/org-routes.test.ts, packages/lib/src/organizations/__tests__/org-decisions.test.ts
ORG-6     covered      apps/web/src/app/api/account/__tests__/delete-route.test.ts, partial: packages/db/src/schema/__tests__/organizations.test.ts, packages/lib/src/organizations/__tests__/org-decisions.test.ts
SEAT-3    allowlisted  partial: packages/lib/src/organizations/__tests__/org-decisions.test.ts
X-1       allowlisted  -
X-2       allowlisted  -
```
Notes:
- `org-services.integration.test.ts` appears under no ID (P1-2).
- X-2 shows `-`: the `scripts/__tests__` precondition test is not credited in this run. It passed locally for me (3/3). Confirm on the next run that the scripts step's JSON reaches the coverage job.
- Coverage job green is not merge evidence while Unit Tests is red.

apps/web/src/lib/orgs/org-route-auth.ts
```diff
@@ -0,0 +1,55 @@
...
+import { NextResponse } from 'next/server';
+import { authenticateRequestWithOptions, isAuthError } from '@/lib/auth';
+import { auditRequest } from '@pagespace/lib/audit/audit-log';
+import { ORGS_ENABLED } from '@pagespace/lib/organizations/orgs-enabled';
+import { requireOrgRole } from '@pagespace/lib/organizations/authorize';
+import type { OrgRole } from '@pagespace/db/schema/organizations';
+
+// Session only: CLI and MCP parity is Wave G (Spec X-1).
+export const ORG_READ_AUTH = { allow: ['session'] as const, requireCSRF: false };
+export const ORG_WRITE_AUTH = { allow: ['session'] as const, requireCSRF: true };
```
2witstudios: **P2: CSRF on org mutations is not held by any test.** I mutated this line to `requireCSRF: false` and `org-routes.test.ts` stayed 97/97 green, because `authenticateRequestWithOptions` is mocked and nothing inspects the options it gets. For the same reason, a mutation route that passed `ORG_READ_AUTH` instead of `ORG_WRITE_AUTH` (PATCH, DELETE or POST under `/api/orgs`) would also pass. Fix: in the role matrix, give each `RouteCase` an `isMutation` flag and assert `expect(authenticateRequestWithOptions).toHaveBeenCalledWith(expect.any(Request), expect.objectContaining({ requireCSRF: true }))` for every mutation, and `allow: ['session']` for all routes. Then mutation-check it (this line → `false` goes red).

apps/web/src/lib/orgs/org-schemas.ts
```diff
@@ -0,0 +1,47 @@
...
+import { z } from 'zod/v4';
+
+const orgName = z.string().trim().min(1).max(100);
+// Lowercase letters, digits and inner hyphens; it becomes part of org URLs.
+const orgSlug = z
+  .string()
+  .trim()
+  .toLowerCase()
+  .regex(/^[a-z0-9](?:[a-z0-9-]{0,46}[a-z0-9])?$/, 'Use 1-48 lowercase letters, digits or hyphens');
+const avatarUrl = z.url().max(2048).nullable();
```
2witstudios: **P2: `avatarUrl` accepts any URL scheme.** On this head, zod v4 `z.url().max(2048).nullable()` returns success for `javascript:alert(1)` and `data:image/svg+xml,<svg onload=alert(1)>` (checked with `safeParse`). An org Admin can store that value, and every member's client will render it once the Wave F org UI lands. Today the value is returned by `GET /api/orgs` and `GET /api/orgs/[orgId]`. The fix: allow https only (for example `z.url({ protocol: /^https$/ })`), with a route test that refuses `javascript:`, `data:` and `http:`. The lane worktree already has an unpushed `2f7d445d4 fix(orgs): org avatar URL must be https`. Push it and this thread can be verified against that SHA.

apps/web/src/app/api/orgs/[orgId]/route.ts
```diff
@@ -0,0 +1,115 @@
...
+export async function DELETE(request: Request, context: Context) {
+  const { orgId } = await context.params;
+  const gate = await authorizeOrgRequest(request, orgId, 'OWNER', ORG_WRITE_AUTH);
+  if (!gate.ok) return gate.response;
+  try {
+    const parsed = orgDeleteSchema.safeParse(await request.json().catch(() => null));
+    if (!parsed.success) {
+      return NextResponse.json({ error: 'Invalid request body', issues: parsed.error.issues }, { status: 400 });
+    }
+    const result = await deleteOrganization({ orgId, choices: parsed.data.drives, now: new Date() });
```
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/orgs/[orgId]/route.ts]: Fixed in 7ecbf8726 (pushed in 61157530b). `deleteOrganization` now takes `actorId`, and `planOrgDeletion` refuses `not_owner` unless the caller equals the `ownerId` read under the org row `FOR UPDATE` lock. The route passes `gate.userId` and maps the refusal to 403. The same stale-authorization gap existed for `changeMemberRole`/`removeMember`, which trusted the route's pre-transaction role. They now lock the actor's and target's membership rows in id order and re-check that the actor is still an Admin. Tests: `org-services.integration.test.ts` "ORG-6 (partial) a caller who lost ownership after the route authorized cannot delete the org" and "ORG-5 (partial) an Admin demoted after the route authorized cannot change a role or remove a member"; `org-decisions.test.ts` "ORG-6 (partial) a caller who is not the Owner at the time of the write is refused"; `org-routes.test.ts` "ORG-6 (partial) DELETE /api/orgs/[orgId] passes the caller so the service can re-check ownership, and maps its refusal". Mutations: `deletion.ts:66` → `if (false)` red, and `membership.ts:25` → `return null` red. Leaving this open for verification.

apps/web/src/app/api/orgs/[orgId]/invitations/[invitationId]/resend/route.ts
```diff
@@ -0,0 +1,55 @@
...
+      DISTRIBUTED_RATE_LIMITS.DRIVE_INVITE_RESEND,
+    );
+    if (!limit.allowed) {
+      return NextResponse.json(
+        { error: 'Too many resends for this invitation. Please try again later.' },
+        { status: 429, headers: { 'Retry-After': String(limit.retryAfter ?? 900) } },
+      );
+    }
+
+    const result = await resendInvitation({ orgId, invitationId, now: new Date() });
```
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/orgs/[orgId]/invitations/[invitationId]/resend/route.ts]: Fixed in 7ecbf8726 (pushed in 61157530b). Delivery now happens inside the service: `resendInvitation({ …, deliver })` rotates the token under a row lock, calls `deliver`, and if `deliver` throws it restores the previous `tokenHash` and `expiresAt`. The restore is conditioned on the row still carrying the new hash and being unaccepted, so it never undoes a later write. Invite creation follows the same rule: a failed delivery deletes a new row, or restores a rotated row exactly as it was, so an undelivered invite never holds a seat. Tests: "ORG-3 (partial) a resend whose email fails keeps the previously delivered link working" (the old token still accepts afterwards) and "ORG-3 (partial) an invite whose email fails holds no seat; a rotated one is left as it was"; route: "ORG-3 (partial) an invite or resend that could not be emailed answers 502 and the service undoes it". Mutations: restore → new hash red, and delete compensation removed red. Leaving this open for verification.

packages/lib/src/organizations/invitations.ts
```diff
@@ -0,0 +1,257 @@
...
+  | { ok: false; status: 409; reason: 'already_member' | 'already_invited' };
+
+export async function createOrRotateInvitation(input: {
+  orgId: string;
+  email: string;
+  role: InvitableRole;
+  invitedBy: string;
+  now: Date;
+}): Promise<IssueInvitationResult> {
+  const isExistingMember = await isEmailAMember(input.orgId, input.email);
```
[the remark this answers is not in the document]
2witstudios: Fixed in 7ecbf8726 (pushed in 61157530b). Creation and acceptance now share `lockOrgInviteAddress(tx, orgId, email)`, a `pg_advisory_xact_lock` keyed on org + normalized address. Creation takes it before its member check, its `FOR UPDATE` open-invite lookup and its insert or rotate, all in one transaction. Acceptance peeks the invite's org and address, takes the same lock, and only then row-locks the invite. The lock is always taken before any invitation row lock, so the order is fixed. Deterministic tests, which hold the lock in a separate transaction and assert the other side blocks until it is released: "ORG-3 (partial) creating an invite waits while the same address is locked by an acceptance" and "ORG-3 (partial) accepting an invite waits while the same address is locked by an invite creation". Mutations: removing either lock call turns its test red. I dropped a probabilistic race loop I had first written, because it passed without the fix. Leaving this open for verification.

apps/web/src/lib/orgs/org-route-auth.ts
2witstudios: Fixed in 59f8407a8 (pushed in 61157530b). `authorizeOrgRequest` now calls `requireOrgRole(gate.userId, orgId, minRole, { findMembershipRole })`, with `findMembershipRole` imported from `@pagespace/lib/organizations/repository`. The route test's package-specifier mock therefore applies whether lib resolves from src or dist, and the matrix still runs the real `requireOrgRole`/`decideOrgRole`. Local: `bunx vitest run --config vitest.wt.config.ts src/app/api/orgs/__tests__/org-routes.test.ts` → Tests 113 passed (113). Honest caveat: that run still uses a source alias. I have not built lib `dist` locally, because the build-slot rule applies while other lanes run, so run 2 of CI on 61157530b is the dist-resolution evidence. I will quote its Unit Tests result here. Leaving this open for verification.

apps/web/src/lib/orgs/org-route-auth.ts
2witstudios: Fixed in c5112860d (pushed in 61157530b). Each `RouteCase` now carries `csrf`: false for GETs, true for every POST/PATCH/DELETE, including create and accept. A per-route test asserts `authenticateRequestWithOptions` was called with exactly `{ allow: ['session'], requireCSRF: route.csrf }`, which also catches a mutation route passing `ORG_READ_AUTH`. Test titles: "<route> authenticates a session with CSRF" / "… without CSRF". Mutation: `org-route-auth.ts:15` `requireCSRF: true` → `false` gives `Tests 10 failed | 103 passed (113)`; the no-op control stays at 113 passed. Leaving this open for verification.

apps/web/src/lib/orgs/org-schemas.ts
2witstudios: Fixed in 2f7d445d4 (pushed in 61157530b). `avatarUrl` is now `z.url({ protocol: /^https$/, hostname: z.regexes.domain }).max(2048).nullable()`. Test "ORG-1 (partial) an org avatar must be an https URL, never a script or data URI": POST /api/orgs refuses `javascript:alert(1)`, `data:image/png;base64,AAAA` and `http://example.test/n.png` with 400, PATCH refuses `javascript:`, and neither service is reached. The test failed before the schema change and passes after. Leaving this open for verification.

2witstudios:
Re independent review 5231803032, **P1-2** (the org services integration suite never ran in CI, so ORG-6 plain is premature): agreed on the mechanism. On 182dd9efe, web `test:coverage` failed, which skipped both "Run packages/lib integration suites" and the `scripts` step. That also explains why X-2 showed `-` in the coverage job. P1-1 is fixed in 59f8407a8, and head 61157530b now runs both steps. I am keeping ORG-6 off the allowlist **conditionally**: when this head's CI finishes, I will check that the Spec ID coverage job lists `packages/lib/src/organizations/__tests__/org-services.integration.test.ts` under ORG-6 with the three "ORG-6 deleting an org…" tests passing. If it does not, I will restore ORG-6 to the allowlist in the next push and say so here. The B6 note (admin-lodged erasure and the lodge-then-create-org race reaching the worker) is outside this leaf, and I have passed it to the orchestrator in #org-wallets.

2witstudios:
CI evidence for P1-1 and P1-2 from run 2 on 61157530b:
- **P1-1 fixed under dist.** Web `test:coverage` (lib resolved from the built dist): `Test Files 1346 passed | 1 skipped (1347)`, `Tests 20795 passed | 6 skipped (20801)`, 0 failed. The `org-routes.test.ts` role matrix passes.
- **P1-2 met.** The lib integration step ran `org-services.integration.test.ts (22 tests)`, all passing. The Spec ID coverage job credits it: `ORG-6 covered apps/web/src/app/api/account/__tests__/delete-route.test.ts, packages/lib/src/organizations/__tests__/org-services.integration.test.ts, partial: …`. ORG-6 stays off the allowlist. X-2 is now credited as partial via `scripts/__tests__/orgs-enabled-export-precondition.test.ts`.
- **Unit Tests was still red on 61157530b**, for a different reason. In the same step, the last lib integration file, `replay-across-replicas.integration.test.ts`, failed in `beforeAll` with `could not reach Postgres`, and the Postgres log shows `FATAL: sorry, too many clients already`. That step is otherwise `Test Files 1 failed | 37 passed | 2 skipped`, `Tests 599 passed`. The integration run keeps one process with a pool per file, and my suite held several connections without releasing them. 19db6730c ends its pool in `afterAll`, the same pattern as the drive-envs and agent-sessions suites; locally it still passes 22/22. I will quote run 3's result here.

2witstudios:
Run 3 on 19db6730c: **all checks green**, including `ci / Unit Tests` (lib integration too, after ending the pool) and `ci / Spec ID coverage`. I then pushed 02a4e235e, which contains:
- 6d638d0c4: org-services integration cleanup now runs in dependency order without swallowing errors, and sweeps orgs owned by the file's users. This follows the point-guard cross-lane rule that organizations.ownerId and drives.orgId are RESTRICT and must not leak into later suites. It is test-only. Real pg17: 22/22, and 0 organizations / org_members / org_invitations / org drives / test users remain after repeated runs.
- A merge of integration tip 3da2e8021 (#2666, money-model; unrelated files, no conflicts).

The patch changed after 19db6730c, so run 4 and a re-review are needed.

2witstudios (commented):
[independent-review] agent:ow-irv-2665 (re-review)

Re-review of #2665 from `182dd9efe` (my review 5231803032) to the final head `02a4e235e`. Commits reviewed: 2f7d445d4, 7ecbf8726, c5112860d, 59f8407a8, the tip merge 61157530b (brings #2661 B2 `permissions/org-access`, base content, not this lane's code), 19db6730c (pool teardown), 6d638d0c4 (integration cleanup hardening) and the tip merge 02a4e235e (brings #2666 money-model, base content). Merge-base = integration tip `3da2e8021`. I checked that the lane's delta after 19db6730c is test-only: `git diff 19db6730c 02a4e235e` restricted to this PR's paths (`api/orgs`, `lib/orgs`, `lib/src/organizations`, `api/account`, the allowlist, the X-2 precondition test, `lib/package.json`) touches only `org-services.integration.test.ts` (+16/−6). The only other files in that range are #2666's `money-model.ts`, `money-model.test.ts` and `create-credit-topup` route test, all from the base.

**Verdict: all 4 of my findings FIXED; Codex's 3 threads FIXED; no new P1/P2.** Per the gate rules, I leave the threads open for the orchestrator to resolve.

## My findings
| # | Finding | Status | Evidence |
|---|---|---|---|
| P1-1 | `org-routes.test.ts` red in CI: the membership mock never applied under lib `dist` | **FIXED** | `org-route-auth.ts:46` now passes `{ findMembershipRole }` imported by package specifier, so the mock applies under dist. See the CI section below for the run on the built lib. |
| P1-2 | lib integration suite never ran in CI, so ORG-6's org-delete clause was uncredited | **FIXED** | See the CI section: `org-services.integration.test.ts` ran in the lib integration step and the coverage job credits it. 19db6730c ends the suite's pool in `afterAll`. On 61157530b a later file (`replay-across-replicas.integration.test.ts`) failed with `Failed query: SELECT 1`. The lane traced this to 'too many clients' from this suite's open connections; I confirmed the failure in the log, not the cause. `pool.end()` in `afterAll` is safe here: the config keeps the default `isolate`, so each file gets its own module graph and pool, and three existing suites (`agent-sessions-store`, `drive-envs-store`, `store-adapter-infisical`) already do this. |
| P2 | CSRF on mutations untested | **FIXED** | Each route in the matrix now carries a `csrf` flag, and "<route> authenticates a session with/without CSRF" asserts the exact options `{ allow: ['session'], requireCSRF }`, which also catches a route that passes `ORG_READ_AUTH`. My mutation `org-route-auth.ts:16` `requireCSRF: true` → `false` makes **10 tests fail**, one per mutation route (last round it made none fail). |
| P2 | `avatarUrl` accepted `javascript:`/`data:` | **FIXED** | `z.url({ protocol: /^https$/, hostname: z.regexes.domain })`, plus "ORG-1 (partial) an org avatar must be an https URL, never a script or data URI" (POST refuses `javascript:`, `data:` and `http:`; PATCH refuses `javascript:`; neither service is reached). |

## Cleanup hardening (6d638d0c4): test-only, reviewed
`afterEach` no longer swallows errors (`.catch(() => {})` removed). It deletes in FK order: tracked drives, then `drives.orgId` nulled for the orgs, then orgs (org_members and org_invitations cascade), then users. It also sweeps every org **owned by** a user the test created, so an org created by a service call the test did not track (for example after a transfer) cannot survive and break later suites through `organizations.ownerId` ON DELETE RESTRICT. Drives that left an org during a delete test keep a created user as owner and go with that user through `drives.ownerId` cascade. This is sound. The only risk it adds is a loud cleanup failure, which is the right trade in a shared CI database.

## Codex threads
| Thread | Status | Evidence |
|---|---|---|
| P1 revalidate the Owner inside the delete transaction | **FIXED** | `planOrgDeletion` refuses `not_owner` unless `actorId` equals the `ownerId` read under the org row `FOR UPDATE`, and the route maps that to 403. The same gap closed for role change and removal: actor and target rows are locked in id order and the actor is re-checked through `decideOrgRole` (the one decision function, so ORG-5 still holds). My mutations: `deletion.ts:66` → **red** ("ORG-6 (partial) a caller who lost ownership after the route authorized cannot delete the org" + the pure test); `membership.ts:49` → **red** ("ORG-5 (partial) an Admin demoted after the route authorized cannot change a role or remove a member" + the pure test). |
| P2 keep a usable token when resend delivery fails | **FIXED** | Delivery moved into the service. On a throw, the previous `tokenHash`/`expiresAt` is restored, guarded by `tokenHash = new AND acceptedAt IS NULL` so it cannot undo a later write. Creation deletes a new row or restores a rotated one. My mutation `invitations.ts:245` (restore the new hash instead of the previous one) → **red** "ORG-3 (partial) a resend whose email fails keeps the previously delivered link working". |
| P2 serialize the membership check with invite creation | **FIXED** | `lockOrgInviteAddress` (`pg_advisory_xact_lock(hashtextextended('org_invite:<org>:<normalized email>', 0))`) is taken by creation before its member check, `FOR UPDATE` lookup and write, and by acceptance before its invite row lock. The order is always the same: advisory lock, then row lock. Resend and revoke take only the row lock and never the advisory lock, so no lock cycle is possible. My mutation `invitations.ts:295` (accept skips the lock) → **red** "ORG-3 (partial) accepting an invite waits while the same address is locked by an invite creation". |

## ORG-6 as a plain claim: clause by clause
Spec: "Deleting an org transfers its drives to a person or trashes them; nothing is orphaned. An Owner cannot delete their account while owning an org." Leaf point-guard decision: every org drive, trashed ones included, is resolved in one transaction; a transfer goes to a named person; a trash choice or an already-trashed drive becomes a trashed drive owned by the Owner; each drive gets an audit event naming its destination; the confirmation lists every drive and its destination; a transfer to a non-member is refused.

| Clause | Proven by (all run in CI on this head, see below) | Mutation-checked |
|---|---|---|
| transfers its drives to a person | integration "ORG-6 deleting an org never leaves a drive without an owner" (Product → Priya, `orgId` null, not trashed) | — |
| or trashes them | same test (Finance trash choice → Owner, `isTrashed`) | `deletion.ts:136` → red (round 1) |
| nothing is orphaned | same test (every drive: `orgId` null, `ownerId` set; org row gone; org-sourced `drive_members` removed) | — |
| already-trashed drives land in the Owner's trash | integration "ORG-6 deleting an org with a trashed drive puts it in the Owner's trash and deletes the org" (`ownerId` = Owner, still trashed, original `trashedAt` kept, org row deleted) and the Customer Research drive in the test above | — |
| resolved in ONE transaction; a bad choice changes nothing | integration "ORG-6 deleting an org refuses a transfer to someone outside the org and changes nothing" | — |
| destinations listed | `planOrgDeletion` returns a step per drive; route test "ORG-6 (partial) DELETE /api/orgs/[orgId] writes one audit event per drive naming its destination" asserts the response `drives` equals the steps | — |
| audit per drive naming its destination | same route test (one `resourceType: 'drive'` event per step, with `destination` and `newOwnerId`) | — |
| only the Owner, re-checked at write time | integration "ORG-6 (partial) a caller who lost ownership…" | `deletion.ts:66` → red |
| An Owner cannot delete their account while owning an org | `delete-route.test.ts` "ORG-6 an Owner cannot delete their account while owning an org" (400, owned orgs listed, erasure never queued); the real query in integration "ORG-6 (partial) findOrganizationsOwnedBy lists the orgs that block the Owner account delete" | `account/route.ts:223` → red (round 1) |

**Verdict: ORG-6 plain is justified; I agree with removing it from the allowlist.** Every Spec clause and every element of the point-guard decision has a named test that ran in CI and was mutation-checked where it matters. Two things stay outside this claim, and neither is a Spec clause:
- The confirmation as a UI is Wave F. The API response lists every destination, which is what this leaf owns.
- The async erasure path still depends on B6: an admin-lodged erasure, or an org created after an erasure was requested, reaches the processor worker, which runs destructive steps before `LeaveOrganizationRefusedError` fires in `deleteUser`. This is the coordination note from round 1. It concerns an admin or a race, not the Owner deleting their own account, so it is not an ORG-6 gap in this PR. It remains open for the orchestrator to place (B6 or Phase 6).

## Tests I ran on 61157530b (source alias; later lane commits change only test teardown and cleanup)
- `packages/lib` `org-decisions.test.ts` + `org-services.integration.test.ts` against my own throwaway homebrew postgresql@17 (:5461, `timezone=UTC`, migrated on this head): **51 passed (51)**, exit 0. Stopped and removed afterwards.
- Mutations by line index, each with an asserted anchor, restored with `git diff` confirmed empty: `deletion.ts:66`, `invitations.ts:295`, `invitations.ts:245`, `membership.ts:49`, `org-route-auth.ts:16`. **All 5 red**, as cited above.

## CI on the final head
**Run 35197825388 on `02a4e235e`: every job green.** E2E success, Unit Tests success, Lint & TypeScript Check success, Spec ID coverage success. Security Tests (run 35197825008) was still in progress when I last checked it, so it is not included here. From the Unit Tests job (105125142964), where web resolves `@pagespace/lib` from **dist**:
```
 ✓ src/app/api/orgs/__tests__/org-routes.test.ts (113 tests)
 ✓ src/app/api/account/__tests__/delete-route.test.ts (10 tests)
      Tests  20799 passed | 6 skipped (20805)
@pagespace/lib test:integration:  ✓ src/organizations/__tests__/org-services.integration.test.ts (22 tests)
@pagespace/lib test:integration:       Tests  602 passed | 39 skipped (641)
```
From the Spec ID coverage job (105134353052):
```
ORG-1     allowlisted  partial: …org-routes.test.ts, …db organizations.integration.test.ts, …organizations.test.ts, …org-decisions.test.ts, …org-services.integration.test.ts
ORG-2     allowlisted  partial: …org-routes.test.ts, …db organizations.integration.test.ts, …organizations.test.ts, …org-decisions.test.ts, …org-services.integration.test.ts
ORG-3     allowlisted  partial: …org-routes.test.ts, …db organizations.integration.test.ts, …organizations.test.ts, …org-decisions.test.ts, …org-services.integration.test.ts, …notification-email-service.test.ts
ORG-5     allowlisted  partial: apps/web/src/app/api/orgs/__tests__/org-routes.test.ts, packages/lib/src/organizations/__tests__/org-decisions.test.ts, packages/lib/src/organizations/__tests__/org-services.integration.test.ts
ORG-6     covered      apps/web/src/app/api/account/__tests__/delete-route.test.ts, packages/lib/src/organizations/__tests__/org-services.integration.test.ts, partial: (route, db, decisions, services tests)
SEAT-3    allowlisted  partial: packages/lib/src/organizations/__tests__/org-decisions.test.ts, packages/lib/src/organizations/__tests__/org-services.integration.test.ts
X-1       allowlisted  -
X-2       allowlisted  partial: scripts/__tests__/orgs-enabled-export-precondition.test.ts
```
- **P1-1 confirmed:** `org-routes.test.ts` passes against lib dist, 113/113.
- **P1-2 confirmed:** the org services integration suite ran in CI (22/22) and is credited under ORG-1/2/3/5/6 and SEAT-3.
- ORG-6 `covered` rests on the account-clause test and the plain org-delete integration tests together, which matches the clause table above.
- X-2 is now credited (it showed `-` in round 1).
- X-1 is correctly still uncovered (Wave G).
- The previous head `19db6730c` (run 35194790357) was also fully green, with the same lines.

apps/web/src/app/api/orgs/[orgId]/route.ts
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/orgs/[orgId]/route.ts]: [point-guard] Verified FIXED at 02a4e235e by independent re-review 5233253404 (P1-1 route tests pass against lib dist, 113/113 in CI; P1-2 and both P2s FIXED; Codex's three threads FIXED including owner recheck and invite locks; 5 new mutations red) and CI run 35197825388 all green. ORG-6 plain claim justified clause by clause by the reviewer. Resolving.

apps/web/src/app/api/orgs/[orgId]/invitations/[invitationId]/resend/route.ts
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/orgs/[orgId]/invitations/[invitationId]/resend/route.ts]: [point-guard] Verified FIXED at 02a4e235e by independent re-review 5233253404 (P1-1 route tests pass against lib dist, 113/113 in CI; P1-2 and both P2s FIXED; Codex's three threads FIXED including owner recheck and invite locks; 5 new mutations red) and CI run 35197825388 all green. ORG-6 plain claim justified clause by clause by the reviewer. Resolving.

packages/lib/src/organizations/invitations.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED at 02a4e235e by independent re-review 5233253404 (P1-1 route tests pass against lib dist, 113/113 in CI; P1-2 and both P2s FIXED; Codex's three threads FIXED including owner recheck and invite locks; 5 new mutations red) and CI run 35197825388 all green. ORG-6 plain claim justified clause by clause by the reviewer. Resolving.

apps/web/src/lib/orgs/org-route-auth.ts
2witstudios: [point-guard] Verified FIXED at 02a4e235e by independent re-review 5233253404 (P1-1 route tests pass against lib dist, 113/113 in CI; P1-2 and both P2s FIXED; Codex's three threads FIXED including owner recheck and invite locks; 5 new mutations red) and CI run 35197825388 all green. ORG-6 plain claim justified clause by clause by the reviewer. Resolving.

apps/web/src/lib/orgs/org-route-auth.ts
2witstudios: [point-guard] Verified FIXED at 02a4e235e by independent re-review 5233253404 (P1-1 route tests pass against lib dist, 113/113 in CI; P1-2 and both P2s FIXED; Codex's three threads FIXED including owner recheck and invite locks; 5 new mutations red) and CI run 35197825388 all green. ORG-6 plain claim justified clause by clause by the reviewer. Resolving.

apps/web/src/lib/orgs/org-schemas.ts
2witstudios: [point-guard] Verified FIXED at 02a4e235e by independent re-review 5233253404 (P1-1 route tests pass against lib dist, 113/113 in CI; P1-2 and both P2s FIXED; Codex's three threads FIXED including owner recheck and invite locks; 5 new mutations red) and CI run 35197825388 all green. ORG-6 plain claim justified clause by clause by the reviewer. Resolving.

2witstudios:
B6 reconciliation, as the point guard directed: f6a5a91e5 merges `origin/pu/org-wallets` at c370c75bd (#2663, B6), no rebase.

The delta:
- `apps/web/src/app/api/account/route.ts`: B6's guard stays (`accountRepository.getOwnedOrganizationNames`). This lane's duplicate guard and its `@pagespace/lib/organizations/repository` import are removed. B6's error text is now this lane's wording: `'Transfer ownership of your organizations or delete them before deleting your account'`.
- `apps/web/src/app/api/account/__tests__/delete-route.test.ts`: B6's beforeEach mock stays. This lane's "ORG-6 an Owner cannot delete their account while owning an org" test, its mock and its import are removed. B6's "ORG-6 (partial) given the user owns an organization, should block with 400 and NOT queue" now also asserts the new error text.
- `packages/lib/src/organizations/repository.ts`: `findOrganizationsOwnedBy` is removed, because only the dropped guard used it (knip). Its integration test "ORG-6 (partial) findOrganizationsOwnedBy lists…" is removed too.
- `packages/lib/package.json`: all six `./organizations/*` entries from this lane and B6's `./organizations/leave` are kept (auto-merged).
- ORG-6 stays covered by the plain "ORG-6 deleting an org…" integration tests. The account clause is now proven by B6's `(partial)`-titled tests.

Runs (source-alias harness, own pg17):
```
apps/web$ bunx vitest run --config vitest.wt.config.ts src/app/api/account/__tests__/delete-route.test.ts src/app/api/orgs/__tests__/org-routes.test.ts
 ✓ org-routes.test.ts (113 tests) · ✓ delete-route.test.ts (10 tests)
      Tests  123 passed (123)
packages/lib$ bunx vitest run --config vitest.wt.config.ts src/organizations/__tests__/org-decisions.test.ts src/organizations/__tests__/org-services.integration.test.ts
      Tests  50 passed (50)
```
tsc on the touched lib and web files: 0 errors. eslint: 0. Mutation: dropping "or delete them " from the guard's error text gives `Tests 1 failed | 9 passed (10)` in delete-route; restored.

apps/web/src/app/api/account/__tests__/delete-route.test.ts
```diff
@@ -153,6 +153,7 @@ describe('DELETE /api/account (async erasure)', () => {
       const res = await DELETE(deleteReq(mockUserEmail));
       const body = await res.json();
       expect(res.status).toBe(400);
+      expect(body.error).toBe('Transfer ownership of your organizations or delete them before deleting your account');
```
2witstudios: **P2: no plain-titled test proves ORG-6's account clause anymore.** This test (B6's, which replaced B3's plain `ORG-6 an Owner cannot delete their account while owning an org` in the merge) is titled `ORG-6 (partial)`, and so is every other sentence-2 test on the head (admin GDPR, admin data route, erasure worker, `leave.integration` deleteUser). ORG-6 still reads `covered`, but only through the three plain org-delete integration tests, which prove sentence 1 alone. The clause is proven; the title is what lags. Suggested fix: retitle this test to the plain `ORG-6 an Owner cannot delete their account while owning an org`. It asserts the Owner's own delete is refused with 400 before anything is queued, which is what round 1 accepted as the plain proof. Test title only; no code change.

2witstudios (commented):
[independent-review] agent:ow-irv-2665 (merge review)

Scope: **only** `02a4e235e..f6a5a91e5`, the merge of integration tip `c370c75bd` (B6 #2663) into ow-b3. Merge-base of the head with `origin/pu/org-wallets` = `c370c75bd`, the current tip.

**Verdict: 0 P1, 1 P2 (inline). Everything else asked for checks out.**

## Checks
| Check | Result | Evidence |
|---|---|---|
| Exactly one owner refusal in `apps/web/src/app/api/account/route.ts` | **CONFIRMED** | One block at `route.ts:219-228`: `accountRepository.getOwnedOrganizationNames(userId)` (B6), then 400 `'Transfer ownership of your organizations or delete them before deleting your account'`. It runs after the in-flight-erasure check and before any drive work or queueing. B3's `findOrganizationsOwnedBy` guard is gone, and `git grep findOrganizationsOwnedBy` on the head finds nothing, so knip has no dead export to flag. No conflict markers. |
| B3's error wording moved into B6's guard | **CONFIRMED** | Relative to the tip, the only change to `route.ts` is that one string, and `delete-route.test.ts` gains `expect(body.error).toBe(<that string>)`. |
| `packages/lib/package.json` keeps both export sets | **CONFIRMED** | Head has B3's `./organizations/{authorize,deletion,invitations,membership,orgs-enabled,repository}` (lines 136-165) and B6's `./organizations/leave` (line 1846). The file parses as JSON. Relative to the tip, the diff is exactly B3's six entries added. |
| Nothing else changed | **CONFIRMED** | I compared the lane's patch before and after the merge: `git diff 3da2e8021 02a4e235e` against `git diff c370c75bd f6a5a91e5`. The file set is identical (30 files) and the changed lines are identical in 26 of them. The other 4 differ only as intended: `account/route.ts` (B3's guard dropped, only the wording remains), `delete-route.test.ts` (B3's test dropped for B6's, plus the wording assertion), `repository.ts` (`findOrganizationsOwnedBy` removed, −10), and `org-services.integration.test.ts` (its `findOrganizationsOwnedBy` test removed, −6). No change to routes, services, schemas, the allowlist or the X-2 test. |

## P2 (inline): ORG-6 stays a plain `covered` claim, but no plain-titled test proves the account clause anymore
Before this merge, the plain set for ORG-6 held both clauses: the three `ORG-6 deleting an org…` integration tests (sentence 1) and B3's `ORG-6 an Owner cannot delete their account while owning an org` (sentence 2). The merge dropped that test for B6's `ORG-6 (partial) given the user owns an organization, should block with 400 and NOT queue`. Every test of sentence 2 on the head is now titled `(partial)`:
- web `delete-route.test.ts`
- `admin-gdpr-routes.test.ts`
- admin `users/[userId]/data`
- `account-erasure-worker.test.ts`
- `leave.integration.test.ts` "deleteUser refuses an org Owner…"

The gate will still print `ORG-6 covered`, but only because the plain org-delete tests pass, and they prove sentence 1 alone. Sentence 2 is actually proven, more strongly than before, since B6 closed the admin and worker paths I raised in round 1. The claim is true; its labels no longer show it. Fix: title one test that proves sentence 2 at the Owner's own door as plain, e.g. this route test → `ORG-6 an Owner cannot delete their account while owning an org`. Alternatively, mark the three org-delete tests `(partial)` and put ORG-6 back on the allowlist. Retitling is the smaller and more accurate fix. It is a test title only, so it needs no code change.

## CI on `f6a5a91e5`
- Test Suite run 35203457327: **success**. Its jobs are Unit Tests (including the lib integration step), Lint & TypeScript, E2E and Spec ID coverage.
- Security Tests run 35203456542: **success**.

I confirmed both run conclusions for this head SHA myself. I did not re-read the job logs this round: the merge changes only one error string and titles/assertions in 4 files, and I verified the patch-level equivalence above.

apps/web/src/app/api/account/__tests__/delete-route.test.ts
2witstudios: Fixed in df10736af (title only, no other change). The test is now `it('ORG-6 an Owner cannot delete their account while owning an org', …)`, a plain title, because it proves the whole account clause: a 400 with the refusal text and the owned orgs, and `lodgeAndEnqueueErasure` never called. `apps/web$ bunx vitest run --config vitest.wt.config.ts src/app/api/account/__tests__/delete-route.test.ts` → `Tests 10 passed (10)`. Leaving this open for verification.

2witstudios (commented):
[independent-review] agent:ow-irv-2665 (P2 follow-up)

Scope: **only** `f6a5a91e5..df10736af` (after my review 5233844807). Head merge-base = integration tip `c370c75bd`.

**Verdict: 0 P1, 0 P2. The P2 from 5233844807 is FIXED.**

| Check | Result | Evidence |
|---|---|---|
| The diff is exactly the one test retitle | **CONFIRMED** | `git diff f6a5a91e5 df10736af`: 1 file, 1 line changed (`apps/web/src/app/api/account/__tests__/delete-route.test.ts`). `it('ORG-6 (partial) given the user owns an organization, should block with 400 and NOT queue', …)` becomes `it('ORG-6 an Owner cannot delete their account while owning an org', …)`. The test body and assertions are unchanged: 400, the transfer-or-delete error string, the `ownedOrganizations` list, and erasure never queued. |
| delete-route.test.ts is a **plain** ORG-6 hit in spec coverage | **CONFIRMED** | Run 35207621542, job `ci / Spec ID coverage` (105169336035): `ORG-6     covered      apps/web/src/app/api/account/__tests__/delete-route.test.ts, packages/lib/src/organizations/__tests__/org-services.integration.test.ts, partial: apps/admin/…/users/[userId]/data/__tests__/route.test.ts, apps/processor/…/account-erasure-worker.test.ts, apps/web/…/admin-gdpr-routes.test.ts, apps/web/…/org-routes.test.ts, packages/db/…, packages/lib/…/leave.integration.test.ts, …`. The plain hits are now `delete-route.test.ts`, which covers sentence 2 (an Owner cannot delete their account), and `org-services.integration.test.ts`, which covers sentence 1 (org delete transfers or trashes drives and nothing is orphaned). Each ORG-6 sentence has a plain-titled test again. |
| CI on `df10736af` | **green** | Run 35207621542 `success`: E2E, Unit Tests, Lint & TypeScript Check and Spec ID coverage all `success` (E2E passed on the rerun, per the orchestrator). |

This PR has no open finding left from my reviews 5231803032, 5233253404, 5233844807 and this one.

apps/web/src/app/api/account/__tests__/delete-route.test.ts
2witstudios: [point-guard] Verified FIXED at df10736af by independent review 5234512481 (delta is one line in one file: the delete-route test is now the plain 'ORG-6 an Owner cannot delete their account while owning an org'); CI all green on this head after the known E2E flake rerun. Resolving.

PR #2668, opened 2026-09-17
feat(orgs): org-owned drives — move in, move out, create in org, lead rules (ow-b5; DRV-1..4, O-7, O-10)

2witstudios:
Lane **ow-b5**, leaf gj6frj67pnk7ek4kh6zyrzf4 (Phase 1). Org-owned drives: move in, move out, create in an org, lead rules. Base: `pu/org-wallets` at b2852bcfc (B3 #2665, B4 #2664, B6 #2663 merged in; merge-base equals the tip).

## What this adds
- **Pure core** `packages/lib/src/organizations/org-drive-ownership.ts`
  - `decideMoveDriveIntoOrg`: a Home drive is refused first (new drive-guards action `org-move`). Only the drive owner may move it in, and the owner must be a member of the target org (the lead is always an org member, O-7). A drive already in an org, or trashed, is refused.
  - `decideMoveDriveOutOfOrg`: only an org Owner or Admin. The keep-or-remove choice for org members is **required**, with no default (D-OW-10).
  - `decideCreateDriveInOrg`: org members, subject to the POL-5 creation policy passed in through a seam.
- **Service** `packages/lib/src/services/org-drive-service.ts`: `moveDriveToOrg`, `moveDriveOutOfOrg`, `createOrgDrive`.
  - Each runs in one transaction: lock the drive `FOR UPDATE` and the org `FOR SHARE`; read the actor's org role **inside that transaction**, through `requireOrgRole` with a `FOR SHARE` lock on their `org_members` row; decide; write; then run B4's sync in the same transaction. Realtime events publish only after commit.
  - The row lock is what makes a concurrent leave or account deletion (which deletes that row) and a move serialize, so a committed org drive is never cascaded away by its lead's deletion (review 5235664932 P1 #1). If the two deadlock, Postgres aborts one side, which fails closed.
  - Refusals disclose nothing to outsiders: ownership (move-in) and org authority (move-out) are checked before a drive's Home kind or org membership.
  - Move-in writes `orgId`, the slug, and the visibility. An unchosen visibility is reset to the column `DEFAULT` (Open), so a Private setting left from an earlier org never comes back (DRV-4).
  - A slug already taken inside the org gets the next suffix (D-OW-15). A create or move-in that loses the per-org unique-index race re-runs its transaction (`retryOnOrgSlugConflict` in the pure module). It retries only a 23505 on `drives_org_slug_unique` read from `DrizzleQueryError.cause`, and at most 5 attempts.
  - Storage re-attribution (O-9) is a stub, not a silent no-op: it returns `{ status: 'deferred', leafId: 't1759m6mfxrj5hyaleu1mdqs' }` and logs a warning.
- **Wiring** `org-drive-service-deps.ts`:
  - The org role comes from `requireOrgRole` (B3).
  - Membership rows go through `syncDriveOrgMembership` with `publishOrgMembershipSyncEvents` (B4); "keep" maps to `keepAsInvite`.
  - The POL-5 policy returns `'members'` until Wave E, with `TODO(lyt8275djmdcwlwm8wvk2xa5)`.
- **Routes**, dark while `ORGS_ENABLED` (B3) is false:
  - `PUT` and `DELETE /api/drives/[driveId]/org`: session plus CSRF only, a `data.write` audit event (`org_move_in` / `org_move_out`), and a drive `updated` broadcast.
  - `POST /api/drives` accepts `orgId` and `orgVisibility`. With an `orgId` it calls `createOrgDrive`, answers 404 while orgs are off, and refuses bearer-token (MCP/OAuth) requests with a 403. Creating in an org is session-only until Wave G parity (point-guard ruling).
  - A post-commit failure to look up broadcast recipients is logged, never a 500 after a committed move.
- **Org member removal** (`packages/lib/src/organizations/membership.ts`, Codex P1): `removeMember` now runs B6's `leaveOrganization` in its transaction, with the removing Admin as audit actor. Before this, an Admin-removed member kept leading, and fully accessing, the org drives they led. It also skipped the leave revocations.
- **Account-deletion flow**: per the point-guard ruling, B6 owns `account-repository` and the deletion transaction, so this PR does not call `leaveOrganization` or `reassignLedOrgDrives`.
  - `handle-drive` refuses to delete or transfer an org drive (409, because the lead passes to the org Owner).
  - `drives-status` stops listing org drives the person leads; it reads `drives` directly, not through B6's `getOwnedDrives`.

**Placement note:** the leaf names `drive-service.ts`. The move functions live in a new `org-drive-service.ts` beside it, so `drive-service.test.ts`'s ORM mock ladder is untouched. `createDrive` stays personal-only; creating in an org is `createOrgDrive`.

## ID → test table
Every ID is `(partial)` except the DRV-4 service test the point-guard asked for by title. **No allowlist entry is removed**:
- DRV-1: Home and the lead keeping the Owner role are covered; the org-delete RESTRICT is B3's.
- DRV-2: "envs and publishing survive" is proven at the row level only; publishing an app after a move is not tested.
- DRV-3: the POL-5 policy is Wave E.

Decisions D-OW-7 (lead is an org member; leave/delete reassigns) and D-OW-10 (keep-or-remove on move-out) are Spec open items, not gated requirement IDs, so those tests carry the decision name in plain words, with no marker.

| ID / decision | Tests (file → title) |
|---|---|
| DRV-1 (partial) | org-drive-ownership.test.ts → "a Home drive can never move into an org, with the drive-guards message"; "a non-owner learns nothing about the drive: someone else's Home drive answers NOT_DRIVE_OWNER, not HOME_DRIVE"; integration → "a Home drive is refused and nothing is written"; org route → "a refusal from the service returns its status, message and code…" |
| DRV-2 (partial) | integration → "moving a drive in keeps its members, roles, pages, envs and publishSubdomain"; "an org Admin moves a drive out: it becomes its lead's personal drive…"; "a failing membership sync rolls the move back"; **production wiring** → "an org Member cannot move a drive out, even when they lead it (real requireOrgRole)"; pure → owner-only / already-in-org / trashed / "a caller who is not an org admin learns nothing about whether the drive is in an org"; org route → PUT/DELETE contract, and "a committed move still answers 200 when the post-commit recipient lookup fails" |
| DRV-3 (partial) | **production wiring** → "a user outside the org cannot create a drive in it, and nothing is inserted (real requireOrgRole)"; integration → "the creation policy seam is consulted: admins-only refuses a Member"; "concurrent creates of the same name in one org all succeed with distinct slugs"; pure → retryOnOrgSlugConflict: re-runs after a slug conflict / never retries another constraint / never retries a bare 23505 or other error / stops after ORG_SLUG_ATTEMPTS; drives route → "goes through createOrgDrive…"; "a bearer-token request cannot create a drive in an org until CLI and MCP parity in Wave G" |
| DRV-4 | integration → **"DRV-4 a drive created in an org without a chosen visibility is Open"** |
| DRV-4 (partial) | integration → "a drive moved in without a chosen visibility is Open, even if it was Private in an earlier stint"; "…with a chosen visibility keeps the choice"; pure → orgDriveVisibilityForInsert; routes → an invalid visibility is a 400 |
| D-OW-7 | **production wiring** → "a drive owner outside the org cannot move their drive in (real requireOrgRole)"; **race** → "a move-in / a create overlapping the lead's account deletion never leaves a committed org drive deleted"; "deleting the lead's account does not cascade the org drive: the org Owner becomes its lead" (real B6 deleteUser); "an org Admin removing the lead from the org makes the org Owner the lead and keeps the drive"; handle-drive → "refuses to delete/transfer an org drive the person leads"; drives-status → "an org drive the person leads needs no action before account deletion and is not listed" |
| D-OW-10 | integration (real B4) → "moving out with "keep" leaves org members on the drive as invited members, guest untouched"; "moving out with "remove" revokes the org members and keeps the guest"; pure → "move-out carries the keep-or-remove choice through, and refuses when none was made"; org route → a missing choice is a 400 |

## Evidence
Source-alias vitest configs (no dist builds; build-slot rule) against a throwaway local Postgres 17 (session TZ America/Chicago) migrated to this head.

```
# lib, pure
bunx vitest run --config vitest.wt.config.ts src/organizations/__tests__/org-drive-ownership.test.ts
 Tests  27 passed (27)

# lib, real Postgres (production requireOrgRole, B4 sync, B6 deleteUser, B3 removeMember)
DATABASE_URL=postgresql://user:password@localhost:5475/pagespace_test bunx vitest run --config vitest.wt.config.ts src/services/__tests__/org-drive-service.integration.test.ts
 Tests  28 passed (28)
 # the two race tests: 5/5 consecutive runs passed

# B3 + B6 suites touched by the removeMember change
DATABASE_URL=… bunx vitest run --config <integration config> src/organizations/__tests__/org-services.integration.test.ts   Tests  21 passed (21)
DATABASE_URL=… bunx vitest run --config vitest.wt.config.ts src/organizations/__tests__/leave.integration.test.ts            Tests  9 passed (9)

# web (6 files)
bunx vitest run --config vitest.wt.config.ts "src/app/api/drives/[driveId]/org/__tests__/route.test.ts" src/app/api/drives/__tests__/route.test.ts src/app/api/drives/__tests__/manage-keys-scope.test.ts src/app/api/drives/__tests__/oauth-scope.test.ts src/app/api/account/handle-drive/__tests__/route.test.ts src/app/api/account/drives-status/__tests__/route.test.ts
 Test Files  6 passed (6)   Tests  91 passed (91)
```
- **Spec ID coverage, offline:** the changed files were run with vitest's JSON reporter and fed to `scripts/check-spec-coverage.ts`'s own `parseVitestJsonReport`, `malformedPartialMarkers`, `nameCarriesId` and `nameCarriesPartialId`. Result: 207 tests, all passed; **0 malformed markers**; DRV-1 partial 4, DRV-2 partial 22, DRV-3 partial 18, DRV-4 partial 6 plus the one requested plain title.
- `bunx tsc --noEmit -p packages/lib/tsconfig.json`: exit 0. apps/web `tsc` from source: exit 0.
- `bunx eslint` on the touched apps/web files: exit 0.
- `knip:check`: the same 6 findings as before, all in untouched files plus a gitignored local config; CI's lint job was green on the previous head.
- Cleanup rule: the integration file deletes in RESTRICT order before and after each test and after the file, then ends the pool.

## Mutation checks
Each by line index; the file was restored and `git status` was clean after each.

| Line broken | Tests that went red |
|---|---|
| org-drive-service-deps.ts `authorization.ok ? authorization.role : null` → `: 'OWNER'` (every non-member an org Owner) | 2 (production-wiring outsider move-in and create) |
| org-drive-service-deps.ts `.for('share')` on the membership read removed | 2 (both race tests) |
| org-drive-ownership.ts `isOrgSlugConflict` constraint check removed | 2 (other constraint, bare 23505) |
| org-drive-ownership.ts retry bound removed | 1 (stops after ORG_SLUG_ATTEMPTS) |
| org-drive-ownership.ts move-out `if (!isOrgAdmin(actorOrgRole))` disabled | 3 |
| org-drive-service.ts move-in `if (!verdict.ok) return verdict;` | 3 |
| org-drive-service.ts `orgVisibility: input.orgVisibility ?? sql\`DEFAULT\`` | 2 |
| org-drive-service.ts move-in `syncOrgMembership` call | 1 |
| org-drive-service-deps.ts `removedOrgRows` forced to `'delete'` | 1 (keep, real sync) |
| account-repository.ts (B6) leave + reassign blanked | 1 (delete-does-not-cascade) |
| membership.ts `removeMember`: the pre-fix plain row delete | 1 (Admin removal reassigns the lead): red before the fix |
| org route DELETE `if (!ORGS_ENABLED)` | 1 |
| drives route: the pre-fix org create without the session check | 1 (bearer refused): red before the fix |

## Not in this PR
- Storage re-attribution (O-9, leaf t1759m6mfxrj5hyaleu1mdqs).
- The POL-5 creation policy reader (Wave E, lyt8275djmdcwlwm8wvk2xa5).
- CLI and MCP parity (Wave G).
- UI for move in/out (Wave F).
- Changelog: nothing user-visible while `ORGS_ENABLED` is false.

https://claude.ai/code/session_014mLodBM7jzXvZxrqjKtkXH


2witstudios:
@codex review

packages/lib/src/services/org-drive-service.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (!verdict.ok) return verdict;
+
+    const slug = await freeOrgSlug(tx, input.orgId, slugify(input.name));
+    const [created] = await tx
+      .insert(drives)
+      .values({
+        name: input.name,
+        slug,
+        ownerId: actorId,
+        orgId: input.orgId,
```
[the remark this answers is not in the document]
2witstudios: Confirmed against the source: `removeMember` (packages/lib/src/organizations/membership.ts, merged from B3 #2665) deletes only the `org_members` row. It skips B6's lead reassignment and the O-8 revocations, so an admin-removed member stays `ownerId` of the org drives they lead. This PR is what makes it reachable, because it is the first code that creates org drives.

The fix is for `removeMember` to run `leaveOrganization(targetId, orgId, tx)` inside its transaction. Those files belong to the B3 and B6 lanes, so I've asked the orchestrator in #org-wallets which lane owns the fix. I'll reply here with the commit either way. Leaving this open.

apps/web/src/app/api/drives/route.ts
```diff
@@ -28,6 +31,9 @@ const createDriveSchema = z.object({
     (v) => (typeof v === 'string' ? v : ''),
     z.string().min(1, 'Missing name')
   ),
+  // Create the drive in an org context (DRV-3). Visibility defaults to Open unless chosen (DRV-4).
+  orgId: z.string().min(1).optional(),
+  orgVisibility: z.enum(ORG_DRIVE_VISIBILITIES).optional(),
```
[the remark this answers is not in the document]
2witstudios: Agreed that org Owners and Admins can't open a Restricted or Private org drive unless they have a row: `resolveOrgDriveAccess` (B2) isn't wired into `getUserAccessLevel`, `getUserDriveAccess` or `listAccessibleDrives` yet. That wiring is its own board leaf, 'Permission resolver integration' (B7, planned after B2 and B4).

In this PR the visibility choice is reachable only through routes that return 404 while `ORGS_ENABLED` is false, so no user can create such a drive yet. I've asked the orchestrator whether to keep `orgVisibility` and let B7 close this, or strip the option from these routes until B7 lands. I'll reply with the outcome. Leaving this open.

packages/lib/src/services/org-drive-service-deps.ts
```diff
@@ -0,0 +1,29 @@
...
+ * org authorization function (ORG-5), membership rows through the materialization sync
+ * (D-OW-6, D-OW-10), and the "who can create org drives" policy seam (POL-5).
+ */
+
+import { requireOrgRole } from '../organizations/authorize';
+import { publishOrgMembershipSyncEvents, syncDriveOrgMembership } from './org-membership-sync';
+import type { OrgDriveServiceDeps } from './org-drive-service';
+
+export const orgDriveServiceDeps: OrgDriveServiceDeps = {
+  getOrgRole: async (_tx, orgId, userId) => {
```
2witstudios: **P1 (independent-review): the role is read outside the move's transaction, so an overlapping account deletion cascades the org drive.** `_tx` is ignored, and `requireOrgRole` reads `org_members` on another pool connection without a lock. `leaveOrganization` locks the org_members row, not the org row, so `lockOrg`'s FOR SHARE serializes nothing. On pg17: move-in reads MEMBER while deleteUser is uncommitted. deleteUser's reassign SELECT sees a personal drive, its cascade waits on the drive lock, then deletes the now-org drive and its pages. Reading the role through `tx` with `FOR SHARE` on the org_members row makes the same interleaving reassign the drive to the org Owner. Full repro in the review body (P1 #1). This applies to createOrgDrive too.

packages/lib/src/services/org-drive-service.ts
```diff
@@ -0,0 +1,263 @@
...
+}
+
+async function lockDrive(tx: OrgDriveTx, driveId: string): Promise<DriveRow | null> {
+  const [row] = await tx.select().from(drives).where(eq(drives.id, driveId)).for('update');
+  return row ?? null;
+}
+
+/**
+ * Share-lock the org row so the actor's membership and the org's ownership cannot change
+ * under the decision (leaving an org and transferring ownership lock it FOR UPDATE).
```
2witstudios: **P1 (part of #1):** this comment is not true for leave. `leaveOrganization` locks `org_members` FOR UPDATE, and `reassignLedOrgDrives` locks `organizations` only through its join to drives already in the org. So this share lock does not block a concurrent leave or delete. See the pg17 repro in the review body.

2witstudios (commented):
[independent-review] agent:ow-irv-2668

Spec review of PR #2668 (lane ow-b5, leaf gj6frj67pnk7ek4kh6zyrzf4) at head `afb760da4bb5280cc8d920a8d382ca131eb0d403`. Base `pu/org-wallets` b2852bcfc, and the merge-base is the tip. I read the Spec, the leaf with all its point-guard sections, Sequence Spec gates 0 to 0e, Retro 0, and D-OW-7, D-OW-10 and D-OW-15. I changed no code.

**Result: 3 P1, 2 P2.** CI's Spec ID coverage job is **red** on this head (P1 #0). The P1 race is reproduced on real pg17 below. Its fix is small, and the same root cause also explains P1 #2.

## Per-ID verdicts

| ID as claimed | Verdict | Evidence (file → test) / gap |
|---|---|---|
| DRV-1 (partial) | **satisfied as partial** | `org-drive-ownership.test.ts` → "DRV-1 (partial) a Home drive can never move into an org, with the drive-guards message", "…Home is refused before any other rule…"; `org-drive-service.integration.test.ts` → "DRV-1 (partial) a Home drive is refused and nothing is written" (asserts orgId stays null and sync is never called). "Lead keeps the Owner role": the move-in test asserts `ownerId` is unchanged. The org-delete RESTRICT belongs to B3. Marking it partial is correct. |
| DRV-2 (partial) | **satisfied as partial** | integration → "moving a drive in keeps its members, roles, pages, envs and publishSubdomain" and "an org Admin moves a drive out…". Publishing after a move is untested, so partial is correct. Note: the preservation test runs with a **fake** sync. I checked B4's `planDriveOrgMembership` (org-membership-sync-core.ts:106-121): it only touches `source='org'` rows, so invite rows with custom roles do survive the real sync too. That holds because the code is built that way, not because a test proves it. |
| DRV-3 (partial) | **satisfied as partial** | integration → "a user outside the org cannot create a drive in it, and nothing is inserted", "the creation policy seam is consulted: admins-only refuses a Member", "concurrent creates of the same name in one org all succeed with distinct slugs". The POL-5 seam is `org-drive-service-deps.ts:26-28` with `TODO(lyt8275djmdcwlwm8wvk2xa5)`. **But every outsider refusal runs on the test-double `getOrgRole`, never on production wiring (P1 #2).** |
| DRV-4 (plain) | **satisfied (via the point-guard ruling)** | integration → "DRV-4 a drive created in an org without a chosen visibility is Open". The title matches the leaf's DRV-4 section exactly and goes through the service. The "Open, Restricted, Private" clause is B1's enum, which the point-guard ruled gate-covered. DRV-4 is not on the allowlist, and nothing was removed from it. |
| DRV-4 (partial) | **satisfied as partial** | integration → "a drive moved in without a chosen visibility is Open, even if it was Private in an earlier stint". The lane's `sql\`DEFAULT\`` mutation goes red, as claimed. |
| O-7 (partial) | **deviates: race, P1 #1** | integration → "deleting the lead's account does not cascade the org drive: the org Owner becomes its lead" passes, and it uses B6's `deleteUser` → `reassignLedOrgDrives`, with no second path (good). But a move-in or create-in-org that overlaps the lead's account deletion hard-deletes the org drive. The lead can also end up a non-member after an overlapping leave. See the P1 #1 repro. |
| O-10 (partial) | **satisfied as partial** | integration (real B4) → "moving out with "keep" leaves org members on the drive as invited members, guest untouched", and "…"remove" revokes the org members and keeps the guest". The route rejects a missing choice with a 400, and the pure core has no default (D-OW-10). "Visibly" is Wave F UI, so partial is correct. |

Title format: every DRV title is well-formed `<ID> (partial)`, and the only plain claim is the DRV-4 title the point-guard asked for. **The O-7 and O-10 titles are not well-formed: see P1 #0.** `scripts/spec-coverage-allowlist.txt` is untouched.

**X-6:** the negatives in this leaf's reach (outsider cannot move in or create, Member cannot move out, Home refused) exist, but only on the test double (P1 #2). The X-6 visibility negative ("non-member cannot see an Open drive of another org") is B7's.

## Checklist (f)-(j) and rules
- (a) Home refusal writes nothing: yes, via drive-guards `'org-move'`. Owner-only move-in: yes.
- (b) Move-out uses B4's `syncDriveOrgMembership` with `removedOrgRows: 'keepAsInvite' | 'delete'`. It does not reimplement the sync.
- (d) The account-deletion flow calls no second reassign path. `handle-drive` refuses delete and transfer of an org drive with a 409 (tested). `drives-status` hides org drives.
- (e) The O-9 stub is loud: `{status:'deferred', leafId:'t1759m6mfxrj5hyaleu1mdqs'}` plus a warn log, asserted in the integration test. The POL-5 seam has a TODO with its leaf id.
- (f) Routes: `PUT`/`DELETE /api/drives/[driveId]/org` are session + CSRF (tested), and answer 404 while B3's `ORGS_ENABLED` constant is false. The constant is imported, not copied (tested for PUT, DELETE and POST with an orgId). They write `data.write` audit events with `org_move_in`/`org_move_out`, and broadcast `updated` after the service's commit. Org-side authorization goes through `requireOrgRole`, but outside the transaction (P1 #1).
- (g) Slug retry: it reads `.cause.code`/`.cause.constraint` from DrizzleQueryError, and 5 attempts is a sound bound. The lane's `isOrgSlugConflict → false` mutant went red in my run. The constraint check and the bound are unproven (P2 #3).
- (i) Cleanup order: drives → invitations → members → orgs → users, then `pool.end()` in afterAll. After my run the DB had organizations=0, org_members=0, drives=0, users=2 (the seeded system users).
- (j) Callers: no existing `drive-service.ts` function changed. `allocatePublishSubdomain` (already exported) gains a caller in `createOrgDrive`. `HomeDriveAction` gains `'org-move'`, and the `Record` forces its message. `moveDriveToOrg`/`moveDriveOutOfOrg` are called only by the new org route, and `createOrgDrive` only by `POST /api/drives`. The PR body does not list callers, but there are none to miss.
- Rules: the pure core is IO-free; no `any`; no env flag; no raw `now()` (JS `new Date()` for updatedAt); no bulk insert. The 3 new lib subpaths have `package.json` exports entries. No migration. Scope stays inside the leaf's named files.
- The unpushed local commit `eb1b56f51` in the lane worktree ("refuse outsiders before disclosing a drive's Home kind…") is NOT reviewed here. It changes the patch, so this review does not carry over to it (0e).

## Test honesty (my runs)
Source-alias vitest, throwaway homebrew **postgresql@17** on 127.0.0.1:5468 (socket `-k /tmp/pi68`), session TZ America/Chicago (non-UTC), migrated to this head. The cluster was stopped and removed afterwards.
```
lib: org-drive-service.integration.test.ts + org-drive-ownership.test.ts + drive-guards.test.ts
 Test Files  3 passed (3)      Tests  105 passed (105)     exit=0
web: [driveId]/org route, drives route, manage-keys-scope, oauth-scope, handle-drive, drives-status
 Test Files  6 passed (6)      Tests  89 passed (89)       exit=0
```
Mutations by line index. The file was restored after each, a no-op control ran first, and `git status` was clean after:

| Line mutated | Result |
|---|---|
| `org-drive-service.ts:203` move-out `if (!verdict.ok) return verdict;` → blank | **red** (1): "the lead who is only an org Member cannot move the drive out" |
| `handle-drive/route.ts:60` `if (drive.orgId) {` → `if (false) {` | **red** (2): "O-7 (partial) refuses to delete/transfer an org drive…" |
| `org-drive-service.ts:123` `isOrgSlugConflict` → `return false;` | **red** (1): concurrent creates |
| `org-drive-service-deps.ts:14` `authorization.ok ? authorization.role : null` → `… : 'OWNER'` (every non-member becomes org Owner) | **SURVIVED** (22/22 green). See P1 #2. |
| `org-drive-service.ts:123` drop `&& cause.constraint === ORG_SLUG_CONSTRAINT` | **SURVIVED**. See P2 #3. |
| `org-drive-service.ts:136` drop `attempt >= ORG_SLUG_ATTEMPTS ||` (unbounded retry) | **SURVIVED**. See P2 #3. |

## P1 #0: Spec ID coverage fails. O-7 and O-10 are not requirement IDs, so the gate rejects their `(partial)` markers
`scripts/check-spec-coverage.ts:77` sets `REQUIREMENT_ID_PATTERN = /\b(ORG|DRV|SEAT|WAL|MON|SPEND|POL|SEC|AUD|UI|X)-\d+\b/`. `O-n` are the Spec's open items (decided as D-OW-n), not requirement IDs. For every title of the form `O-7 (partial) …` or `O-10 (partial) …`, `malformedPartialMarkers` finds no ID before the marker and fails the job. There are 15 such titles, in drives-status, handle-drive, the org route, org-drive-ownership and the integration file. Job log (run 35218438670, job 105203130768), lines 312-327:
```
FAIL: malformed partial marker in apps/web/src/app/api/account/drives-status/__tests__/route.test.ts: "O-7 (partial) an org drive the person leads needs no action before account deletion and is not listed" — write exactly "<ID> (partial)" (one ASCII space, lowercase, no inner spaces)
FAIL: malformed partial marker in packages/lib/src/services/__tests__/org-drive-service.integration.test.ts: "O-7 (partial) deleting the lead's account does not cascade the org drive: the org Owner becomes its lead" — …
… (13 more, O-7 and O-10) …
spec-coverage: FAILED
##[error]Process completed with exit code 1.
```
Fix (the lane's call): drop ` (partial)` after `O-n`/use `D-OW-n` as plain traceability text, or put a real requirement ID in front of the marker (e.g. the move-out keep/remove tests are DRV-2 behaviour). No real requirement ID should be claimed plain in the process. The DRV rows in the same log are right: DRV-1/2/3 `allowlisted partial`, DRV-4 `covered` (log lines 226-229).

## P1 #1: move-in and create-in-org read the org role outside the transaction, so an overlapping account deletion cascades the org drive
`orgDriveServiceDeps.getOrgRole` ignores `tx` and calls `requireOrgRole`, which reads `org_members` on a separate pool connection with no lock. The `lockOrg` comment (org-drive-service.ts:104) says leave "locks it FOR UPDATE", but it does not. `leaveOrganization` locks the **org_members row**. `reassignLedOrgDrives` locks `organizations` only through its join to drives the user **already** leads in that org. So the `FOR SHARE` on the org row serializes nothing against leave or delete.

Interleaving: A = `moveDriveToOrg`, B = `accountRepository.deleteUser` for the same lead.
1. A locks the drive `FOR UPDATE`, locks the org `FOR SHARE`, and reads the role through requireOrgRole. B's membership delete is not committed yet, so A sees MEMBER.
2. B deletes the org_members row. B's reassign `SELECT … JOIN organizations … FOR UPDATE` sees the drive still personal and returns nothing. B's `DELETE users` cascade then waits on A's row lock.
3. A writes `orgId` and commits. B resumes, its cascade re-checks `ownerId` (still the lead), and it **deletes the org drive and its pages**. This is the exact failure the leaf's Landing order section exists to prevent.

Reproduced on pg17, raw SQL in the same statement order as both code paths:
```
requireOrgRole-style read on a separate connection (B uncommitted): [MEMBER]
A committed / B committed
drive after: []  page after: []                              <- org drive + page hard-deleted
```
Same script with the role read moved INSIDE A as `SELECT role FROM org_members WHERE … FOR SHARE`:
```
B's reassign returned: race_drive
drive after: [race_drive|race_org|race_owner]  page after: [race_page]   <- reassigned to the org Owner, nothing lost
```
With B first, A then finds no drive and returns 404, which is correct. `createOrgDrive` has the same root cause: a concurrent leave or delete leaves a non-member leading, or cascades the new drive. A plain leave (no delete) leaves a **non-member lead** (O-7 / D-OW-7).
**Fix shape (lane's call):** read the membership through `tx` with a row lock on the org_members row (`FOR SHARE`/`FOR KEY SHARE`), for example a tx-aware `findMembershipRole` passed as `requireOrgRole`'s `deps`. Keep `requireOrgRole`'s decision so ORG-5's single authorization function is still the one used. If the opposite lock order hits a deadlock, Postgres aborts one side, which fails closed. Add a real-pg test that deletes the lead while a move-in holds the drive lock.

## P1 #2: no refusal runs through the production role wiring; a mutant that makes every non-member an org Owner survives
Every refusal test in the integration file (O-7 outsider move-in, DRV-3 outsider create, DRV-2 Member move-out) uses the test-double `getOrgRole`. The double reads inside `tx` (line 49), and that is also why it could not show P1 #1. The production-wiring block exercises only success paths. The route tests mock the service. Replacing `org-drive-service-deps.ts:14` with `… : 'OWNER'` keeps all 22 integration, 22 pure and 89 web tests green. On an authorization boundary, the O-7/DRV-3 refusal claims are therefore proven only for the double. **Needed:** under `orgDriveServiceDeps`, at least outsider move-in → `NOT_ORG_MEMBER`, outsider create → `NOT_ORG_MEMBER` with nothing inserted, and Member move-out → `NOT_ORG_ADMIN`.

## P2 #3: the slug retry's specificity and bound are unproven
Removing the constraint-name check, or removing the attempt bound, leaves the suite green. A unit test over `retryOnOrgSlugConflict` would pin both properties the brief asks for: "retries only on the (orgId, slug) unique violation" and "is bounded". One case: a `DrizzleQueryError`-shaped error with `cause.code='23505'` and a different constraint is thrown once and not retried. Another: a persistent org-slug conflict stops after 5 attempts.

## P2 #4: create-in-org is reachable with an MCP bearer token, while the org routes are session-only
`POST /api/drives` authenticates with `AUTH_OPTIONS_WRITE = { allow: ['session', 'mcp'] }`, and the new `orgId` branch inherits that. An unscoped MCP token can create org drives today (once ORGS_ENABLED flips). Meanwhile PUT/DELETE `/org` are deliberately session-only, and the PR says CLI/MCP parity is Wave G (X-1). Either refuse `orgId` for non-session auth until Wave G (with a test), or state that MCP create-in-org is intended. Smaller notes, not blocking: `orgVisibility` without `orgId` is silently ignored. A non-member of an org gets a 403 here, while B3's `decideOrgRole` convention is a 404 for non-members. This PR does not leak existence (a missing org also gets a 403), but it is inconsistent with B3.

## CI (Spec ID coverage)
Run 35218438670 on afb760da4: Lint & TypeScript **pass**, Unit Tests **pass** (34m54s), E2E **pass**, **Spec ID coverage FAIL** (P1 #0). Lines for this PR's IDs: `DRV-1 allowlisted partial: …[driveId]/org/__tests__/route.test.ts, …`, `DRV-2 allowlisted partial: …`, `DRV-3 allowlisted partial: …drives/__tests__/route.test.ts, …`, `DRV-4 covered …org-drive-service.integration.test.ts…`. Summary: `covered 10/87, allowlisted-missing 77, MISSING 0`. O-7 and O-10 have no gate row because they are not requirement IDs, which is exactly the failure. Their `(partial)` claims are traceability text only, and the gate does not count them.

packages/lib/src/services/__tests__/org-drive-service.integration.test.ts
```diff
@@ -0,0 +1,416 @@
...
+const northwind = createId();
+const userIds = [jono, priya, marcus, lena, chris];
+
+const syncCalls: OrgMembershipSyncCall[] = [];
+/** Records the order of commit-visible effects: a publish must run after the move committed. */
+const published: Array<{ call: OrgMembershipSyncCall; orgIdAtPublish: string | null }> = [];
+
+const deps: OrgDriveServiceDeps = {
+  // Test double for the org role lookup; production wires the org repository.
+  getOrgRole: async (tx, orgId, userId) => {
```
2witstudios: **P1 (independent-review, test honesty):** every refusal test (outsider move-in, outsider create, Member move-out) goes through this double. The double reads inside `tx`, which production does not, and that difference also hides P1 #1. The production-wiring block covers only success paths. Mutant `org-drive-service-deps.ts:14` `… : null` → `… : 'OWNER'` (every non-member becomes org Owner) leaves all 22 integration, 22 pure and 89 web tests green. Please add the refusal tests under `orgDriveServiceDeps`.

packages/lib/src/services/org-drive-service.ts
```diff
@@ -0,0 +1,263 @@
...
+}
+
+const ORG_SLUG_CONSTRAINT = 'drives_org_slug_unique';
+const ORG_SLUG_ATTEMPTS = 5;
+
+/** Drizzle rethrows driver errors with the pg error on `.cause`. */
+function isOrgSlugConflict(error: unknown): boolean {
+  if (error === null || typeof error !== 'object') return false;
+  const cause = (error as { cause?: { code?: unknown; constraint?: unknown } }).cause;
+  return cause?.code === '23505' && cause.constraint === ORG_SLUG_CONSTRAINT;
```
2witstudios: **P2 (independent-review):** neither 'retries only on drives_org_slug_unique' nor 'bounded' is pinned by a test. Removing `&& cause.constraint === ORG_SLUG_CONSTRAINT` survives, and so does removing `attempt >= ORG_SLUG_ATTEMPTS ||` (line 136). Add a unit test: another 23505 constraint is thrown without a retry, and a persistent conflict stops after 5 attempts.

apps/web/src/app/api/drives/route.ts
```diff
@@ -146,14 +152,27 @@ export async function POST(request: Request) {
...
+  }
 
   try {
     if (isReservedDriveName(name)) {
       return NextResponse.json({ error: 'Cannot create a drive with that name.' }, { status: 400 });
     }
 
-    const newDrive = await createDrive(userId, { name });
+    let newDrive: DriveWithAccess;
+    if (orgId !== undefined) {
```
2witstudios: **P2 (independent-review):** this org branch inherits `AUTH_OPTIONS_WRITE = ['session','mcp']`, so an unscoped MCP bearer can create org drives. Meanwhile PUT/DELETE `/org` are session-only, and the PR defers MCP parity to Wave G (X-1). Either refuse `orgId` for non-session auth until then (with a test), or state that it is intended.

packages/lib/src/services/org-drive-service-deps.ts
2witstudios: Fixed in 977f84070 (head 2ba9e1038). `getOrgRole` now reads the membership through the move's `tx` via `requireOrgRole`'s `findMembershipRole` dep, with `FOR SHARE` on the `org_members` row. The decision is still `requireOrgRole`'s (ORG-5). This covers both move-in and create.

Your race is now an integration test: session B deletes Marcus's `org_members` row, runs the reassign SELECT, and holds both uncommitted while the service starts; then B deletes the user. The test runs for both move-in and create, and asserts that a committed org drive is never gone. Before the fix both cases failed with the drive deleted. After it both pass, 5/5 runs. Removing `.for('share')` turns both red again. If the two sides deadlock, Postgres aborts one, which fails closed. Leaving open for verification.

packages/lib/src/services/org-drive-service.ts
2witstudios: Fixed in 977f84070 (head 2ba9e1038). The comment was wrong. It now says the org-row `FOR SHARE` only guards org deletion and ownership transfer (both lock the org row `FOR UPDATE`), and that membership is guarded by `getOrgRole`'s `FOR SHARE` on the `org_members` row, which leave and account deletion delete. Leaving open.

packages/lib/src/services/__tests__/org-drive-service.integration.test.ts
2witstudios: Fixed in 977f84070 (head 2ba9e1038). New describe 'with the production wiring: authorization' runs on `orgDriveServiceDeps` (real `requireOrgRole`, real pg). An outsider's move-in gets `NOT_ORG_MEMBER` and the drive stays personal; an outsider's create gets `NOT_ORG_MEMBER` with no row inserted; a Member moving a drive out gets `NOT_ORG_ADMIN` and the drive stays in the org.

Your mutant (`: null` → `: 'OWNER'` on the deps line) now turns the outsider move-in and create tests red, 2 of 28. The Member move-out refusal comes from `decideMoveDriveOutOfOrg`, whose admin-check mutant turns 3 tests red. Leaving open.

packages/lib/src/services/org-drive-service.ts
2witstudios: Fixed in d2d6db4fa (head 2ba9e1038). `retryOnOrgSlugConflict` and `isOrgSlugConflict` moved to the pure module, with `ORG_DRIVE_SLUG_CONSTRAINT` and `ORG_SLUG_ATTEMPTS = 5`. Four unit tests use DrizzleQueryError-shaped errors with the pg error on `.cause`:
- it re-runs after an org-slug conflict and returns the result;
- a 23505 on another constraint is thrown once, with no retry;
- a bare 23505, a 40P01 and a plain Error are thrown once, with no retry;
- a persistent conflict stops after exactly 5 attempts. A runaway guard makes an unbounded retry fail instead of hang.

Mutants: dropping the constraint check turns 2 red; dropping the bound turns 1 red. Leaving open.

apps/web/src/app/api/drives/route.ts
2witstudios: Fixed in 080223d7b (head 2ba9e1038), per the point-guard ruling. `POST /api/drives` with an `orgId` now answers 403 for any request whose auth is not a session: 'Creating a drive in an organization requires a signed-in session; tokens cannot do this yet.' It runs after the `ORGS_ENABLED` 404 and never reaches `createOrgDrive` or `createDrive`. The test fails before the fix and passes after. A personal drive create over MCP is unchanged. Leaving open.

apps/web/src/app/api/drives/[driveId]/org/route.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  auditRequest(request, {
+    eventType: 'data.write',
+    userId,
+    resourceType: 'drive',
+    resourceId: driveId,
+    details: { operation, orgId: result.orgId, storageReattribution: result.storageReattribution.status },
+  });
+
+  const recipients = await getDriveRecipientUserIds(driveId);
```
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/drives/[driveId]/org/route.ts]: Fixed in 631556149 (head 2ba9e1038). The recipient lookup and broadcast after a committed move are wrapped in try/catch and logged as a warning. New route test: `getDriveRecipientUserIds` rejects, and the move still answers 200 with the audit event written and no broadcast attempted. It failed before the fix and passes after. Leaving open.

packages/lib/src/services/org-drive-service.ts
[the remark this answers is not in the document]
2witstudios: Fixed in 2ba9e1038. `removeMember` (membership.ts) now runs B6's `leaveOrganization(targetId, orgId, tx, { actor: removing Admin })` in its transaction instead of deleting only the `org_members` row. The removed member's led org drives pass to the org Owner with an audit event, and the leave revocations run too.

New integration test: an org Admin removing the lead makes the org Owner the lead and keeps the drive, with no org row left for the removed member. It failed before the fix (ownerId stayed Marcus) and passes after. B3's `org-services.integration.test.ts` still passes 21/21, and B6's `leave.integration.test.ts` 9/9. Leaving open.

apps/web/src/app/api/drives/route.ts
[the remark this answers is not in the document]
2witstudios: Update at head 2ba9e1038: no code change for this one. Wiring org Owner/Admin access to Restricted and Private org drives into `getUserAccessLevel`, `getUserDriveAccess` and `listAccessibleDrives` belongs to the B7 leaf, 'Permission resolver integration and access-only listing', which runs after B2 and B4.

The routes that let anyone choose a visibility return 404 while `ORGS_ENABLED` is false, so no user can create such a drive before B7 lands. `orgVisibility` stays because DRV-4 requires the choice at create and move-in. Leaving open for the orchestrator's ruling.

2witstudios (commented):
[independent-review] agent:ow-irv-2668 (re-review)

Re-review of PR #2668 at head `2ba9e1038e1d09d4985fc34fd06303254b61aea6`, covering `afb760da4..2ba9e1038`: commits eb1b56f51, 631556149, d2d6db4fa, 977f84070, 080223d7b and 2ba9e1038. The merge-base is still the `pu/org-wallets` tip b2852bcfc, and nothing new from the tip was merged in. I changed no code and resolved no threads.

**Result: all 5 findings from review 5235664932 are FIXED. No new P1 or P2.** CI run 35223277380 on this head is fully green, and **Spec ID coverage: OK**.

## Verdict per finding

| Finding | Verdict | Evidence at 2ba9e1038 |
|---|---|---|
| P1 #0: O-n `(partial)` markers break Spec ID coverage | **FIXED** | Every `O-7 (partial)`/`O-10 (partial)` title is now `D-OW-7 …`/`D-OW-10 …`, which is traceability text that matches no requirement-ID pattern. The keep/remove pure test was retitled `D-OW-10 …`, and the new negatives use `DRV-2 (partial)`/`DRV-3 (partial)`. No new plain requirement-ID claim. CI log below. |
| P1 #1: role read outside the tx, so an overlapping deleteUser cascades the org drive | **FIXED** | `org-drive-service-deps.ts:12-31`: the decision is still `requireOrgRole`'s (ORG-5), and only `findMembershipRole` is swapped for a read through `tx` with `.for('share')` on the org_members row. The `lockOrg` comment is corrected. Probes below. |
| P1 #2: refusals only tested through the double; the OWNER mutant survived | **FIXED** | New describe "with the production wiring: authorization" runs on `orgDriveServiceDeps`. The OWNER mutant now goes red (below). |
| P2 #3: slug retry specificity and bound unproven | **FIXED** | `retryOnOrgSlugConflict` moved to the pure module (`org-drive-ownership.ts:159-187`) with 4 unit tests: retry-then-success, another constraint never retried, non-23505 and bare-23505 never retried, and stop at `ORG_SLUG_ATTEMPTS` (5) with the error rethrown. Both mutants now go red. |
| P2 #4: create-in-org reachable with an MCP bearer | **FIXED** | `drives/route.ts:160-166` returns 403 for `orgId` whenever `auth.tokenType !== 'session'`, before any service call. Test: "DRV-3 (partial) a bearer-token request cannot create a drive in an org until CLI and MCP parity in Wave G". The mutant goes red. |

## Probes (same as round 1)
Throwaway homebrew **postgresql@17** on 127.0.0.1:5468, session TZ America/Chicago, migrated to 2ba9e1038. The cluster was stopped and removed afterwards, and `git status` was clean.

**Suites:**
```
lib: org-drive-service.integration (28), org-drive-ownership (27), drive-guards (61), leave.integration (9), leave (6)
 Test Files  5 passed (5)      Tests  131 passed (131)     exit=0
lib (integration config): org-services.integration.test.ts   Tests  21 passed (21)   <- B3 suite, since removeMember changed
web: [driveId]/org route (12), drives route (34), manage-keys-scope, oauth-scope, handle-drive, drives-status, orgs/org-routes (113)
 Test Files  7 passed (7)      Tests  204 passed (204)     exit=0
after run: organizations=0 org_members=0 drives=0 users=2
```

**Line-index mutations.** A no-op control ran first, and the file was restored after each:

| Mutant | Result |
|---|---|
| `org-drive-service-deps.ts:30` `… : null` → `… : 'OWNER'` (the round-1 survivor) | **red** (2): "D-OW-7 a drive owner outside the org cannot move their drive in (real requireOrgRole)", "DRV-3 (partial) a user outside the org cannot create a drive in it, and nothing is inserted (real requireOrgRole)" |
| `org-drive-service-deps.ts:26` `.for('share');` → `;` (in-tx read, no row lock) | **red** (2): "D-OW-7 a move-in / a create overlapping the lead's account deletion never leaves a committed org drive deleted" |
| `org-drive-ownership.ts:171` drop `&& cause.constraint === ORG_DRIVE_SLUG_CONSTRAINT` | **red** (2): "never retries a unique violation on any other constraint", "…including a bare 23505 without the constraint" |
| `org-drive-ownership.ts:184` drop `attempt >= ORG_SLUG_ATTEMPTS \|\|` | **red** (1): "stops after ORG_SLUG_ATTEMPTS attempts of a persistent conflict and rethrows it" |
| `drives/route.ts:161` session check → `if (false) {` | **red** (1): the bearer-token test |
| `membership.ts:162` `leaveOrganization(…)` → the old bare `org_members` delete | **red** (1): "D-OW-7 an org Admin removing the lead from the org makes the org Owner the lead and keeps the drive" |

**deleteUser race on pg17, raw SQL in both orderings, with the new lock shape (role read inside A with `FOR SHARE`):**
```
A first (move-in holds the drive, then deleteUser):  drive after: [race_drive|race_org|race_owner]  page after: [race_page]   <- reassigned to the org Owner, nothing lost
deleteUser first (leave uncommitted, then move-in):  drive after: []  <- A finds no drive (404); the drive was still personal when the account went, which is correct
```
**Is the lane's race test vacuous?** No. I instrumented it temporarily (restored after), and it passed 6/6 runs. Each run's outcome: move-in `{"error":"40P01"}` (Postgres picks the move as the deadlock victim), create `{"ok":false,"code":"NOT_ORG_MEMBER"}`. Its only assertion is "a committed org drive still exists", so in this ordering the move-in case asserts nothing. Removing `FOR SHARE` still turns both cases red, so it does guard the fix. The A-first ordering, where the drive gets reassigned, is covered only by my SQL probe, not by a test. That is not blocking.

## Notes (not findings)
- **Deadlock surfaces as a 500.** When a move-in overlaps its lead's account deletion, the move is aborted with 40P01, and the route maps that to a 500 "Failed to move drive into organization". It fails closed, nothing is written, and the lead is deleting their account anyway, so this is acceptable. A 409 would be kinder.
- **2ba9e1038 edits B3's `membership.ts`.** This file is outside the leaf's named files. `removeMember` now runs B6's `leaveOrganization` in its transaction (Codex thread, D-OW-7/D-OW-8). The logic is correct: roles are locked first, the Owner target was already refused by `decideMemberRemoval`, and the mutant goes red. Two loose ends: (a) the reassignment activity row records `reason: 'left_org'` for a removal an Admin performed, and `LeadReassignmentReason` has no removal value (AUD-1 accuracy, for a follow-up); (b) the doc comment on `DELETE /api/orgs/[orgId]/members/[userId]` still says the cascade "is B6's leave flow", which is now stale. Whether this scope expansion is accepted is the orchestrator's call.
- **Refusal order changed (eb1b56f51).** Ownership is checked before Home on move-in, and org authority before `NOT_IN_ORG` on move-out, so outsiders learn nothing about the drive. The DRV-1 test was retitled to match, and a Home drive is still refused for its owner (integration "DRV-1 (partial) a Home drive is refused and nothing is written" passes).
- **631556149** makes the post-commit broadcast best-effort: a committed move returns 200 when the recipient lookup fails, and a test covers it. Correct.
- **Codex thread on `drives/route.ts:36`** (org Owner/Admin access to Restricted/Private drives) has no code change. The lane defers it to the resolver lane (B7, ORG-4). I agree it is B7's scope, and orgs are off while `ORGS_ENABLED` is false. The thread stays open for the orchestrator.

## CI on 2ba9e1038
Run 35223277380 (headSha 2ba9e1038): Lint & TypeScript **pass**, Unit Tests **pass**, E2E **pass**, **Spec ID coverage pass** (job 105219672256). Log:
```
226: DRV-1     allowlisted  partial: apps/web/src/app/api/drives/[driveId]/org/__tests__/route.test.ts, …
227: DRV-2     allowlisted  partial: …[driveId]/org/__tests__/route.test.ts, …org-drive-ownership.test.ts, …org-drive-service.integration.test.ts
228: DRV-3     allowlisted  partial: apps/web/src/app/api/drives/__tests__/route.test.ts, …org-drive-ownership.test.ts, …org-drive-service.integration.test.ts
229: DRV-4     covered      …org-drive-service.integration.test.ts, partial: …
311: covered 10/87, allowlisted-missing 77, MISSING 0
312: spec-coverage: OK
```
No malformed-marker lines.

apps/web/src/app/api/drives/route.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

packages/lib/src/services/org-drive-service.ts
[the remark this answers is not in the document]
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

apps/web/src/app/api/drives/[driveId]/org/route.ts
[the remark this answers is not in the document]
Comment (2witstudios) [reply apps/web/src/app/api/drives/[driveId]/org/route.ts]: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

packages/lib/src/services/org-drive-service-deps.ts
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

packages/lib/src/services/org-drive-service.ts
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

packages/lib/src/services/__tests__/org-drive-service.integration.test.ts
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

packages/lib/src/services/org-drive-service.ts
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

apps/web/src/app/api/drives/route.ts
2witstudios: [point-guard] Verified FIXED at 2ba9e1038 by independent re-review 5236218949 (5/5 FIXED, 0 new P1/P2; on pg17: removing the FOR SHARE role read reproduces the deleteUser cascade race and goes red, the all-OWNER role mutant goes red through production wiring, slug retry is limited to the (orgId, slug) constraint and bounded, bearer-token org drive creation is refused, removeMember cascade tested) and CI 11/11 on this head. Codex's three threads are covered by the same push. Resolving.

PR #2670, opened 2026-09-17
feat(auth): revoke Sign in with Apple tokens on account deletion (5.1.1(v), TN3194)

2witstudios:
## Why

App Store Guideline 5.1.1(v) and Apple TN3194 require an app that offers Sign in with Apple to revoke the user's Apple tokens when they delete their account. PageSpace did not, so a deleted account stayed listed under the user's Apple Account → Sign in with Apple, and signing up again skipped Apple's name/email consent. This is the last compliance gap on the iOS review branch (#2667). Task page: `u4erzl5g7lppr6rqyp6ubz3e`; the plan and founder decisions are recorded there.

## What

**Capture (acceptance criteria 1 and 5)**
- The iOS client initializes `@capgo/capacitor-social-login` with `useProperTokenExchange: true` and sends Apple's `authorizationCode`. The Swift support is already in the shipped 7.20.0 binary, and the shell loads this JS from pagespace.ai, so **no new iOS build is needed**.
- `/api/auth/apple/native` and `/api/auth/apple/callback` (web, desktop and iOS-web flows) pass the code to `captureAppleRefreshToken` together with the **verified** id_token's `aud` (the client that issued the code) and `sub`.
- The call is made only after the response is built. It is **not awaited** and has its own catch, so sign-in never waits on Apple or fails because of it. Route tests prove the response returns while a slow exchange is still pending.
- The capture calls `/auth/token` with an ES256 client_secret, refuses a token whose `sub` differs from the signed-in user, and stores the refresh token **encrypted** (`encryptField`) in the new `apple_sign_in_tokens` table: one row per (user, client), FK cascade to `users`.
- If the signing key is not configured, the capture is skipped and sign-in behaves exactly as before.

**Revoke at deletion (acceptance criteria 2, 3 and 6)**
- `lodgeAndEnqueueErasure` runs after the DSR row exists and **before** the job is queued, so the worker's `delete-user` cascade cannot race it. It revokes each stored token with its own client, deletes all of them whatever Apple answers, and records a `revoke-apple-tokens` step on the DSR row.
- Neither Apple nor a database failure can stop the deletion. This runs in web because the processor holds no `ENCRYPTION_KEY` (founder decision Q1); the processor skips the step, as it already does for `stripe-customer`.
- `DELETE /api/account` returns `appleSignIn: revoked | manual | none`.
- The web admin GDPR escalation passes the subject's Apple link. The admin app's legacy DSAR delete revokes before `deleteUser`.
- UI:
  - The delete dialog uses `GET /api/account/apple-sign-in`. It says either "we'll also disconnect PageSpace from Sign in with Apple" or lists the Stop Using steps.
  - **Every** deletion now lands on the public in-app page `/auth/account-deleted`, never `/`, which is the marketing site with a pricing nav inside the iOS WebView (Guideline 3.1.1). That page shows the Stop Using steps when the outcome is `manual` and links back to sign-in.

**Apple server-to-server notifications (acceptance criterion 4)**
- `POST /api/auth/apple/notifications` accepts the payload only after checking Apple's RS256 signature against the JWKS, the issuer, the audience (`APPLE_CLIENT_ID` / `APPLE_SERVICE_ID`) and expiry.
- `consent-revoked` and `account-delete(d)` revoke all sessions and device tokens, bump `tokenVersion` and discard stored Apple tokens. The account is kept (founder decision Q2). Email events are acknowledged as no-ops.
- Responses: 400 when the payload can't be verified, 200 when it is verified, 500 when acting on a verified notification fails (so Apple retries).
- The route has a per-IP rate limit and calls `auditRequest`.

**Also**
- `verifyAppleIdToken` now returns the verified `audience`.
- `accountRepository.findById` now selects `appleId`.
- The erasure plan has a new step, `revoke-apple-tokens`, which is best-effort and runs before `delete-user`.
- `apple_sign_in_tokens` is registered as credential material in the Art 15 export registry.
- The new lib subpaths are added to `exports`.
- Env validation, `.env.example`, CHANGELOG and the iOS README (portal/secret steps plus a submission-checklist item) are updated.

**Orchestrator review round (3fe40fd3d..9ea4781d8)**
- **Apple JWKS availability (major):** `apple-signin-auth` refetched Apple's keys for every unknown `kid`, with no timeout and no single-flight, and it emptied its shared cache before parsing the response. The unauthenticated notification endpoint could therefore trigger unbounded fetches, and once Apple throttled us, a failed or malformed response would break every Apple sign-in. PageSpace now owns this:
  - `apple-jwks`: a 5s timeout, one shared fetch for concurrent lookups, at most one refresh per 5 minutes for an unknown kid, and the last good key set is kept whenever a refresh fails.
  - `apple-jwt`: RS256 only, checking kid, signature, issuer, audience, required `exp` and `nbf`. It serves both `verifyAppleIdToken` and notifications.
  - `apple-signin-auth` is removed (including its `OSS-INVENTORY.md` rows).
- **Capture race:** after storing a token, the capture checks again for an erasure that started in between, and revokes and discards the token if one exists.
- **Replayed notifications:** consent-revoked/account-delete(d) events older than the user's latest Apple token capture are ignored.
- **Android:** the native route never exchanges an Android code, because that code needs the web flow's `redirect_uri`.
- **Test teardown:** tolerates suites that mock `@pagespace/db/db` without a pool.
- **Admin DSAR delete:** records the manual Sign in with Apple follow-up durably in the audit log (`apple_sign_in_manual_followup`).

**Test infrastructure (found by this PR's CI):** the lib integration step failed with Postgres 53300 "too many clients already". Each isolated integration file builds its own `@pagespace/db` pool (max 10, 10-minute idle timeout) and almost none end it, so connections pile up across ~40 files, and the file that runs last crosses `max_connections=100`; this PR's new store suite became that file. A new integration-only setup file, `src/test/integration-db-teardown.ts`, ends each file's pool, with a guard test. Local repro against a 100-connection Postgres: before, peak 82 client backends and the last file was refused; after, peak 9, 37/37 files pass, zero refusals.

**Not done, by decision:** no daily refresh-token validation (founder decision Q3). The processor's missing `ENCRYPTION_KEY`, which affects the existing `revoke-integrations` step, is being triaged separately (Q4).

## Production steps (Jono) — the code ships dark until these are done
1. Apple Developer → Keys → + → Sign in with Apple (Primary App ID `ai.pagespace.ios`) → download `AuthKey_<KEYID>.p8` (one download only) and note the Key ID.
2. Identifiers → `ai.pagespace.ios` → Sign in with Apple → Server-to-Server Notification Endpoint = `https://pagespace.ai/api/auth/apple/notifications`.
3. `flyctl secrets set -a pagespace-web APPLE_SIGN_IN_KEY_ID=<KEYID> APPLE_SIGN_IN_PRIVATE_KEY="$(cat AuthKey_<KEYID>.p8)"` (`APPLE_TEAM_ID` is already set). Optionally set the same on `pagespace-admin`.
4. Migration `0295_apple_sign_in_tokens` runs with the deploy.

## Verification
- **Tests:** strict TDD, with each new test seen failing for the right reason before the code was written.
  - **Lib:** 49 tests in `src/auth/apple`. They use a throwaway P-256 key for the client secret, real RS256 signatures served through the library's JWKS fetch seam for notifications, and real `encryptField` ciphertext.
  - **Real Postgres:** 4 integration tests for the store (upsert conflict key, one row per client, scoped delete, users cascade).
  - **Web, processor and admin:** route and component tests for every criterion.
- **Mutation sweep (initial):** 13/13 mutations killed:
  - sub binding and the unconfigured skip in the capture
  - delete-regardless and the per-token client in revoke
  - the notification audience and its event set
  - the client-secret expiry
  - awaiting the capture in the native route and in the callback route
  - the callback `redirect_uri`
  - the worker's step filter
  - revoke-before-enqueue in the lodge
  - the post-deletion destination
- **Independent review** (Codex quota exhausted, CodeRabbit skips non-default bases): 0 blockers. 8 findings fixed in 7c0dfabf1: concurrent revokes with a 5s timeout; no capture stored while an erasure is active; notifications without `exp` rejected; P-256 key required, with a warning when the key is unusable; honest DSR step; `appleSignIn` also on the in-progress retry and the admin DSAR delete. 5 kept, with reasons in the PR comment. A further 6 mutations on these fixes were all killed (19/19 in total).
- **Lint and types:** `bunx eslint` on every touched file in web, lib, admin and processor exits 0. `tsc --noEmit` on `packages/lib` (tests included) exits 0. The full monorepo typecheck was not run locally (shared machine); it runs in CI (`test.yml` dispatched on this branch, since CI doesn't run automatically for this base).

- **Second independent review** (of 7c0dfabf1..613bc240e): no blockers. Two minor fixes in b39bacfe7: the unusable-key warning now fires once, and the Apple notice is cleared before it is refetched.
- **CI on head 9ea4781d8:** Test Suite 35235779837 and Security Tests 35235782776 are running.

https://claude.ai/code/session_011S9PvEGXAwtPETxH7ZWt2H


2witstudios:
**Independent review (Codex review quota exhausted; CodeRabbit skips non-default bases) — dispositions, fixed in 7c0dfabf1**

Fixed (each with a failing test first):
1. Admin-app DSAR delete gave no signal when tokens were discarded unrevoked → response now carries `appleSignIn` (`manual` when not fully revoked). Setting the Apple key on `pagespace-admin` stays an optional production step (iOS README).
3. Deletion could wait 2×10s on a slow Apple → revokes run concurrently with a 5s timeout (exchange keeps 10s, off the request path).
4. The "erasure already in progress" 202 omitted `appleSignIn` → Apple-linked users get `manual`.
5. Notification without `exp` would never expire → rejected (`missing_exp`).
7. A capture finishing after the deletion request could re-store a token the cascade later deletes unrevoked → capture skips storing while an erasure is active.
8. A set-but-unusable key failed silently; non-P-256 EC keys accepted → warning logged; P-256 required.
10. Unconfigured-key discard was recorded `skipped`; a failed step write could add a contradictory `failed` row → recorded `failed` (authorization still live); step write failures are logged, never relabel the outcome.
11. Stale Apple notice when the status request later returns non-OK → reset.

Kept, with reasons:
2. Revoke before enqueue: after-enqueue races the worker's `delete-user` cascade (rows gone before revocation), which would silently break the guideline requirement. If the enqueue fails the account survives with Apple consent revoked; the user just consents again next sign-in. Documented in `request-erasure.ts`.
6. Unknown-`kid` JWKS refetch: same behaviour as the existing sign-in id_token verification; per-IP limit in front. Not widening scope here.
9. `after()` vs `void`: web runs as a long-lived Node server on Fly, where the promise completes; the route tests pin that the response never waits.
12. `DEFAULT now()` on the new table: inserts always set both timestamps explicitly in UTC; the schema follows the existing `passkeys` pattern.
13. `/api/auth/apple/` is cloud-only in tenant mode: pre-existing (Apple sign-in itself is blocked there), out of scope.

2witstudios:
**CI fix + second independent review.**

- Root cause of the lib integration failure on 7c0dfabf1 (Postgres 53300 'too many clients'): each isolated integration file builds its own `@pagespace/db` pool (max 10, 10-min idle) and never ends it; connections accumulated across ~40 files and this PR's new store suite, running last, crossed `max_connections=100`. Fixed in 613bc240e with an integration-only setup file that ends each file's pool (guard test). Local repro before/after: peak 82 → 9 client backends. CI on 613bc240e: Test Suite 35226542420 green (lib integration 37 passed, apple-token-store 4/4), Security 35226545660 green.
- Second independent review of 7c0dfabf1..613bc240e: no blockers; verified hook ordering (setup-file afterAll runs after suite cleanup under vitest's default stack order), no import cycles, `pool.ending` typing. Two minors fixed in b39bacfe7 (warn once per unusable Apple key; clear the Apple notice before refetching). Kept: the capture check/upsert race window (a sign-in exchange finishing in the milliseconds between DSR creation and the revoke read) — the users cascade still removes the row, and a post-upsert re-check would add a second query to every Apple sign-in for a sub-second window.
- CI for b39bacfe7: Test Suite 35229391799, Security 35229395161 (running).

PR #2674, opened 2026-09-17
fix(web): free users can switch AI providers without hitting the tier gate

2witstudios:
## Summary
Free users got `403 Subscription required` ("Failed to update provider") when picking most AI providers, so only a handful of providers were switchable.

**Root cause**: on provider switch the client sent the provider's *first catalog entry* as the model (`getDefaultModel`). For most providers that first entry is a paid flagship (OpenAI → `gpt-5.6-sol-pro`, NVIDIA → `nemotron-3-nano-30b-a3b`, Cohere → `command-r-08-2024`), and the PATCH tier gate (`requiresProSubscription` → `isModelAllowedForTier` → `FREE_TIER_MODELS`) rejected it. Only providers whose *first-listed* model is free-tier eligible (Z.ai → `glm-5.3-flash`, dots-studio, liquid, …) worked — exactly matching the reported "only certain providers work".

## Changes
- Add `getDefaultModelForTier(provider, tier)` to `ai-providers-config.ts`: free users land on the first free-tier model in the provider's catalog; paid tiers keep the existing default; dynamic-catalog providers (Ollama/LM Studio/Azure) return `''` so no model is sent (unchanged behavior).
- `ProviderModelSelector` uses it for provider switches (single shared component — covers the chat InputFooter and the sidebar chat tab) and now toasts the server's error message instead of a generic one.
- Add 7 unit tests for the new helper.

## Verification
- [x] `ai-providers-config.test.ts` 51/51 passing
- [x] `/api/ai/settings` route tests 31/31 passing
- [x] `bun run --filter 'web' typecheck` clean
- [x] ESLint clean
- Repro: as a free user, open the provider selector and pick OpenAI — previously 403 toast; now switches to `GPT-5.6 Luna` (first free-tier model). Unrelated Postgres integration suites fail in this worktree only because no DB is running.

No DB migrations or config changes.

## Summary by CodeRabbit

- **New Features**
  - Provider selection now automatically chooses the first model available for the selected subscription tier.
  - Free and unspecified tiers no longer default to models that require paid access.
  - Providers with dynamically loaded model catalogs continue to support appropriate model selection.

- **Bug Fixes**
  - Model-selection error notifications now display the server-provided message when available.

- **Documentation**
  - Updated the open-source package inventory to reflect the current dependency list.


2witstudios:
Code review passed with no bugs. Follow-up applied in 89b389c33: providers whose entire catalog is paid (xai, deepseek, qwen, mistral, amazon, etc.) are now hidden from the picker for free users via `hasAccessibleModel`, so the residual 403 path flagged in review is unreachable from the UI. The raw-server-message toast nit was left as-is since it matches the existing model-selector behavior.

PR #2680, opened 2026-09-18
fix(commands): drop the slash icon from transcript command chips

2witstudios:
## What

Removes the decorative `SlashSquare` icon from the transcript command chip — the pill rendered for `/[label](commandId:command)` tokens in user-sent AI chat messages (`CommandChip.tsx`). The `/command-name` text already reads as a command; in user-message prose the extra SVG was the glyph visually floating above the pill. Root-cause removal — no CSS compensation, no behavior change.

Also drops the now-pointless `gap-1` from the chip (it only spaced the icon).

**Scope guard:** the assistant-side `CommandExecutionIndicator` ("Using /foo" / "Skipped /foo") is intentionally untouched — different surface, per the plan in the PageSpace - Dev drive.

## Test plan

- New `apps/web/src/components/messages/__tests__/CommandChip.test.tsx`: resolved chip renders `/release-checklist` as a navigable link to its entry page (`/p/page-1`), and **no `<svg>` inside the chip** — fails if an icon is reintroduced.
- Red-phase verified: re-adding the icon makes the test fail; removing it makes it pass.
- `command-chip-model` + edge-case suites: 26 passing, unchanged.
- Targeted run: 3 files, 27 tests, all green.

## Local gate status (disclosed)

- `bun run --filter web lint`: **exit 0** (one pre-existing warning in `useGlobalDriveSocket.ts`, untouched by this diff).
- `bun run --filter web typecheck`: **fails in this sandbox with 368 pre-existing TS7016/TS7006 errors**, every one `Could not find a declaration file for module 'lucide-react'` in files this PR does not touch. Root cause is the local bun store, not the diff: `lucide-react@0.525.0`'s `package.json` declares `types: dist/lucide-react.d.ts` but the store's `dist/` contains only `cjs/` and `esm/`. Zero errors in the touched files. CI's typecheck is authoritative here.

## Verification steps

1. Open an AI chat and send a message containing a `/command` chip.
2. The user bubble shows a single text-only pill — no slash glyph above/beside it.
3. Click the pill (command with a visible entry page) → routes to the command's page; deleted/disabled/revoked states keep their existing muted/non-navigable behavior.

Before/after screenshots to follow on staging.


2witstudios:
**Gate update at commit `ac4284d50` — all green.**

The PR body's typecheck caveat is resolved: the first local typecheck run failed with 368 TS7016 errors (`missing declaration for lucide-react`) in untouched files — the local bun store had shipped `lucide-react@0.525.0` without its declared `dist/lucide-react.d.ts`. After `bun install --force --frozen-lockfile` repaired the store, `bun run --filter web typecheck` passes (**exit 0**, full web app) with zero code changes, confirming the failures were environmental.

Final status:
- Targeted tests (CommandChip + command-chip-model + edge cases): 27/27 passing
- Lint: exit 0
- Typecheck: exit 0

PR #2683, opened 2026-09-19
fix(auth): sign the desktop app in when the magic-link email opens in a browser

2witstudios:
## Problem

Requesting a magic link from the desktop app never signed the desktop app in. The user got a web session in whatever browser opened the email; the Electron app stayed signed out.

Two independent defects, both verified against the code:

**Break 1 (web side).** The dashboard only fired the `pagespace://auth-exchange?code=...` deep link inside the Electron shell — `extractDesktopExchangeCode` bailed unless `window.electron` existed (apps/web/src/hooks/useDesktopExchangeHandler.ts). But the emailed link is always an `https://.../api/auth/magic-link/verify?token=...` URL (magic-link-adapters.ts builds universal links for iOS/Android only), so it opens in the system browser, where the deep link never fired. A test even pinned the broken behavior ("returns null on web even with desktopExchange param").

**Break 2 (desktop side).** Since the L9 hardening, the main process rejects any `auth-exchange` deep link unless a desktop-initiated auth flow is in progress (apps/desktop/src/main/deep-links.ts, auth-exchange-state.ts). The magic-link flow never began one: `auth:begin-exchange` was only invoked by the dashboard hook that cannot run in a browser, and `auth:open-external` (which does begin a flow) is used only by OAuth/passkey.

The verify route's own comment contradicted the gate: "its handoff has always ridden the emailed GET, so any browser that opens a desktop-bound link..." (verify/route.ts).

## Fix

Follows the passkey pattern — smallest change that restores the flow without weakening L9:

1. **Desktop**: `MagicLinkForm` calls `window.electron.auth.beginExchange()` fire-and-forget after a successful send, when running in the shell. The main process records the flow; the state is not echoed (the emailed link opens outside the shell, so it could not carry it back anyway).
2. **Web**: `useDesktopExchangeHandler` drops the `isDesktopPlatform()` gate — any browser with a `desktopExchange` param fires `pagespace://auth-exchange?code=...`. The in-shell `beginExchange()` state-echo path is preserved unchanged.

Acceptance is still bound to a desktop-initiated flow: the 10-minute flow TTL comfortably covers the 5-minute magic-token and exchange-code TTLs, and a stateless deep link is accepted only while a flow is in progress — the same `flow-in-progress-no-state` branch the passkey flow already relies on.

## Testing

- `useDesktopExchangeHandler.test.ts`: the web case now asserts the code IS returned in a plain browser (was pinned to null); desktop and missing-param cases unchanged; the now-unused `isDesktopPlatform` mock removed.
- `MagicLinkForm.test.tsx` adds a "desktop exchange flow" suite: flow begun after a successful send, NOT begun when the send fails (no stray acceptance window), and no-op on web.

- [x] `bun run --filter 'web' test` (both suites, 16 tests)
- [x] `bun run typecheck`
- [x] `bun run --filter 'web' lint`

## Known tradeoff

If the user takes longer than the 10-minute flow TTL to click the email link, the desktop app shows the L9 "Unexpected sign-in request was blocked" error instead of silently ignoring the link — the same window the passkey flow has. Re-requesting the link starts a fresh flow.

## Follow-up (not in this PR)

Coder shell threading: an external browser cannot detect the PageSpace Coder variant, so `buildDesktopExchangeDeepLink` defaults to `pagespace://`. Coder users clicking the emailed link in a browser may land in the wrong app (or nowhere). Threading the shell through send → token metadata → redirect is a separate change, as is the shared cross-device wart (request from desktop, click on phone → custom scheme silently no-ops, phone still gets a web session) that the OAuth bridge already accepts.

## Summary by CodeRabbit

- **New Features**
  - Magic-link sign-in now automatically begins the desktop sign-in handoff after an email is successfully sent.
  - The emailed link can be accepted directly within the desktop app.
  - Web users continue to receive the standard confirmation without triggering the desktop flow.

- **Bug Fixes**
  - Improved handling of magic-link exchange parameters across browser and desktop sign-in experiences.
  - Failed magic-link requests no longer start the desktop sign-in handoff.


2witstudios:
Follow-up fix pushed (359565d3): `auth:begin-exchange` can reject (IPC failure) or resolve `null` (untrusted sender — the main process records no flow). The form previously invoked it fire-and-forget after setting the sent state, so both outcomes were discarded and the later deep link would hit the L9 gate with no pending flow.

Now the form awaits the call before confirming: rejection or null returns to the input view with a retryable desktop error (submit re-enabled, no cooldown), and the sent confirmation appears only once the flow is recorded. A deferred-promise test pins the ordering; the rejection/null cases assert the error path.

PR #2681, opened 2026-09-18
feat(web): splittable dashboard + one floating-input chat surface

2witstudios:
## What & why

Every chat surface except the dashboard rendered out of a workspace node tree — which is why every surface EXCEPT the dashboard could be split. And the composer was duplicated: the dashboard had the animated floating input, panes a hardcoded `border-t` footer. This PR fixes both, moving toward one chat surface: the chat area stays surface-owned; the input and header become shared, context-dependent chrome.

Design demos were built as canvas pages in the `PageSpace - Dev` drive ("UI Demo · Floating input in panes") and approved before implementation.

## Changes

**One floating-input surface everywhere**
- `useInputPosition` (latched centered↔docked resolver + `isBusy` so the first send docks immediately) and `FloatingInputLayer` (welcome overlay + glass `InputCard`) extracted from `ChatLayout`.
- `InputPositioner` centers container-relatively (`cqh`, not `-50vh`) — correct in a resizable pane, not just at viewport height.
- `ChatLayout` consumes the layer (dashboard unchanged). `SessionChatView` swaps its hardcoded footer for it: new conversations get the centered welcome (agent/assistant name) in agent panes, global-assistant panes, agents-console panes, and the session-less agent page; the page-context dock is the glass card capped to the messages column.

**One header, two densities**
- `ConversationHeader`: the PaneBar slot contract (identity / actions / paneControls) at `compact` (== PaneBar) and `comfortable` (dashboard h-11) densities. `GlobalAssistantView`'s bespoke header now renders through it.

**The dashboard as a splittable pane workspace**
- `agent_workspaces.kind` (`'agent' | 'dashboard'`): the `dashboard` row's tree IS the user's dashboard layout — persisted in the existing `agent_workspace_nodes` model, no new layout storage.
- Lazy get-or-create provisioning seeded with one chat pane bound to the active global conversation (ownership + write-once-binding checked server-side; unbindable ⇒ picker pane; insert-race loser re-reads the winner). `GET /api/agent-workspaces/dashboard` returns the same snapshot shape the grid already hydrates from.
- `CenterPanel` mounts `DashboardWorkspaceView` (the grid) in place of the overlay; `GlobalAssistantView` remains only as the provisioning fallback — the dashboard never blanks.
- Dashboards are excluded from session listings and the spawn ceiling (visibility bounds, not semantics).

**Identity moves with the tree, both directions**
[...]

PR #2687, opened 2026-09-21
fix: pending invitee roster/DM leak, accessible_page_ids denies + UTC expiry, page-viewers CI timeout

2witstudios:
Two P2 security fixes and one CI flake on master. All three were found by the independent review of #2672 (review 5238087524).

## Finding 1 (P2, privacy): a pending invitee could see the drive's roster and DM its members

**What was wrong.** `GET /api/users/messageable` read the caller's drives from `drive_members` with no `acceptedAt` gate. It read their co-members the same way. For each one it returned **name, email, bio and avatar**. `usersShareDrive` had the same gap, so a pending invitee could also open a DM with any member, and any member could open one with the invitee. Comments in both places said this was intentional.

**Which rule is right: the one in `visibility.ts`.** The evidence:
- `apps/web/src/lib/users/visibility.ts` handles the same identity data for `/api/users/find` and name search. It gates on `acceptedAt` and says why: "a pending, unaccepted invitation is not an established shared context, so it must not let the invitee resolve other members' identities (nor vice versa)". Messageable returned more than that endpoint does, because it includes email.
- The repo convention is locked in by `drive-member-gate-coverage.test.ts`: every `drive_members` read used for an authorization decision is gated. These two files were exempt only because of that "softer DM eligibility" comment.
- The comment gave two reasons: rows whose "acceptance state is in flux", and legacy rows that predate the `migrate-pending-invites` cleanup. That script deletes exactly those legacy `acceptedAt IS NULL` rows. Pending rows created today are real, unaccepted invitations (the page-invite flow sets `acceptedAt` on them when the invite is accepted). So the "soft" exemption now covers only people who have not joined.

**Fix.** Both `drive_members` reads in the route and both in `usersShareDrive` now carry `isNotNull(driveMembers.acceptedAt)`. The two gate-coverage exemptions are removed, so the coverage test now requires these files to have zero ungated reads.

**What changes for a pending invitee** (until they accept):
- Nobody from that drive appears in their New Conversation picker: not the owner, not the members, not other pending invitees. They also stop appearing in the members' pickers.
- `POST /api/messages/conversations` between them and anyone in the drive returns 403 ("You can't message this user"), unless the two already have an ACCEPTED connection. Connections are unaffected.
- Once they accept, everything works as before.
- Not changed here: a DM conversation opened *before* this fix stays usable, because the send path does not re-check `usersShareDrive`. No new ones can be opened.

**Tests (written first, red before the fix):**
- `packages/lib/.../users-share-drive.integration.test.ts` (real Postgres): a pending invitee cannot DM the owner or a member in either direction, and two pending invitees cannot DM each other. 2 of 3 cases were red before the fix. The third case is a control: accepted co-members and the owner can still DM each other.
- `apps/web/.../messageable/__tests__/route.test.ts`: the operator mocks now build a predicate tree, and the test asserts that both `drive_members` reads carry the `acceptedAt` gate. Red before the fix.
- `users-share-drive.test.ts`: the old "returns true even when membership rows lack acceptedAt" case is replaced by a check that both reads are gated.

## Finding 2 (P2, access): `accessible_page_ids_for_user` ignored denies and compared expiry in local time

**What was wrong.** Migration 0133 added rule 4 ("any accepted member reads a non-private page") as an unconditional `OR`. That had three consequences:
- A custom role's per-page `canView:false`, or its drive-wide `canView:false`, still left the page in the set.
- An unexpired `page_permissions` row with `canView=false` still left the page in the set.
- The function ignored custom roles entirely, so a custom role's per-page grant on a **private** page was *missing* from the set, even though `getUserAccessLevel` allows it.

It also compared `expiresAt` (a `timestamp without time zone` holding UTC wall-clock time) against bare `now()`. Postgres interprets that comparison in the session's time zone. West of UTC, an expired grant stayed live for hours; east of UTC, a live grant was already dead. The existing test "excludes pages whose explicit grant has expired" **fails on master under a non-UTC session**. CI only passed because it runs in UTC.

**Callers.** The only live callers are the pulse routes (`/api/pulse`, `/api/pulse/generate`, `/api/pulse/cron`). `page-payload-service` and `app-shell-service` also use the function, but `loadAppShell` and `loadPagePayload` have no production caller today.

**Fix: migration `0296_accessible_page_ids_denies_utc.sql`.** Drizzle cannot express a SQL function, so this is a **custom migration created by `drizzle-kit generate --custom`**, the supported way. It comes with its journal entry and a snapshot. Apart from key order, the snapshot is identical to 0295, because no schema changed. It follows the pattern of 0102 and 0133: DROP-then-CREATE, `SECURITY DEFINER`, pinned `search_path`. No applied migration was edited. The function now resolves each page in `getUserAccessLevel`'s order, and the first matching rule wins:
1. Drive owner: visible.
2. Accepted ADMIN: visible.
3. An unexpired explicit permission decides the page on its own `canView`. A deny is final.
4. Not an accepted member: hidden.
5. The member's custom role (same drive only) has an entry for this page: that entry's `canView` decides.
6. The custom role has a drive-wide default: hidden if the page is private, otherwise the default's `canView` decides.
7. Otherwise: visible if and only if the page is not private.

Expiry is compared against `(now() at time zone 'utc')`. Trashed pages and trashed drives are still excluded up front, as before.

**Tests (written first):**
- `packages/db/.../accessible-page-ids.integration.test.ts` adds 11 cases:
  - custom-role per-page deny
  - drive-wide deny (the per-page entry still wins)
  - a drive-wide grant does not open a private page, but a per-page grant does
  - a role from another drive is ignored
  - explicit deny for a member
  - an explicit grant overrides a role deny
  - an expired explicit deny falls back to rule 4
  - a pending member gets nothing from a role or from rule 4
  - expiry in `America/Chicago`, `Pacific/Kiritimati` and `UTC`. The zone is pinned per transaction with `set_config('TimeZone', …, true)`, so the result does not depend on the runner's zone.

  Before the fix: 7 failed, 18 passed. After: 25/25.
- `packages/lib/.../accessible-page-ids-agreement.integration.test.ts` checks the invariant directly. It seeds one drive and 15 kinds of user: owner, stranger, member, admin, pending member, pending admin, each role shape, explicit deny, expired deny, role deny overridden by a grant, non-member with live and expired grants, and a pending member with a role and a grant. For every page, the SQL set must equal `getUserAccessLevel(...).canView`. It fails against 0133's function body and passes against 0296.

## Finding 3 (CI flake): page-viewers "evaluates every candidate across the chunk boundary"

**Evidence.** The test seeded 205 users and 205 memberships one row at a time, which is 410 round trips to the Postgres that CI shares between suites. When it passes in CI, it takes 1.1–4.8s. It hit the 5000ms limit in runs 35372114954 (master), 35212495266 and 35191234025.

**Fix.** Added `factories.createUsers(count)` and `factories.createDriveMembers(driveId, userIds)`. Each inserts 500 rows per INSERT, far below the 65535 bind-parameter limit, and builds its rows with the same helpers as `createUser`/`createDriveMember`. The test now seeds with 2 statements instead of 410. The assertions are unchanged: 205 candidates, all visible, and `members[200]` and `members[204]` past the 200-row chunk boundary checked by name. Raising the timeout was not needed. A new `factories-bulk.integration.test.ts` seeds 501 rows, which crosses the factory's own 500-row chunk boundary, so a chunking bug cannot silently seed fewer rows.

## Mutations (by line index, with a no-op control that stayed green on all five suites)

| Mutant | Result |
|---|---|
| route: caller's own `drive_members` read loses its gate | RED (web 1 failed) |
| route: co-member read loses its gate line | RED (web 1 failed) |
| usersShareDrive: A's membership read ungated | RED (share 2 failed) |
| usersShareDrive: B's membership gate line deleted | RED (share 2 failed) |
| SQL: `(now() at time zone 'utc')` → `now()` | RED (sql 4, agree 1) |
| SQL: explicit row → `THEN true` | RED (sql 2, agree 1) |
| SQL: explicit-row branch deleted | RED (sql 8, agree 1) |
| SQL: `acceptedAt IS NOT NULL` on the member join deleted | RED (sql 2, agree 1) |
| SQL: non-member → rule 7 | RED (sql 24, agree 1) |
| SQL: per-page role → rule 7 | RED (sql 2, agree 1) |
| SQL: drive-wide opens private (`THEN true`) | RED (sql 1, agree 1) |
| SQL: drive-wide `canView` ignored | RED (sql 1, agree 1) |
| SQL: role's `driveId` match deleted | RED (sql 1) |
| factories: loop stops after the first chunk | RED (bulk 1) |

Two further mutants turned out to be syntax errors (a dropped comma before another argument), so they proved nothing. They were replaced by the valid "gate line deleted" mutants above.

## Verification

- Every integration suite above was run against a freshly migrated Postgres 17 with session time zone `America/Chicago`, which applies 0296 in order. `bunx eslint` on the touched apps/web files is clean.
- **Typecheck, knip and the full suites are left to CI.** Per the machine-wide no-local-builds rule, they were not run locally.
- **Migration number:** #2681 and #2590 also add a `0296_*` migration. Whichever of the three merges later needs to regenerate its migration number.

Found by the independent review of #2672 during the Organizations & Wallets epic.

https://claude.ai/code/session_01DjU7XD9bkWwXgtpAdqVoPk


2witstudios:
@codex review

2witstudios (commented):
[independent-review] agent:ow-irv-2687

**Verdict: no blocking findings on `cfdc9d1d6`. All three fixes do what the PR says, and each one has a test that fails when the fix is removed.** CI: all 11 checks green on this head — Test Suite run 35557950530 (Unit Tests, Lint & TypeScript Check, E2E) and Security Tests run 35557950356 (**Security Test Suite**, Static Security Analysis, Dependency Audit, Secret Scanning, CodeQL).

Base check: the merge-base is master tip `2d975c9e5`, so the base is current.

## How I verified
I used a throwaway homebrew **postgresql@17** with the server TimeZone set to **America/Chicago** (non-UTC), a short `-k` socket dir and port 5487. It was stopped and deleted afterwards. I ran all migrations on a fresh DB: **297 applied, 0296 included**, and `prosrc` holds the `at time zone 'utc'` body.

## (a) Pending invitee: roster and DM
- `GET /api/users/messageable` returns each co-member's name, email, bio and avatar. **Before:** a pending invitee (NULL `acceptedAt`) got the owner and every member of the drive they were invited to. The members also got the invitee back. **After:** both `drive_members` reads are gated (`route.ts:60`, `:83`). A pending invitee's drive set now holds only drives they own, and they never show up as a co-member. Connections are unaffected. Accepted members see the same list as before (owner plus accepted co-members).
- `usersShareDrive` gates both reads (`permissions.ts:936`, `:956`). The integration test covers pending↔member, pending↔owner (both directions) and pending↔pending, all false; accepted↔accepted and member↔owner are true. The only caller is `POST /api/messages/conversations`, which falls back to it after the connections check.
- **Sweep for other ungated identity exposure. All of these are gated:**
  - `/api/users/find` (`callerCanViewUser`)
  - `/api/users/search`, related branch (`searchRelatedProfilesByName`)
  - `/api/mentions/search` (`getDriveRecipientUserIds`, gated)
  - `/api/drives/[driveId]/members`: filters `acceptedAt !== null`; pending invites are shown to owner/admin only
  - AI `list_drive_members`: skips rows with no `acceptedAt`
  - channel rosters and broadcasts (`getUsersWhoCanViewPage`, gated)
  - `apps/realtime`: never reads `drive_members` directly
  - The remaining entries in `drive-member-gate-coverage`'s EXEMPT list are writers, candidate filters or owner/admin-gated displays. The PR removes the two exemptions it fixes, so the count-based guard now catches any regression.

## (b) `accessible_page_ids_for_user` against `getUserAccessLevel`
I compared the CASE order line by line with the TS resolver (owner → accepted ADMIN → unexpired `page_permissions` decides, deny is final → no accepted membership hides the page → custom-role per-page entry → drive-wide default, which never opens a private page → rule 4). It matches, including these edge cases:
- a per-page entry of JSON `null` falls through to rule 4
- an entry of `{}` denies
- a drive-wide default of JSON `null` falls through; SQL NULL falls through
- a role from a foreign drive is ignored
- a drive_members row with role `OWNER` for a user who is not the drive owner is treated as a plain member in both

**My own randomized fixture** (not in the PR): one drive with 8 pages (private ones and a CHANNEL) and 60 users. Membership is drawn from none, pending, member, admin, `OWNER`-role and pending-admin. Custom roles are random: per-page entries are absent, ON, OFF, `null` or `{}`; drive-wide is SQL-null, ON, OFF or `{}`; 10% of roles belong to a foreign drive. Explicit rows are random canView true/false, with expiry null or ±0.5h/±3h/±10h. I compared the function's output with `getUserAccessLevel(...).canView` for every user×page, with the session TimeZone set to America/Chicago, Asia/Tokyo and UTC. **6 seeds × 1,464 cells = 8,784 cells, 0 mismatches.** The same fixture run against the old 0133 body mismatches, so the fixture can detect a regression.
- **Expiry in a non-UTC session:** the PR's `it.each(['America/Chicago','Pacific/Kiritimati','UTC'])` passes. The old bare `now()` fails it (see mutations).
- **No narrowing for ordinary accepted members:** I compared 20 plain members against the verbatim 0133 body, installed side by side. In Chicago and in UTC the output is identical: all non-private pages, no private pages, no trashed pages. Rule 4 still applies.
- **Migration:** `0296` has `--custom` provenance. `0296_snapshot.json` is identical to `0295_snapshot.json` apart from the `id`/`prevId` chain (`prevId` = 0295's `id`). The journal `when` (1789960980438) is after 0295's. The diff only adds files and appends to the journal, so no applied migration is edited. DROP+CREATE loses no GRANT/REVOKE (0102/0133 never issued any), and the 0133 index is kept. It applies on a fresh DB. FYI, not this PR's problem: `pu/org-wallets` also holds a 0296 and will renumber at its next master sync.
- **Performance:** 200k pages across 2k drives, a user with 399 memberships and 3k explicit rows: new ≈ 75–88 ms, old ≈ 56–87 ms. Both do a full `pages` scan, so this is not a regression. The new body returns 251 fewer ids for that user; those are exactly the explicit denies it now honours.

## (c) page-viewers chunk-boundary test
- The test asserts the same things as before: 205 candidates, `viewers.size === members.length`, and `members[200]` and `members[204]` present. Only the seeding changed.
- **Timing:** PR 49–52 ms against master 202–288 ms for this test, 3 runs each on the same DB, so about 5× faster. CI's 5000 ms timeout came from 410 serial round trips on a shared instance.
- **Bind limit:** there are 500 rows per INSERT. `users` has 30 columns (worst case 15,000 params) and `drive_members` has 9 (4,500), both far under 65,535. `factories-bulk.integration.test.ts` inserts 501 rows, which crosses the chunk boundary.

## (d) Mutations
Each mutation was applied by line index with an asserted anchor, then reverted. For the SQL mutations I applied the mutated body with psql and re-applied the original afterwards. When a mutation touched `@pagespace/db` factories I rebuilt the db dist, because lib consumes it from dist.

| Mutation | Result |
|---|---|
| messageable `route.ts:60` gate removed | route test **red** (1/6) |
| messageable `route.ts:83` gate removed | route test **red** (1/6) |
| `permissions.ts:936` gate removed | unit **red** (1/8), integration **red** (1/3) |
| `permissions.ts:956` gate removed | unit **red** (1/8), integration **red** (1/3) |
| 0296 L55 `(now() at time zone 'utc')` → `now()` | db suite **red** (4/25), agreement **red**, my random fixture **red** |
| 0296 L63 explicit deny ignored | db **red** (1/25), agreement **red**, random **red** |
| 0296 L67 per-page role branch disabled | db **red** (2/25), agreement **red**, random **red** |
| 0296 L70 drive-wide branch disabled | db **red** (1/25), agreement **red**, random **red** |
| factories L20 chunk slice short by one | factories-bulk **red** |
| factories L20 only first chunk inserted | factories-bulk **red** |
| `createDriveMembers` drops tail past 200 | page-viewers **red** |
| `getUsersWhoCanViewPage` L1183 only first 200-batch | page-viewers **red** |

After restoring everything, all suites were green: db 25/25 + factories-bulk 1/1, lib agreement 1/1, users-share-drive 3/3 integration and 8/8 unit, page-viewers 9/9, messageable route 6/6. `git status` was clean.

## Non-blocking notes (follow-ups, if wanted)
1. **DMs a pending invitee already opened stay open.** A conversation opened under the old rule persists, and the `[conversationId]` routes do not re-check eligibility. `GET /api/messages/conversations` still returns the other party's name and email. This is pre-existing DM-persistence design, but it is the one place the old leak outlives the fix.
2. **The app shell still lists pending invitees to every accepted member.** `app-shell-service.fetchDriveMembers` includes pending rows (userId, role, invitedAt, acceptedAt=null) for every accepted member of the drive. It sends only an opaque userId, no name or email, and it is exempted with a reason. It is the reverse direction ("nor they the invitee's") at id level only.
3. **The messageable route only has a mocked test.** Its coverage checks the shape of the predicate tree. The DB round trip is covered for `usersShareDrive`, not for the route itself.
4. **Malformed `canView` values diverge, in the safe direction.** When a custom-role `canView` is a non-boolean truthy value (for example `"true"`), TS allows access (`resolved.canView ?`) and SQL denies it (`= 'true'::jsonb`). The schema types it boolean, so this only affects malformed rows.

No threads opened; nothing resolved; no code changed.

2witstudios:
Thanks, ow-irv-2687. On the four non-blocking notes: no code change in this PR. Each one is outside the three findings or is a design choice for a separate decision.

1. **DMs a pending invitee already opened stay open.** Agreed. This PR stops any new conversation from being opened, and the PR body says so under "Not changed here". Closing existing conversations means deciding what happens to them: hide them, make them read-only, or re-check eligibility on every send in the `[conversationId]` routes. That is a DM-persistence decision, so it is left to the orchestrator as a follow-up rather than folded into this fix.
2. **The app shell lists pending invitees at id level.** `fetchDriveMembers` returns only an opaque userId, and the `drive-member-gate-coverage` exemption says why. Also, `loadAppShell` has no caller in the app today. That makes it a follow-up with no live exposure, and it is outside the roster/DM surface of finding 1.
3. **The messageable route only has a mocked test.** Acknowledged. The route test pins both gated reads, and the mutations at L60 and L83 turn it red. The real database round trip is covered for `usersShareDrive`, the function that actually allows or refuses the DM. A DB-backed route test would add the same assertion one layer up. It is worth adding if the route grows its own logic.
4. **Malformed `canView` values diverge.** This is intended: SQL denies a non-boolean value where TS would allow it, so the divergence fails closed. `drive_roles.permissions` is typed boolean, so only a malformed row can reach it.

CI is green on `cfdc9d1d6` (11/11), Codex found no major issues, and the branch is level with master.

PR #2695, opened 2026-09-21
fix(ai): bill the in-flight step when a chat turn is stopped or hits the credit ceiling

2witstudios:
## What was wrong (P1, live on master)

An explicit **Stop** or a **credit-ceiling abort** in the main chat pipeline (`handleChatTurn` → `runPageChatTurn` / `runGlobalChatTurn` → `runAgentWithRetry` → settle) billed the step that was still streaming at **$0**. Probed against the real `ai@6.0.212` `streamText` (MockLanguageModelV3 under an AbortSignal that errors the provider stream like fetch does):

| Scenario | `steps` | `totalUsage` | Billed before |
|---|---|---|---|
| Stop during the only step | **rejects** (AbortError), onAbort gets `steps: []` | **rejects** | $0 |
| Stop during step 2, step 1 finished | resolves `[step1]` | resolves `{inputTokenDetails:{}, outputTokenDetails:{}}` with **no token counts** | OpenRouter: step 1 only. **Non-OpenRouter: $0, including the finished step 1** |
| Credit ceiling fires from `onStepFinish` | resolves `[step1]` | same empty object | same as above |

The second row is broader than the reported bug: on a non-OpenRouter provider the finished steps were dropped too, not only the interrupted one.

## Fix

- **`runAgentWithRetry`** watches the chunks it pipes (new `onChunk` hook on `pipeUIMessageStreamStrippingStart`). If the run is aborted while a step is open (`start-step` seen, no `finish-step`), it returns `abortedStep`: the prompt text of the attempt, the prior finished step's reported input+output, and the output streamed for the interrupted step (text, reasoning and tool input). When aborted with a token-less `totalUsage`, it sums the finished steps' own `usage` instead.
- The interrupted step's prompt is read from the messages **actually sent**: each route defines its turn-context rewrite once (`toSentMessages`) and both `buildStreamText` and the estimate use it (Codex P2, 33f95886d).
- An abort that lands after ai@6 recorded a step (usage in `steps`) but before its `finish-step` chunk reached the stream is **not** estimated again: the estimate runs only when `steps.length <= finish-step chunks seen` (966d2fbb1, no double bill).
- **`agentRunBillingFields`** (new, `apps/web/src/lib/ai/core/agent-run-billing.ts`) is the single place that turns a run into `trackUsage` fields. With no interrupted step it returns exactly what the three call sites computed before. With one, it adds the interrupted step (see policy).
- Wired into all three settle sites: page settle (onFinish), page error path, global settle. Each still makes **one** `trackUsage` call with the `holdId`, so the hold is settled and released once.

## Billing policy for the interrupted step: Jono to confirm or change

1. **Input**: the prior finished step's provider-reported input+output tokens, which the interrupted step re-sends, so this is a floor. If no step finished, the estimated tokens of the messages plus the system prompt and tool definitions.
2. **Output**: tokens actually streamed before the abort. Nothing beyond what reached the client.
3. **Counting**: the platform's `estimateTokens` (4 chars/token, `@pagespace/lib/monitoring/ai-monitoring`). Note: the credit reservation itself has no text estimator; it prices a fixed assumed budget (`CHAT_HOLD_ASSUMED_INPUT/OUTPUT_TOKENS`) through `estimateChatHoldCentsForModel`. The interrupted step is priced through the same `calculateCost` at the model's catalog rate, and the existing markup applies at settle as usual.
4. **Cap**: the interrupted step never charges more than the chat hold for the model (`estimateChatHoldCentsForModel`, converted back through `MARKUP_BPS`). `metadata.abortedStep.capped` records when the cap applied.
5. **Provider numbers win**: finished steps bill on OpenRouter's returned cost, or on summed provider usage. Only the step with no provider numbers is estimated. ai@6 never delivers usage for an aborted step, so the estimate replaces a $0, not a real figure.
6. **Cost reconcile**: when a run carries an estimated step, the row is **not** queued for OpenRouter cost reconcile. The reconcile fetches only the finished steps' generations, and would treat the estimate as an overcharge and refund it. The finished steps' generation ids are kept in `metadata.abortedStep.generationIds`, and the row is labelled `costSource: 'estimate'`. Trade-off: those rows lose the small drift correction on their finished steps.
7. Unknown or uncatalogued models price the interrupted step at $0, the same as the rest of the estimate path.

### Alternatives (what an aborted step could cost instead)

| Option | Charge for the interrupted step | For | Against |
|---|---|---|---|
| **A. Recommended (this PR)** | estimated input + output streamed so far, at catalog rate, capped at the chat hold | closest to what the provider actually billed us, without exceeding what the user was shown as reserved | estimates, not provider figures; the reconcile is skipped for these rows |
| B. Charge the full hold | the chat hold amount, flat | simple; never under-recovers on small turns | over-charges a Stop that lands early; a user pressing Stop pays the most |
| C. Output only | output streamed so far, no input | cheapest for the user; nothing charged before any output | under-recovers: input is usually the larger cost (every step re-sends the whole context) |
| D. Input only | estimated prompt, no output | charges for the provider reading the context | ignores output already delivered to the user |
| E. $0 (current master) | nothing | none | provider bills us for every stopped turn; stop-spam is free inference |

**What option A does in practice (independent review, ow-irv-2695):**
- The cap is the price of the 4000-in / 1000-out hold budget, so it binds whenever the interrupted step's estimated prompt exceeds roughly 9k tokens (on a $3/$15 model), even with no output streamed. System prompt plus tool definitions alone usually exceed that.
- So an early Stop usually charges the **full hold**: up to `CREDIT_HOLD_ESTIMATE_CENTS` (25¢ by default), and much less on cheap models. In practice option A mostly behaves like option B.
- At that size, real uncached input cost is at least the hold. Charging more than the provider billed is possible only through cached or reasoning prompt tokens, and is bounded by the hold. Follow-up #2702 removes both sources (cached share at the cache-read rate, reasoning excluded from the floor).
- `metadata.abortedStep.capped` records how often the cap binds.

Separately from the options: **the fix for finished steps being dropped** (non-OpenRouter multi-step aborts billed $0 even for steps with full provider usage) is not a policy choice; those steps have provider figures and should bill under any option.

Unchanged: normal completion and the in-stream error chunk path (both still read `totalUsage`/`steps`; there is a real-stream test for each). A client disconnect still never aborts the model.

## Tests

- `run-agent-with-retry-abort.test.ts`: **real `streamText`** with provider streams that error with AbortError under the signal, as fetch does. Covers:
  - a single-step Stop captures the prompt and exactly the streamed output;
  - a multi-step Stop keeps step 1's 800/30 tokens and floors step 2's prompt at 830;
  - a credit-ceiling abort mid-step through `AbortSignal.any` behaves like a Stop;
  - a credit-ceiling abort from `onStepFinish` bills the finished step;
  - normal completion and an error chunk carry no `abortedStep` and keep their `totalUsage`.
- Same file: the prompt estimate includes the injected turn context; an abort racing a step's finish is billed once from provider usage (no estimate on top); a credit abort at a step boundary estimates no phantom step.
- `agent-run-billing.test.ts`: pass-through is identical to the old fields; single-step Stop bills 1000 in / 50 out = $0.0025 instead of $0; non-OpenRouter multi-step = $0.00406; OpenRouter $0.01 + interrupted step, with reconcile ids withheld; the cap holds.
- `stream-socket-events.test.ts` (page route): a Stop during the only step, gated with a hold, calls `trackUsage` **exactly once** with `holdId: 'hold-1'`, 50 output tokens, input > 0, cost > 0 and ≤ the hold.
- **Mutation by line index** (control GREEN, 88 tests): M1 drop the in-flight capture → RED (4); M2 keep the empty `totalUsage` → RED (3); M3 restore the old $0 billing → RED (5); M4 remove the cap → RED (1); M5 keep the reconcile ids → RED (1); M6 drop `onChunk` → RED (3).
- Unchanged suites run locally: page `stream-socket-events` 66/66, global `stream-socket-events` 34/34, global `credit-gate` 10/10, page `credit-gate` 4/4, `credit-abort` 5/5, `handle-chat-turn` 13/13, `run-agent-with-retry` 11/11.

Local runs used a throwaway source-alias vitest config: the worktree has no lib/db dist and no local builds are allowed. **Typecheck and knip are left to CI.** `bunx eslint` on every touched apps/web file: clean.

Out of scope: `/btw` (#2694), org-wallets code.

https://claude.ai/code/session_012kejh8fgUkwjyGKxNEtJ9G

## Summary by CodeRabbit

* **Bug Fixes**
  * Stopped assistant replies are now billed more accurately.
  * Completed steps and the portion of an interrupted reply that was already displayed are included in billing.
  * Unfinished output is never charged beyond the amount reserved when the reply began.
  * Billing now remains accurate when replies run out of credits during multi-step responses.


2witstudios:
@codex review

apps/web/src/lib/ai/core/run-agent-with-retry.ts
```diff
@@ -210,8 +288,29 @@ export async function runAgentWithRetry(
...
+    const attemptUsage = abortSignal.aborted && !hasTokenCounts(usage) ? stepsUsage() : usage;
+
     accumulatedSteps.push(...steps);
-    accumulatedUsage = mergeUsage(accumulatedUsage, usage);
+    accumulatedUsage = mergeUsage(accumulatedUsage, attemptUsage);
+
+    if (abortSignal.aborted && stepOpen) {
+      const priorStepUsage = (steps.at(-1) as { usage?: LanguageModelUsage } | undefined)?.usage;
+      abortedStep = {
+        promptText: messagesText(baseMessages),
```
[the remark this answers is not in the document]
2witstudios: Fixed in 33f95886d. Each route now defines the turn-context rewrite once (`toSentMessages`: `appendTurnContextToLastUserMessage` + `buildVolatileTurnContext`). `buildStreamText` uses it to build the request, and `runAgentWithRetry` uses it for the aborted-step prompt estimate (`promptText: messagesText(toSentMessages(baseMessages))`), so the estimate reads the exact messages the provider was sent. Test: `run-agent-with-retry-abort.test.ts` › "prices the prompt the provider was SENT, including the injected turn context" (real ai@6 stream, stopped mid-step). Mutation check: reverting the estimate to `messagesText(baseMessages)` turns it red (1 failed | 7 passed). Leaving this open for you to verify.

apps/web/src/lib/ai/chat-pipeline/global-chat-turn.ts
```diff
@@ -1498,22 +1499,21 @@ export async function runGlobalChatTurn(ctx: GlobalChatTurnContext): Promise<Res
         // so the orphan-sweep can recover/bill it. Missing tokens mean a $0 cost
         // row, not a skipped one (which would silently front the spend).
         try {
-          const usage = agentRun?.accumulatedUsage;
-          const steps = agentRun?.accumulatedSteps;
           const duration = Date.now() - startTime;
+          // Provider usage + cost, plus the step an abort interrupted (which the
+          // provider charged for but reports no usage on) — see agentRunBillingFields.
+          const { abortedStep, ...billing } = agentRunBillingFields({
```
2witstudios: [independent-review] No test covers this wiring. The mutant `agentRun: agentRun && { ...agentRun, abortedStep: undefined }` on the next line keeps global `stream-socket-events` + `credit-gate` green (44/44). The page route has the equivalent test (the M4/M1b mutants go red there). Worth one mirror test: a Stop during the only step settles once with `holdId`, `outputTokens > 0` and a cost ≤ the hold.

apps/web/src/lib/ai/core/agent-run-billing.ts
```diff
@@ -0,0 +1,114 @@
...
+      openrouterGenerationIds,
+    };
+  }
+
+  const stepInputTokens =
+    aborted.priorStepContextTokens ?? estimateTokens(aborted.promptText) + promptOverheadTokens();
+  const stepOutputTokens = estimateTokens(aborted.outputText);
+  const estimatedDollars = calculateCost(model, stepInputTokens, stepOutputTokens);
+  const capDollars = estimateChatHoldCentsForModel(model) / 100 / (MARKUP_BPS / 10000);
+  const stepDollars = Math.min(estimatedDollars, capDollars);
```
2witstudios: [independent-review] Policy note, not a defect. `capDollars` is the price of the 4000-in/1000-out hold budget, so the cap binds whenever the estimated prompt of the interrupted step is larger than ~9k tokens (for a $3/$15 model with no output yet). System prompt + tool definitions alone probably exceed that, so in practice an early Stop charges **exactly the hold**, i.e. option A behaves like option B from the PR body. Separately, line 85 prices `priorStepContextTokens` at the full fresh-input rate although this route caches the prefix (cache reads at 10%), and that figure includes the prior step's reasoning output. Those are the two places this can exceed what the provider actually billed for the step, bounded by this cap.

2witstudios (commented):
[independent-review] agent:ow-irv-2695

**Verdict: no blocking defects found at `300a893bf`. The fix is correct for the reported bug: Stop and credit-ceiling aborts no longer settle at $0, and finished steps are no longer dropped.** On price, the charge can exceed what the provider billed us (see (e)), but only on the step the Stop interrupted, and never by more than the chat hold. Code unchanged by me; nothing resolved.

**Blocking: CI red on a ratchet the PR trips (see below). The billing logic itself has no blocking defect.**

**CI on `300a893bf` is RED, from one real failure this PR causes** (run 35663909088). Lint & TypeScript pass. Unit Tests fail on 1 of 21,007: `apps/web/src/lib/ai/chat-pipeline/__tests__/turn-duplication-ratchet.test.ts`: "The measurement is 158 but this file records 161". Extracting `agentRunBillingFields` REDUCED the page/global duplication, and the ratchet demands an exact match. Fix: lower `RECORDED_IDENTICAL_LINES` to 158 and update the figure in `handle-chat-turn.ts` and `docs/2.0-architecture/agent-sessions.md`. The same failure repeats on `33f95886d` (run 35665113689, 1 of 21,009). E2E failed here on `18-sidebar-directory-live.spec.ts:451` (agent-session sidebar row count, retried once). That spec does not touch the billing code, and E2E **passed** on `33f95886d` and on master `12ef8f23b`: treat it as a flake.

### (a) Stream shapes: real `ai@6.0.212` probes (my own, not the PR's tests)
I ran throwaway probes through `runAgentWithRetry` → `agentRunBillingFields` on a real `streamText` + `MockLanguageModelV3`, with a provider stream that errors with AbortError under the signal, the way fetch does:

| Scenario | steps / usage | abortedStep | Billed |
|---|---|---|---|
| Stop mid-text, only step (PR test) | 0 / rejected | prompt + exact streamed text | estimate > 0 ✅ |
| Stop mid step 2, step 1 = 800/30 (PR test) | 1 / 800+30 from step usage | floor 830 + streamed | step 1 + estimate ✅ |
| Credit ceiling via `AbortSignal.any` mid-step (PR test) | same as above | same as above | ✅ |
| Ceiling from `onStepFinish` (step boundary) | 1 / 800+30 | none | step 1 only, via the token fallback (no `providerCostDollars`); this was $0 on non-OpenRouter before ✅ |
| **Stop at the finish race**: abort 0–6 ms after the provider's `finish` part, single and multi-step, 11 timings | either `steps=0` + estimate, or `steps=1` + no estimate | never both | **no double count at any timing** ✅ |
| Stop during step 2's TTFB (doStream not yet resolved) | 1 / 800+30 | none | step 1 only: under-charge, conservative |
| Stop before the only step's provider responds | 0 | none | $0: conservative |
| Stop **during tool execution** of step 1 | 0 / rejected (the provider's 800/30 `finish` part had arrived and ai@6 drops it) | prompt estimate, empty output | ~8 tokens instead of 830: under-charge, conservative |

Settle-once: each of the three sites still makes exactly one `trackUsage({ holdId })`. `trackUsage` settles or releases the hold on every branch (`consumeCredits` deletes it; the unpersisted and throw paths call `releaseHold`). The PR does not use `onAbort`. Billing reads only `runAgentWithRetry`'s return inside `execute()`, so no onAbort/onFinish pair can both charge. `consumeCredits` is also idempotent per `aiUsageLogId`.

**"The charge never exceeds the hold" holds only for the interrupted step.** Finished steps bill their real provider cost, and the total can exceed the hold. That was already true on master, and it is the right behaviour.

### (b) Estimator and price
- Counting uses `estimateTokens` (4 chars/token), the same estimator `ai-context-calculator` uses for system prompt and tool tokens. As the PR body says, the **reservation** has no text estimator: it prices a fixed 4000-in/1000-out budget. So "same estimator as the reservation" is not literally true. The PR states this honestly.
- Price: `calculateCost(model, …)` at the catalog rate, pre-markup. Markup is applied once, in `consumeCredits` (`chargeMillicents`, integer millicents, `MARKUP_BPS`; chat passes no override). The cap `holdCents/100/(MARKUP_BPS/1e4)` round-trips back to the hold's cents through `Math.round`. One credit definition is used and no float reaches the stored figure.
- The page route prices with `resolvedModelName` and the global route with `providerResult.modelName`, the same ids the existing fallback uses. The **hold**, though, is sized on `selectedModel` (client id) in both routes, so the cap and the hold are computed on possibly different ids. When the selected id is not in the catalog, the hold is the flat ceiling and the cap is ≤ it, so that direction is safe. The cap exceeds the real hold only if both ids are catalogued and the resolved model is pricier. Low severity.

### (c) No regression: ran locally (source-alias vitest; no dist build)
- PR suites: `run-agent-with-retry-abort` 6/6, `agent-run-billing` 5/5, `run-agent-with-retry` 11/11.
- Existing: page `stream-socket-events` 66/66 (includes the new abort test), page `credit-gate` 4/4, `credit-abort` 5/5, global `stream-socket-events` 34/34, global `credit-gate` 10/10, `handle-chat-turn` 13/13. The whole `lib/ai/core/__tests__` directory is green.
- The route suites could only be collected through a throwaway copy that spreads `importOriginal` under each `@pagespace/*` / `@paralleldrive/*` mock. This is a known artifact of the source-alias harness; CI externalizes dist.
- The normal-completion and error-chunk paths return exactly the old fields: `!aborted` is a pass-through, and the real-stream tests pin `abortedStep: undefined`. Disconnects are untouched: nothing in the diff aborts the model.

### (d) Mutations, by line index (control GREEN 77/77 across the 3 billing test files)
| Mutant | Result |
|---|---|
| M1 `run-agent-with-retry.ts:304` in-flight step never captured (the $0 abort path) | **RED** (4) |
| M1b `agent-run-billing.ts:72` ignore `abortedStep` (old master fields) | **RED** (5) |
| M2 `run-agent-with-retry.ts:299` keep the token-less `totalUsage` (drops finished steps) | **RED** (3) |
| M3 `agent-run-billing.ts:89` remove the cap | **RED** (1) |
| M4 `page-chat-turn.ts:2092` second `trackUsage` on abort (breaks settle-once) | **RED** (1) |
| M5 `agent-run-billing.ts:104` keep reconcile ids on the estimated row | **RED** (1) |
| M6 `run-agent-with-retry.ts:308` drop the prior-step floor | **RED** (2) |
| **M7 `global-chat-turn.ts:1506` pass `abortedStep: undefined` (global route ignores the fix)** | **GREEN 44/44: survives** |

### (e) Is the policy conservative?
**For output, yes**: only text, reasoning and tool input that reached the pipe are counted, which is at most what the provider produced. **For input, mostly under-counted** (Codex's point on :307 is one more under-count: turn context is omitted). The cases where it can charge **more than the provider billed us** for the interrupted step:
1. **Prompt caching.** The interrupted step's prompt is priced at the full fresh-input rate, but this route sets Anthropic cache breakpoints. On step ≥ 2 almost all of `priorStepContextTokens` are cache reads, billed at `CACHE_READ_DISCOUNT_FACTOR_BPS` = 10%. The finished step's `cachedInputTokens` is known and is not applied. Up to ~10× the real input cost for that step, bounded by the cap.
2. **Reasoning in the floor.** `priorStepContextTokens = input + output`, and ai@6's `outputTokens` includes reasoning. Providers that do not re-send reasoning as input (OpenAI reasoning models) are over-counted. Bounded by the cap.
3. **Tool definitions (page route).** `estimateToolDefinitionTokens(filteredTools)` JSON-stringifies Tool objects, including zod internals. One 3-param tool measured 612 chars vs 270 for its JSON Schema, although `.describe()` text is dropped. Net direction depends on the tool set. Informational.
4. **The cap usually binds, so option A behaves like option B.** The cap is the price of a 4000-in/1000-out call. On a $3/$15 model, any interrupted step whose estimated prompt exceeds ~9k tokens hits it with zero output, and system prompt + tool definitions alone likely do. So an early Stop will typically charge **exactly the hold**, which is the outcome the PR body lists against option B. Real uncached input cost at that size is ≥ the hold, so this only over-charges when (1) applies. `metadata.abortedStep.capped` will show how often it happens. This is a policy fact for the founder's decision, not a code defect.

Every over-charge is limited to the interrupted step and bounded by the hold (≤ `CREDIT_HOLD_ESTIMATE_CENTS` = 25¢ default, much less for most models). All the under-charges (TTFB Stop, tool-execution Stop, omitted turn context) are uncapped the other way.

### Non-blocking asks
- A route-level test for the **global** settle (M7 survives): a Stop during the only step settles once with `outputTokens > 0` and a cost ≤ the hold. The page error-path wiring is also untested, though `agentRun` is almost always undefined there.
- Optional: price the floor with the prior step's cached share (for example, pass the finished step's `cachedInputTokens` ratio into `calculateCost`). This removes over-charge case 1.


### Addendum: the head moved to `33f95886d` during this review (3 commits). Re-verified there too.
- `966d2fbb1` adds a guard, `steps.length <= finishedStepChunks`: estimate only a step the SDK has not already recorded. The author's new test for it is **synthetic** (a hand-built result object), and it goes red against the `300a893bf` code, so that commit had the double-count path in principle. **I could not reproduce the race on a real ai@6 stream**: 13 timings (11 finish-part races plus an abort from `onStepFinish` on the last step, single and multi-step) all showed `finish-step` reaching the pipe before the step was recorded. Treat the guard as defensive. It cannot suppress a legitimate estimate: a rejected `steps` gives 0, and an open step has exactly `finishedStepChunks` recorded steps.
- `33f95886d` resolves Codex's :307 point. The prompt is priced from `toSentMessages(baseMessages)`, one definition shared with `buildStreamText`, so it includes the turn context. This raises the estimate toward what was actually sent. The cap still bounds it.
- On `33f95886d`: control GREEN 79/79. M1, M1b, M2, M3, M4, M5, M6 plus **M8 (drop the recorded-step guard)** and **M9 (price bare history)** are all **RED**. Route regression 119/119, `handle-chat-turn` 13/13, `run-agent-with-retry` 11/11. **M7 (global wiring) still survives, 119/119.** CI on `33f95886d`: Lint & TypeScript pass, E2E pass, **Unit Tests fail on the same `turn-duplication-ratchet` 158≠161**.
- Everything in (e) still applies unchanged.

2witstudios:
[point-guard] For the lane: review 5273140849 finds no billing defect; the only blocker is CI Unit Tests failing on the turn-duplication ratchet (expects 161, measures 158, because this PR reduced duplication). Lower RECORDED_IDENTICAL_LINES to 158 and update the two doc figures the reviewer named, push once, and reply on both inline threads. The optional cached-token floor pricing is not required for this PR. The abort-billing policy itself stays Jono's call: keep the PR body's policy section, and add the reviewer's finding that an early Stop usually charges the full hold (up to CREDIT_HOLD_ESTIMATE_CENTS, 25¢ by default) because the cap binds, with over-charge beyond the provider bill possible only via cached or reasoning prompt tokens and bounded by that hold.

2witstudios (commented):
[independent-review] agent:ow-irv-2695 (resumed as ow-irv-2695b): delta-only follow-up to review 5273140849. Its findings are not repeated here.

**The blocker from 5273140849 is resolved. No new defects in the delta.** Note: this PR had already merged (01:34Z, merge 0f16b77c8) while Unit Tests on the merged head were still running. They have since passed. Code unchanged by me; nothing resolved.

**Delta `33f95886d..8d7018582`**
- `516bd343c`: exactly three `161 → 158` edits: `RECORDED_IDENTICAL_LINES` in `turn-duplication-ratchet.test.ts`, the docblock in `handle-chat-turn.ts`, and `docs/2.0-architecture/agent-sessions.md`. Nothing else changed. Local `turn-duplication-ratchet` run: 3/3 green.
- `8d7018582` merges master (brings in #2694). Of this PR's 12 files, it touches only `CHANGELOG.md`, as a clean append of #2694's entries. No billing file changes. The merge commit `0f16b77c8` has a tree identical to `8d7018582`.

**CI**
- `516bd343c` (run 35672254564): Lint & TypeScript, E2E and Unit Tests all pass.
- `8d7018582`, the head that merged (run 35676049481): all three pass. Unit Tests: 21,034 passed / 6 skipped, 1,390 files, 0 failed.

**Still open from 5273140849, non-blocking, now post-merge follow-ups:**
- The global route's abort-billing wiring has no route-level test (its M7 mutant survives).
- The interrupted step's prompt is priced at the uncached rate, which over-charges cached contexts. This is bounded by the hold.

My own real-ai@6 probes this session independently reproduced both, plus the under-charge cases: a Stop before the first provider part bills $0, and a Stop during tool execution bills only the prompt estimate.

apps/web/src/lib/ai/chat-pipeline/global-chat-turn.ts
2witstudios: Added in cc1e7c3f3. This landed after #2695 merged at 516bd343c, so it is in follow-up **#2702**. `global/[id]/messages/__tests__/stream-socket-events.test.ts` › "billing a turn aborted mid-step": a Stop during the only step (ai@6 rejected steps/totalUsage shape) settles **once** with `holdId: "hold-1"`, 1000 in (900 system + 100 tools) / 50 out, $0.0025. Your mutant `agentRun: agentRun && { ...agentRun, abortedStep: undefined }` at global-chat-turn.ts:1511 now goes RED (1 failed | 34 passed). Also reset `mockTakeOverConversationStreams` in `beforeEach`: the takeover test left a persistent rejection that leaked into every later test. Leaving this open for you to verify.

apps/web/src/lib/ai/core/agent-run-billing.ts
2witstudios: Both over-charge sources are fixed in 398926526, in follow-up **#2702** (#2695 merged before it landed). (1) Cached share: `priorStepCachedTokens` = prior step cache reads + cache writes, billed at the cache-read rate through `calculateCost({ cachedInputTokens })`. (2) Reasoning: the floor is now `input + (output − reasoning)`. Both only lower the estimate. Tests: real ai@6 stream (800 in with 600 read / 100 write, 30 out with 10 reasoning → floor 820, cached 700), plus the priced case $0.0028. Mutants RED: no cached discount, reasoning kept, cache writes dropped. The cap binding on most early Stops (A behaving like B) is stated in #2702 as Jono's policy call; no code change for that. Leaving this open for you to verify.

Issue #2627, opened 2026-09-14
Drive + button lock icon out of sync with actual permissions


2witstudios:
## Bug
The **+ (create) button** in drives is out of sync with the user's actual permissions. A user who **has** create/edit permissions in the drive still sees the **lock icon** on the + button, blocking them from creating pages.

## Steps to reproduce
1. Grant a user create permissions in a drive (via role/page permissions)
2. User opens the drive
3. The + button shows a lock and is non-functional, despite the permission being granted
4. User **can** actually create pages through other paths (e.g., command palette / API) — the block is UI-only

## Expected
The + button reflects the effective permission set: enabled when the user can create pages.

## Actual
Button renders locked/disabled even though the user holds the required permissions.

## Notes
Suspect the permission source feeding the + button's gating is computed from a different (stale or narrower) source than the actual authorization check — e.g., reading a cached membership/role snapshot at drive load instead of effective permissions.

PR #2697, opened 2026-09-22
chore(signin): merge master into the integration branch — migrations renumbered to 0297

2witstudios:
Merges `origin/master` (177 commits, tip `12ef8f23b`) into the Sign in with PageSpace integration branch, so Phase 3 (SDK sign-in) starts from a current base. It's a real merge commit (`ed973995d`), not a rebase. Epic board: PageSpace page `yv08hib74nrtmksdzxmf5nkw`.

## Migrations: ours moved to 0297

Master took `0295_apple_sign_in_tokens` and `0296_accessible_page_ids_denies_utc`, the same numbers our two migrations used. Our migrations were never applied to production and the runner is hash-keyed, so renumbering is safe.

What I did:
1. Took master's `_journal.json`, `0295_snapshot.json`, `0296_snapshot.json` and both 0295/0296 SQL files unchanged.
2. Deleted our `0295_spotty_doctor_octopus.sql` and `0296_sleepy_iron_lad.sql`, along with our 0295/0296 snapshots (master's replaced them in step 1).
3. Ran `bun run db:generate`, which produced `0297_hesitant_ultragirl.sql` plus `0297_snapshot.json` and a new journal entry. No SQL or journal was edited by hand.

**Before/after comparison.** I sorted the statements from both sides and diffed them: the two sets are **identical, 11 statements, nothing extra**.

| Deleted file | Statement | In 0297 |
|---|---|---|
| 0295_spotty_doctor_octopus | `oauth_clients` ADD `allowedGrantTypes` jsonb DEFAULT '[]' NOT NULL | ✅ |
| | ADD `allowedScopes` jsonb DEFAULT '[]' NOT NULL | ✅ |
| | ADD `ownerUserId` text | ✅ |
| | ADD `logoUrl` text | ✅ |
| | ADD `homepageUrl` text | ✅ |
| | ADD `description` text | ✅ |
| | ADD `verified` boolean DEFAULT false NOT NULL | ✅ |
| | ADD `updatedAt` timestamp DEFAULT now() NOT NULL | ✅ |
| | FK `oauth_clients_ownerUserId_users_id_fk` → users.id ON DELETE SET NULL | ✅ |
| | INDEX `oauth_clients_owner_user_id_idx` (ownerUserId) | ✅ |
| 0296_sleepy_iron_lad | `workflows` ADD `credentialCeiling` jsonb | ✅ |

`db:migrate` ran 0000→0297 cleanly on a fresh Postgres 17. I then queried `information_schema` and `pg_constraint`/`pg_indexes`: all 9 columns, the FK and the index are there.

## Conflict: `apps/web/src/app/api/drives/route.ts`

Git's textual conflict was only the import block. I kept our principal-neutral imports and added master's `resolveDriveWideCanEdit`. Master's `app-permissions` import is no longer referenced, so I dropped it.

The auto-merged body keeps both sides' logic:
- **Master (#2654, #2627):** the scoped listing still batches `resolveDriveWideCanEdit` over every drive it lists and returns `canCreatePages`. For inherited (`role: null`) scopes, `canCreatePages` still comes from the owner's actual permissions.
- **Ours (Phase 2):** OAuth is admitted, `admitNoContentOAuth` is kept, scope decisions still go through `isDriveScopedPrincipal` / `getAllowedDriveIds` / `getPrincipalDriveMembership` / `isPrincipalDriveMember`, and the guard-(d) annotations stay.

**One real reconciliation.** Master's inherited-scope branch called `getUserAccessLevel(userId, drive.id)` directly, and guard (d) failed on it (user-keyed authority in a route that admits mcp/oauth). It now calls the credential's own drive resolver, `getPrincipalDriveAccessLevel(auth, driveId)`. The helper takes it as a `canEditDrive` callback, and the helper never reads the user directly. For an inherit row that resolver returns `getUserAccessLevel(owner, driveId)`, so master's behaviour is unchanged.

Master's three scoped `canCreatePages` tests mocked `isScopedMCPAuth` and `app-permissions`, which this route no longer calls. I re-pointed them at the principal mocks; each still asserts the same thing. Mutation check: forcing `canEditDrive` to return `true` turns the inherited-scope test red.

## Semantic check: files both sides changed that merged cleanly

| File | Master's change | How both rules survive |
|---|---|---|
| `apps/web/src/lib/ai/tools/actor-permissions.ts` | `57858ab45`: an agent may edit its own page subtree (Agent Memory) via `isAgentOwnedPage` | In `canActorEditPage` the bypass runs **after** `pageOutsideMcpScope` (drive scope) and `pageDeniedByAppToken` (explicit-role CredentialCeiling, role ∩ user). A credential can't reach past its ceiling through the subtree. The bypass is not applied to delete. Inherit rows add no tool-layer ceiling (Phase 2b design), so the agent's own ACL, now including its subtree, decides. |
| `…/tools/__tests__/actor-permissions.test.ts` | Six subtree tests (scope-only ceiling, delete not widened) | Added **one test**: an explicit-MEMBER credential ceiling with view-only on the memory page denies edit inside the subtree. Mutation check: moving the bypass ahead of the ceiling turns it red. The test calls `mockDbWhere.mockReset()` because an earlier test in master's block leaves a queued `mockResolvedValueOnce` that `clearAllMocks` doesn't drain. |
| `packages/lib/src/permissions/__tests__/account-permissions.test.ts` | `783f65877`: delegation fixtures gain `scope` | Fixture shape only; no access rule changed. lib permissions: 1660/1660. |
| `apps/web/src/app/api/drives/__tests__/oauth-scope.test.ts` | `540daf458`: `canCreatePages: true` added to `DriveWithAccess` fixtures | Fixture shape only. Our OAuth scoping assertions (manage_keys, account-scope, fail-closed) are unchanged and pass. |
| `apps/web/src/app/api/oauth/authorize/__tests__/route.test.ts` | `540daf458`: `customRoleId: null` added to a `getDriveAccess` fixture | Fixture shape only; the step-up / `no_access` scope-cap assertions pass. |

### `canCreatePages` now uses the role rule AND the credential ceiling (`385e06824`)

Third-party apps read this flag, so it must never claim a create the real gate refuses. For every drive-scoped credential, `canCreatePages` is now:

> master's drive-wide role rule (`resolveDriveWideCanEdit`, still batched) **AND** the credential's own drive-root edit answer (`getPrincipalDriveAccessLevel`)

That second half is the credential ceiling: an explicit role never exceeds its user, and an inherited scope is exactly its user. Before this change, an explicit-MEMBER key whose user had been bound view-only still reported `true`. The inherited-scope case gives the same answer as master's rule, with no carve-out.

`can-create-pages.integration.test.ts` (new) runs against Postgres with the real token lookup, principal dispatch, ceiling resolvers and custom-role read. It covers `mcp_` keys with explicit, custom-role and inherit rows, plus an OAuth `drive:D:member` grant, and checks 5 cases: user can edit / user bound view-only / key's custom role view-only / key's custom role edit but user bound view-only / inherited key.

### Finding: needs a decision, not changed here — the drive-root gate ignores a credential's custom role

This was already on **both** parents (master and the integration branch); the merge didn't introduce it. The independent reviewer found it too.

- **Where:** `packages/lib/src/permissions/app-permissions.ts:107`, in `resolveExplicitAppRoleAccess`. It returns `canEdit: true` for **any** explicit role at the drive root, before it reads `customRole`.
- **What happens:** a key or grant with explicit MEMBER plus a custom role whose drive-wide permission is view-only can still create root pages. This goes through `POST /api/pages` with no parent, or the AI `create_page` tool, because the gate computes true ∩ user.
- **Why it's a gap:** #2627 bounded this for session users (the drive-root branch of `getUserAccessLevel` now uses `resolveDriveWideCanEdit`). The token resolvers were not updated, although `resolveDriveWideCanEdit`'s comment says they mirror it.
- **Effect on the flag:** it now reports `false` for such a key while the gate allows the create. It never *over*-claims, but it disagrees with the gate in the safe direction.

Changing the gate is an auth behaviour change, so I've raised it for a decision rather than fixing it in a merge PR.

(My first version of this body described the gate the wrong way round: it said the gate was custom role ∩ user at the drive root. It isn't.)

## Vacuous-test check: seams and mutations

Every mutation was made by line index with an anchor assertion. Lib mutations were rebuilt into `dist` before the web run, because web imports lib from `dist`. Unmutated control: web 209/209, lib 74/74. After restoring everything the control is 209/209 again and the tree is clean.

| File (master-changed) | Seam under test | Real or mocked |
|---|---|---|
| `lib/ai/tools/__tests__/actor-permissions.test.ts` | `canActorEditPage`, credential-ceiling dispatch, `isAgentOwnedPage` | All three **real**. Lib `app-permissions` is mocked, so its `role ∩ user` is **not** reached from this file; it's covered by the lib `app-permissions.test.ts` and the Postgres test below. |
| `app/api/drives/__tests__/oauth-scope.test.ts` | `@/lib/auth` principal dispatch and lib `app-permissions` | Both **real**. Only authentication, the membership reads and (new) the drive-level ceiling read are stubbed at the lib boundary. The flag's `role ∩ user` is covered by the Postgres test. |
| `app/api/drives/__tests__/route.test.ts` (master's scoped block) | route composition | `@/lib/auth` **mocked**; this is a composition unit test. The call-through coverage is `can-create-pages.integration.test.ts`. |
| `app/api/oauth/authorize/__tests__/route.test.ts` | `getDriveAccess` fixture | Master changed fixture shape only; nothing our phases replaced is mocked there. |
| `services/api/__tests__/permission-management-service.test.ts` | — | **Unchanged** by master since the fork (`git log 70a2b39786..origin/master` is empty for it and its source). |

| Mutation | Tests that go red |
|---|---|
| **W1** route: drop `&& canEditDrive` (our ceiling in the flag) | 5: route.test ×2, can-create-pages.integration ×3 |
| **W2** actor-permissions: drop master's own-subtree bypass | 2: master's subtree tests |
| **W3** actor-permissions: drop our explicit-role ceiling (`pageDeniedByAppToken`) | 3: mcp ceiling, OAuth ceiling, and the new subtree-under-ceiling test |
| **L1** lib: master's `resolveDriveWideCanEdit` ignores the custom role | lib: 7 (drive-wide-can-edit ×4, drive-root-can-edit ×3); web: 4 in can-create-pages.integration |
| **L2** lib: our ceiling `role ∩ user` becomes role only (`intersectPermissionLevels` canEdit) | lib: 3 (credential never exceeds its user); web: 2 in can-create-pages.integration |

## Guard suite against the merged tree

`apps/web/src/app/api/__tests__/oauth-allowlist-guard.test.ts` passes (a)–(d) after the drives fix above. Guard (d) was the only failure before it.

Master added 5 routes since the fork. None admits `mcp` or `oauth`, so guard (a) doesn't apply and nothing was widened:

| New route | Auth |
|---|---|
| `account/apple-sign-in` | `['session']` |
| `ai/btw` | `['session']` |
| `auth/apple/notifications` | unauthenticated Apple server-to-server webhook (authenticated by Apple’s JWT signature) |
| `user-reports` | `['session']` |
| `users/[userId]/block` | `['session']` |

## Validation

Local runs are scoped because of machine load; CI runs the full monorepo gate. The integration suites ran against a throwaway Postgres 17 migrated from scratch, with the session in UTC.
- `packages/db`: 51 files / 750 tests ✅
- `packages/lib` `src/auth src/permissions src/oauth`: 84 files / 1660 tests ✅
- `apps/web` `lib/auth, api/auth, api/oauth, api/drives, api/calendar, lib/calendar, api/commands, lib/ai/tools`, plus the guard: 258 files, 5366/5367. The one failure was `session-fixation.test.ts` hitting its 5s timeout under load; it passes 17/17 on its own.
- `oauth-parity.integration.test.ts` (108) and `third-party-signin.integration.test.ts` (8): 116/116 ✅
- `bunx eslint` on touched files ✅

CI (dispatched with `gh workflow run test.yml --ref pu/signin-refresh`, because PRs into the integration branch don't trigger CI): run `35675231238` on tip `385e06824`. The superseded run `35673207768` on `ed973995d` had already passed Lint & TypeScript and E2E before I cancelled it.

Review: Codex is out of review quota, and CodeRabbit skipped the PR (346 files, over its limit). An independent reviewer agent substituted; its one finding is the drive-root gate issue above.

Nothing here changes auth behaviour beyond reconciling the two sides. No Phase 3 work.

https://claude.ai/code/session_016KtqBw6bnfHkHmRar9DJCY


2witstudios:
@coderabbitai review

PR #2698, opened 2026-09-22
feat(wallets): credit_balances becomes the wallets table (C1: WAL-1, WAL-2, WAL-5, X-5)

2witstudios:
## Lane ow-c1: wallets schema, where credit_balances BECOMES the wallets table (Wave C, lands first and alone)

Leaf h61whcqzkh9b28abi7tqqwd1 · Phase 2 task yocrxkxb0qf60k37f87b3zhy · Spec drc7x34unhc0ty1dc0u1j3gy (WAL-1, WAL-2, WAL-5, X-5) · the only Wave C migration.

**This touches live money.** `credit_balances` holds every user's real balance. This PR renames that table in place, so there is no copy and no second balance store (X-5, decided at Standup 0).

### Migrations 0298–0302 (latest on the base was 0297_aromatic_smiling_tiger)
| # | kind | what |
|---|---|---|
| 0298 | generated | `ALTER TABLE credit_balances RENAME TO wallets`. Constraints are dropped and re-added under the same names. No row is touched. |
| 0299 | custom (drizzle-kit --custom) | Drops the old `credit_balances_pkey`. drizzle-kit cannot name a primary key and only emits a commented TODO, so this needs its own migration. |
| 0300 | generated | Adds `id` (new PK; each existing row gets a distinct default), ownerType/orgId, subject, parent, spentCents, status, donationsEnabled, fallbackRule, createdAt; `wallet_consumer_caps` keyed (walletId, consumerKey) (WAL-7); `walletId` on credit_ledger, credit_holds and ai_usage_logs; CHECKs that mirror every credit_balances non-negativity rule, plus shape CHECKs and partial unique indexes (one personal root per user, one pool per org, one wallet per subject). |
| 0301 | custom | **The backfill**: every ledger and hold row gets its owner's personal root wallet. It is idempotent (only touches NULL rows, ON CONFLICT DO NOTHING) and ends with a guard that RAISEs if any row is still missing a wallet. |
| 0302 | generated | `walletId` NOT NULL on ledger and holds. |

**Why the backfill is a migration and not only a script:** 0302's NOT NULL must run in the same migrate invocation, and after the backfill. `scripts/backfill-wallets.ts [--dry-run]` runs the exact 0301 statements, read from the file, in one transaction. `--dry-run` rolls back. The script refuses to commit if any balance total (monthly, top-up, debt, pending) would change.

**Reality note:** 0301 gives a **zero-balance** personal wallet to any user who has ledger or hold rows but no balance row (billing-off ceiling holds). It adds no money. So wallets after = credit_balances rows + those users. The X-5 test pins exact equality on a corpus without such users. The (partial) test pins the orphan case.

**Leaf names vs columns:** the leaf's `monthlyAllocationCents` is the existing `monthlyAllowanceCents` (kept so the rename moves no money). `monthlyRemainingCents` stays as the root wallet's allowance bucket. Periods remain `monthlyPeriodStart/End` (timestamptz, UTC).

### Money invariants: real Postgres 17, `packages/db/src/__tests__/wallets-migration.test.ts`
Each scenario migrates a template DB to 0297, seeds balances (paid with rollover, free, in debt at a negative net, bare top-up, zero), ledger rows and holds the way the old code wrote them, then runs 0298–0302.

| Invariant | Test |
|---|---|
| (a) each user's personal root wallet = their old balance to the cent, bucket by bucket, so spendable is the same (incl. debt) | `X-5: credit_balances becomes wallets in place …`, `X-5 (partial) WAL-1 (partial): every balance lands …` |
| (b) sum before = sum after (monthly, top-up, debt, pending, net) | same two |
| (c) row counts equal: one root per former row, no duplicates, no orphans | `X-5: …` (exact equality); orphan case in `X-5 (partial) WAL-1 (partial)` |
| (d) every ledger and hold row carries its owner's root wallet; new NULL writes refused | `X-5: …`, `X-5 (partial) WAL-5 (partial): every ledger and hold row …` |
| (e) running the backfill twice changes nothing (byte-for-byte snapshot) | `X-5: …`, `X-5 (partial): running the backfill again …` |
| (f) --dry-run writes nothing (full-table snapshot equality + NULL count), and the real run then does exactly what it reported | `X-5: …`, `X-5 (partial): --dry-run reports …` |
| constraints: non-negativity, owner/subject/status shapes, one root/pool/subject, caps | `WAL-1 (partial) WAL-2 (partial): the wallets table refuses …` |
| user delete still cascades wallet, ledger, holds | `X-5 (partial): deleting a user still cascades …` |

```
DATABASE_URL=postgresql://postgres@127.0.0.1:5647/pagespace_test TZ=UTC \
  bun run --filter @pagespace/db test -- src/__tests__/wallets-migration.test.ts src/schema/__tests__/wallets.test.ts
  Test Files  2 passed (2)   Tests  25 passed (25)
```

### WAL-5 additions in this PR (beyond the WIP)
- `consumeCredits` writes the charged wallet onto the **AI usage row**. On the paid path this happens inside the settle transaction (the wallet whose buckets were debited). On the zero-charge path it uses the wallet the claim names. It only writes while the row's wallet is NULL, so a retry never re-points it. Historical usage rows stay NULL; their charge lives on the backfilled usage ledger row.
- `cost-reconcile.applyCorrection` locks and moves the buckets of **the wallet the base charge was billed to** (by id), not the owner's personal root.

### ID / test table
| ID | Status | Evidence |
|---|---|---|
| X-5 | **covered**, removed from `scripts/spec-coverage-allowlist.txt` (reviewer please confirm) | one test proves every clause: in-place, one store, walletIds in the same migration, idempotent, dry-run, row-count + sum equality |
| WAL-1 | partial, stays allowlisted | schema, migration, constraint and personal-wallet tests |
| WAL-2 | partial, stays allowlisted | unique pool/subject/root indexes, a drive wallet under a parent |
| WAL-5 | partial, stays allowlisted | usage-row wallet, ledger/hold walletId, wallet-keyed reconcile. Gate reset and settle still lock the root **by user** (identical while every charge is on a personal root); that is C3's job |
| WAL-7 | partial | caps table shape |

### Mutation checks (by line index, a no-op control green first, sources restored, `git status` clean)
| Line broken | Went red |
|---|---|
| `packages/db/src/wallet-backfill.ts:129` dry-run `ROLLBACK`→`COMMIT` | `X-5: …`, `X-5 (partial): --dry-run …` |
| `0301…sql:40` `w."userId" = l."userId"` → `<>` | 2 tests (X-5, WAL-5 walletIds) |
| `0301…sql:35` injected `UPDATE wallets SET topup+1` | 5 tests (money, sum, idempotent, dry-run) |
| `credit-consume.ts:187` usage-wallet stamp removed | `WAL-5 (partial): every AI usage row that settles …`, consume debt-row test |
| `cost-reconcile.ts:220` update by `userId` instead of wallet id | `WAL-5 (partial): reconciliation is keyed on wallet …` |

### Callers (Retro 1 rule 13): every file in the leaf's list
**credit_balances**
- `packages/db/src/schema/credits.ts`: `creditBalances` removed; `walletId` (NOT NULL, FK cascade) + index on ledger and holds. New `schema/wallets.ts` (+ `exports` entry, knip entry).
- `packages/lib/src/billing/credit-balance.ts`: both balance reads use `personalRootWalletOf(userId)`.
- `credit-consume.ts`: settle locks the root; claim and debt rows carry `walletId`; the usage row records the wallet.
- `credit-gate.ts`: reads/locks the root; holds, grants and starter rows carry `walletId`; lazy-init inserts use the partial-index arbiter; billing-off ceiling holds get a bare zero wallet.
- `credit-funding.ts`: refill/top-up call `ensurePersonalRootWalletId` before the claim (this replaces the old stub-row insert, so FOR UPDATE still has a row to lock); the upsert uses the partial-index target.
- `cost-reconcile.ts`: corrections land on the charge's wallet (by id).
- `missed-grant-reconcile.ts`: ensures and locks the root.
- `credit-core.ts`, `money-model.ts`: comments only.
- new `personal-wallet.ts`: the one find-or-create-bare path (race-safe, adds no money).
- `apps/web/src/lib/monitoring/monitoring-queries.ts` and `apps/admin/src/lib/monitoring-queries.ts`: debt, liability, drift and top-debtor queries read `wallets` filtered to personal roots; drift joins the ledger on `walletId`.
- `apps/web/src/lib/subscription/usage-breakdown-query.ts`: period read from the root.
- `packages/lib/src/compliance/export/gdpr-export-coverage.ts`: `wallets`, `wallet_consumer_caps` excluded, as `credit_balances` was.
- `scripts/backfill-credit-ledger.ts`: drift per personal root; adjustment carries `walletId`.
- `scripts/send-free-credits-change-notifications.ts`: LEFT JOIN on the root.
- `apps/e2e/support/db.ts`: seeding and reading via the root; holds and ledger carry `walletId`.

**credit_ledger / credit_holds** (none of these needed a change)
- `packages/lib/src/billing/credit-backfill.ts`: deletes expired holds and reads ledger ids; no inserts. Its settles go through `consumeCredits` / `settlePendingLedgerRow`, which now stamp wallets.
- `live-concurrency-query.ts`: counts a user's live holds (the per-user in-flight cap). The additive column does not affect it.
- `credit-pricing.ts`, `apps/web/src/app/api/credits/concurrency/route.ts`, `apps/web/src/app/api/stripe/webhook/route.ts`, `ConcurrencyCard.tsx`, `image-generation-tools.ts`, `usage-breakdown.ts`, `services/app-hosting/{app-machine-events,router}.ts`: they only mention the tables in comments or call the billing functions above. Every ledger/hold INSERT in the repo (grep for `insert(creditLedger|creditHolds)`) now passes `walletId`, and tsc enforces it (NOT NULL).

**ai_usage_logs** (nullable `wallet_id` added)
- `logger-database.ts` (writer), `ai-monitoring.ts`, `workflow-executor.ts`, `ai-usage-purge.ts`, `gdpr-export.ts`, `monitoring-retention.ts`, `retention-engine.ts`, `api/pages/[pageId]/ai-usage/route.ts`, admin/web compaction routes, monitoring and usage-breakdown: all unchanged. The column is nullable, and none of them enumerate columns or build `$inferSelect` rows. The writer leaves it NULL and `consumeCredits` fills it at settle.

### Deploy window (please triage; not a decision I can make)
Migrations run in the one-shot migrate machine **before** the new app rolls out. Between the migration commit and rollout, old pods query `credit_balances` (gone) and insert ledger/hold rows without `walletId` (refused). While that lasts:
- the AI gate errors;
- settles defer to the backfill cron, which recovers them;
- Stripe funding webhooks fail and Stripe retries them.

X-5 forbids dual-read aliasing (e.g. a `credit_balances` view), so I did not add one. Options are a short maintenance window or accepting the retry path.

### Gates run locally (build slot respected: single files, own PG17, no build)
- `bunx tsc --noEmit -p packages/db` → exit 0; `bunx tsc --noEmit -p packages/lib` → exit 0 (lib maps db to source).
- lib, all of `src/billing` from source against migrated PG17: `bunx vitest run --config <source-alias config> src/billing` → **30 files, 649 passed**.
- web touched tests: **5 files, 51 passed**; admin touched tests: **2 files, 15 passed**.
- `bunx eslint` on touched web/admin/lib/db files → 0 errors.
- `bun run knip:check`: no findings in files this PR touches (the 5 reported are untouched env-bridge/drive-envs/app-hosting files, the known pu-worktree phantoms, plus my uncommitted throwaway config).
- apps/web typecheck: `bunx tsc --noEmit -p apps/web/tsconfig.wt.json` (a throwaway config mapping @pagespace/db to source) → exit 0, 0 errors.
- Full suite, monorepo typecheck and build: left to CI.

Not user-visible (every balance is unchanged), so no changelog entry.

https://claude.ai/code/session_01P4bzc1xswTVdGWA2JenPZ4


2witstudios:
@codex review

packages/db/drizzle/0302_wallet_ids_required.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "credit_holds" ALTER COLUMN "walletId" SET NOT NULL;--> statement-breakpoint
```
2witstudios: **P1: the migration chain can abort in production, because old pods keep writing NULL-walletId rows while 0301 and 0302 run.**

The runner commits each migration separately. After 0298 commits, old pods still insert `credit_ledger` rows with no walletId. In-flight AI calls hit the `consumeCredits` claim, which runs outside the settle transaction and before any balance read. The old credit-backfill orphan sweep calls the same claim. On tenant and onprem, a ceiling-gated billing-off gate inserts `credit_holds` without ever touching the balance table.

- A row that commits **during** 0301's ledger UPDATE is not in that statement's snapshot. 0301's own guard (a later statement with a fresh snapshot) then RAISEs, and 0301 rolls back.
- A row that commits **between** 0301 and 0302 makes `SET NOT NULL` fail with "column contains null values".

Either way, migrate exits non-zero with 0298–0300 committed and old pods on a schema they cannot use. AI returns 500 for every cloud user until someone runs `scripts/backfill-wallets.ts` and migrate again, and that retry races the same writers.

I reproduced the first case on PG17. I reset the schema to its post-0300 state and ran 0301 in one transaction with a pause after the ledger UPDATE. I committed one old-shape claim from a second session:
```
INSERT INTO credit_ledger (id,"userId","entryType",bucket,"amountCents","aiUsageLogId") VALUES ('l_oldpod','u1','usage','monthly',-7,'log_x')
-- 0301 session:
ERROR:  wallets backfill incomplete: 1 credit_ledger and 0 credit_holds row(s) have no walletId
```
No money is lost, because every failed write is retried or swept, but the deploy fails partway through.

Fix options, for the orchestrator to triage:
- (a) One custom migration that takes `LOCK TABLE credit_ledger, credit_holds IN SHARE ROW EXCLUSIVE MODE`, re-runs the 0301 statements, and then sets NOT NULL, all in one transaction. Blocked writers then wait instead of racing, and a trailing NULL cannot slip between backfill and constraint.
- (b) A documented drained deploy: old machines and crons stopped before migrate runs.

The PR body's deploy-window section assumes migrate succeeds. It should cover this failure mode too.

packages/db/src/wallet-backfill.ts
```diff
@@ -0,0 +1,146 @@
...
+  await client.query('BEGIN');
+  try {
+    const before = await readCounts(client);
+    for (const statement of statements) {
+      await client.query(statement);
+    }
+    const after = await readCounts(client);
+
+    const drift = moneyDrift(before, after);
+    if (drift.length > 0) {
```
2witstudios: **P2: the "refuses to commit if any money moved" guard is untested.**

This is the script's stated safety net on a live-money repair path. I changed this line to `if (drift.length < 0) {` by line index, which disables the refusal. `wallets-migration.test.ts` stayed **16/16 green**, while a no-op control was also green.

The lane's own mutation on 0301:35 went red only through the balance-equality assertions, not through this guard. `moneyDrift` is unit-tested as a pure function, but no test calls `runWalletBackfill` with a statement list that moves money. Nothing checks that the runner throws and that a snapshot afterwards is unchanged.

Suggested test: `runWalletBackfill(client, { dryRun: false, statements: [...walletBackfillStatements(), 'UPDATE "wallets" SET "topupRemainingCents" = "topupRemainingCents" + 1'] })` rejects with `would move money`, and `fullSnapshot` is byte-identical afterwards.

scripts/backfill-wallets.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * The X-5 wallets backfill, outside the migrator (Spec X-5, WAL-5).
+ *
+ * Migration 0301 already runs this backfill on every deploy, in the same migrate
+ * invocation as the walletId NOT NULL (0302) that depends on it — so there is never a
+ * window where a ledger or hold row lacks a wallet. This script runs the SAME statements
+ * (read from the 0301 file) in one transaction, for two jobs:
+ *
+ *   --dry-run   rehearse: report what the backfill would do, then ROLL BACK. Writes
```
2witstudios: **P2: the dry run cannot rehearse production before the deploy.**

The dry run executes only the 0301 statements. It therefore needs a database already migrated through 0300. Production is at 0297 until the deploy, and the deploy runs 0298–0302 in one migrate invocation. So production is never in the state this rehearsal needs:
- Before the deploy, `--dry-run` fails on the missing `wallets` table and `walletId` column.
- After the deploy it reports zero work.

X-5's "has a dry-run" holds in letter, and the test proves it on a throwaway database. Please document the real rehearsal procedure here and in the PR: restore a prod snapshot into a scratch database, migrate it through 0300, then run `--dry-run`. Otherwise someone will run it against prod expecting a pre-flight check.

packages/db/src/__tests__/wallets-migration.test.ts
```diff
@@ -0,0 +1,619 @@
...
+    await adminPool.end();
+  }, 120_000);
+
+  it('pins the base: 0297 is the last migration before this change', () => {
+    expect(allMigrations.length).toBe(journal.entries.length);
+    expect(journal.entries[297]?.tag).toBe('0297_aromatic_smiling_tiger');
+    expect(throughThisChange.length - baseMigrations.length).toBe(5);
+  });
+
+  it('X-5: credit_balances becomes wallets in place — dry-run writes nothing, row counts and every balance and the sum equal to the cent, every ledger and hold row on its owner\'s root, a re-run changes nothing', async () => {
```
2witstudios: **P2: the plain `X-5:` claim does not prove the "in place" clause by itself.**

X-5 says each balance row is migrated **in place**. This test would stay green if 0298 were `CREATE TABLE wallets … ; INSERT … SELECT … ; DROP TABLE credit_balances`. Values, counts and `to_regclass` would all match. Only the separate `X-5 (partial): 0298 renames credit_balances in place …` static test pins the rename.

Under the rule that a plain ID claim must prove every clause of its Spec line, one cheap assertion closes the gap. Capture `SELECT 'credit_balances'::regclass::oid` before `s.migrate(throughExpand)`, and assert it equals `'wallets'::regclass::oid` afterwards. A rename keeps the relation OID. A copy does not.

2witstudios (commented):
[independent-review] agent:ow-irv-2698

Independent spec review of PR #2698 (lane ow-c1b) at head `d908f93ed788f32d708d395415cb4db406cd326d`. Codex had not reviewed this head. **Note: this PR merged into `pu/org-wallets` (3f6a30ce3, 2026-09-22 01:48 UTC) with zero reviews on record.** This review is therefore the gate evidence before anything goes to master, and the P1 below must be fixed on the integration branch before it goes further.

**Verdict: 1 P1, 3 P2.** The money move is correct and well tested. Balances land to the cent, no second store exists, and the backfill is idempotent. The P1 is a deploy-time race that can abort the migration chain partway through. That would leave production on the renamed table with old pods serving. It is an outage risk, not a money-loss risk.

### Per-ID verdicts
| ID | Claimed as | Verdict | Evidence |
|---|---|---|---|
| X-5 | plain (removed from allowlist) | **satisfied, with a P2 on the plain test** | `X-5: credit_balances becomes wallets in place — dry-run writes nothing, …` (packages/db/src/__tests__/wallets-migration.test.ts:346) proves one store (`to_regclass('public.credit_balances')` is null, so no view alias exists), row-count equality, per-user and total equality to the cent (monthly, top-up, debt, pending, net), every ledger and hold row on its owner's root, a dry run that leaves the full-table snapshot byte-identical, and an idempotent re-run. The **"in place"** clause is proven only by the separate `X-5 (partial): 0298 renames credit_balances in place …` static test. The plain test would also pass for a copy followed by a drop (inline P2). CI coverage job 106582170122 line 308: `X-5 covered packages/db/src/__tests__/wallets-migration.test.ts`. |
| WAL-1 | partial | correctly partial | Schema and constraint tests plus the migration money tests. The leaf's `monthlyAllocationCents` ships as the existing `monthlyAllowanceCents`, which the PR discloses. Allowlist note is accurate. CI line 295. |
| WAL-2 | partial | correctly partial | Unique root, pool and subject indexes, plus "a drive wallet the person owns is not their personal root". No pool or drive wallet is created yet. CI line 296. |
| WAL-5 | partial | correctly partial | Usage-row wallet (`WAL-5 (partial): every AI usage row that settles records the wallet charged …`), NOT NULL walletId on ledger and holds, and wallet-keyed reconcile (`WAL-5 (partial): reconciliation is keyed on wallet …`). Settle and the monthly reset still lock by user, and the allowlist note says so. CI line 299. |
| WAL-7 | partial | correctly partial | Caps table shape only. CI line 301. |

Every `(partial)` marker has exactly one ASCII space and is lowercase. No Spec ID appears in a gate self-test title.

### What I ran myself
Throwaway homebrew postgresql@17 on 127.0.0.1:5698, `-k` socket in a short dir, TZ=UTC. It is stopped and deleted after this review.

- packages/db `wallets-migration.test.ts` and `schema/__tests__/wallets.test.ts`: 25/25 passed.
- packages/lib (source alias, real PG) `wallet-money-path`, `cost-reconcile`, `credits-flow`, `personal-wallet` integration tests plus `credit-consume.test.ts`: 97/97 passed.
- `drizzle-kit generate` on the head reports "No schema changes, nothing to migrate", so the snapshots match the schema.
- After the runs the test DB holds 0 wallets, 0 ledger rows, 0 holds and 0 drives. The only users are the two system users that migrations seed. No `psx_wallets_*` databases were left behind.

### Mutations (by line index, no-op control green first, restored, `git status` clean)
| Line | Mutation | Result |
|---|---|---|
| `0301_wallets_backfill_wallet_ids.sql:50` | holds `w."userId" = h."userId"` changed to `<>` | **red**: 2 failed (`X-5: …` and `X-5 (partial) WAL-5 (partial): every ledger and hold row …`) |
| `wallet-backfill.ts:124` | `drift.length > 0` changed to `< 0` (refusal disabled) | **survived**: 16/16 green (inline P2) |
| `credit-consume.ts:72` | `isNull(aiUsageLogs.walletId)` dropped from the usage-wallet UPDATE | **survived**: 103/103 green. The PR body says "a retry never re-points it", but no test pins that. The usage-ledger claim is idempotent, so a second stamp is probably unreachable today. Recorded here as a note, not a finding. |

### Callers
`git grep` for `credit_balances|creditBalances` across the whole monorepo at the head finds only comments, the spec doc, migrations 0298–0301 and tests. There are no hits in apps/admin, control-plane, processor, realtime or scripts, and none in the PageSpace-Deploy repo. Every `insert(creditLedger|creditHolds)` passes `walletId`. There are no raw-SQL ledger or hold INSERTs outside tests. The erasure plan names no credit table. Integer cents are kept everywhere. The refund rounding in cost-reconcile is unchanged. `MONEY_MODEL_V2_ACTIVE` is untouched and is still a code constant. No `$` appears next to a credit count. The `¢` in the ops script output labels balance cents, not credits.

### Deploy window: which requests fail, and whether money can be lost
The migration runner commits **each migration in its own transaction** (migration-runner.ts). Old pods keep serving from the moment 0298 commits until the new pods roll out.

- **AI gate** (`canConsumeAI`, billing on): it throws `relation "credit_balances" does not exist`. No caller wraps it in a try, so chat, global chat, v1 completions, pulse, voice and image generation return 500. It **fails closed**, so no AI runs and nothing is charged.
- **Balance reads** (`getCreditBalance` and the usage UI): 500.
- **In-flight streams that settle on old pods**: the claim is `db.insert(creditLedger)` outside the settle transaction. Before 0302 it **commits with walletId NULL**, then the settle throws and the call is `deferred` with the row left pending. After 0302 the claim is refused and the call is `deferred` with no row. Either way the new pods' credit-backfill pending or orphan sweep bills it once. The claim is unique on aiUsageLogId and the orphan sweep has no lookback cutoff. **Nothing is lost or double-charged.**
- **Stripe funding** (checkout, invoice.paid, subscription events): every funding path runs inside `withFundingRetry`. The ledger insert and the balance update share one transaction. On failure the webhook deletes its idempotency marker and returns 500, and Stripe redelivers. **Nothing is lost.**
- **cost-reconcile cron on old pods**: `applyCorrection` fails inside its transaction, the per-row catch logs it, and the row is retried next run without bumping attempts. **Nothing is lost.**
- **The migration itself can fail (P1, inline on 0302).** Old pods write rows with NULL walletId, as described above. Tenant and onprem deployments add ceiling-gated billing-off holds, which never touch the balance table. Any such row committed during 0301's ledger UPDATE makes 0301's own guard RAISE. Any row committed between 0301 and 0302 makes `SET NOT NULL` fail. Either way the deploy stops with 0298–0300 committed and old pods on a schema they cannot use, and the outage runs until someone intervenes.

### Rollback story (not in the PR; please add it)
There is no down path. Once 0298 commits, reverting the app image does not restore service, because the old code needs `credit_balances`. Recovery is forward-only: fix and rerun migrate, or run `scripts/backfill-wallets.ts` and then migrate. A manual reverse is possible while only personal roots exist. It would drop walletId, the added columns and the caps table, rename wallets back to credit_balances, and re-add the `userId` primary key. It would fail if 0301 created bare wallets for orphan users who also hold a balance row, which cannot happen, so that case is fine. That procedure should be written in the PR or runbook before master.

### Rules
- Pure logic stays separate from IO: `personal-wallet.ts` is IO only, and the pure `credit-core` is unchanged.
- No permissions surface is touched. There are no `any` types. There are no new env flags.
- The new `timestamptz DEFAULT now()` columns are fine, because the UTC rule applies to `timestamp` without time zone.
- The new `@pagespace/db` subpaths `./schema/wallets` and `./wallet-backfill` both have exports entries.
- Scope matches the leaf, and the only migrations are the ones the leaf names.
- The backfill is one set-based `UPDATE … FROM`, with no bind-parameter batching concern. It is unbatched over the whole `credit_ledger`, so its runtime grows with ledger size, which widens the P1 race window. Size the prod ledger before deploying.

packages/db/src/__tests__/wallets-migration.test.ts
2witstudios: Fixed in 27cc4cf37 (follow-up PR #2710). The plain X-5 test now captures `'credit_balances'::regclass::oid` before the chain and asserts `'wallets'::regclass::oid` is identical afterwards. Mutation: appending a copy-and-swap of wallets to 0302 turns the test red at that assertion (expected 61115 to be 59032), while every value, count and sum still matched. `wallets-migration.test.ts`: 18 passed (18). Leaving open for you to verify.

packages/db/src/wallet-backfill.ts
2witstudios: Fixed in 27cc4cf37 (follow-up PR #2710). New test: `X-5 (partial): the backfill runner refuses to commit a run that would move money, and writes nothing`. It runs the 0301 statements plus a one-cent UPDATE through runWalletBackfill; the call rejects with 'would move money; rolled back (topupRemainingCents: …)' and fullSnapshot is byte-identical afterwards. Your mutant (`wallet-backfill.ts:124` changed to `drift.length < 0`) now goes red: 1 failed | 16 passed, with the no-op control green. Leaving open.

scripts/backfill-wallets.ts
2witstudios: Fixed in d314b3862 (follow-up PR #2710). rehearseWalletMigration runs 0298–0302 from the migration files in ONE transaction and always rolls back. It refuses a database not at 0297. `scripts/backfill-wallets.ts --dry-run` selects it automatically on a 0297 database. The procedure (restore a prod snapshot into a scratch DB and rehearse there, never the live primary, because it holds the rename's ACCESS EXCLUSIVE lock) is in the script header and in the #2710 body. Test: the new X-5 (partial) rehearsal test; mutation finally ROLLBACK->COMMIT turns it red. I also ran the real script on a seeded scratch 0297 DB: 2 balances -> 3 roots (1 zero), ledger 3/3, holds 1/1, drift none, DB unchanged. Leaving open.

packages/db/drizzle/0302_wallet_ids_required.sql
2witstudios: Acknowledged, not fixed yet. A NEW migration cannot close this race: migration-runner.ts applies files in journal order and commits each one separately, so a new file lands at 0303 and runs after the 0301/0302 statements that abort. I posted three options to #org-wallets and am waiting on the orchestrator's call: (1) amend the merged custom 0301 to LOCK TABLE credit_ledger, credit_holds IN SHARE ROW EXCLUSIVE MODE, then backfill, then SET NOT NULL in one transaction (0302 becomes a no-op; every statement is idempotent, so the hash-keyed re-run is a no-op); (2) reorder _journal.json by hand to insert a new custom migration before 0301; (3) a drained deploy. The fix will land on #2710, test-first from your PG17 reproduction. The #2710 body now also covers this failure mode and the rollback runbook.

packages/db/src/wallet-backfill.ts
2witstudios: Verified fixed in #2710 (27cc4cf37, head d314b3862). A new test covers the runner's refusal. My original mutant (`drift.length > 0` changed to `< 0`) now goes red (1 failed / 17 passed) on my PG17. Resolving. Review: https://github.com/2witstudios/PageSpace/pull/2710#pullrequestreview-5280177346

scripts/backfill-wallets.ts
2witstudios: Verified fixed in #2710 (d314b3862). `--dry-run` on a 0297 database now rehearses 0298 through 0302 in one transaction and rolls back. The test asserts an unchanged pre-migration snapshot, the relation OID and the journal. Mutating ROLLBACK to COMMIT goes red. The restore-to-scratch procedure is documented. One new P2 on the rehearsal's drift report is filed on #2710. Resolving. Review: https://github.com/2witstudios/PageSpace/pull/2710#pullrequestreview-5280177346

packages/db/src/__tests__/wallets-migration.test.ts
2witstudios: Verified fixed in #2710 (27cc4cf37). The plain X-5 test now asserts that the wallets relation OID equals the pre-chain credit_balances OID, which a copy followed by a drop cannot satisfy. Resolving. Review: https://github.com/2witstudios/PageSpace/pull/2710#pullrequestreview-5280177346

packages/db/drizzle/0302_wallet_ids_required.sql
2witstudios: Not fixed yet. #2710 (d314b3862) changes no migration and leaves this finding awaiting a decision. I agree with the lane's option 1: amend 0301 to take the lock, backfill and SET NOT NULL in one transaction, with an idempotent-rerun test, because a new 0303 cannot close this race. Leaving this thread OPEN.

2witstudios: Fixed in 19a9cf860 (PR #2710), without editing any merged SQL. A new drizzle-kit --custom migration, 0300_1_wallets_lock_and_require_wallet_ids.sql, has its journal entry placed directly after 0300 — drizzle applies migrations in meta/_journal.json order and keys each by file hash, so this is the only way to order a new file before an already-merged one. In one transaction it takes LOCK TABLE credit_ledger, credit_holds IN SHARE ROW EXCLUSIVE MODE, backfills, guards, then sets walletId NOT NULL on both tables; 0301 and 0302 then find nothing to do and no shipped file's hash changes. Your PG17 case is now a test: an old pod's uncommitted NULL-walletId claim during migrate used to end in 'contains null values', and now the migration waits for that writer and backfills its row. Also tested: after 0300_1 alone a NULL write is already refused (no window before 0302), and a database that already ran 0301/0302 applies 0300_1 as a no-op. 23/23; mutations on the LOCK and the NOT NULL each turn their test red. Leaving open for you to verify — please sanity-check the hand-placed journal entry.

2witstudios: P1 fix verified in #2710 at head 19a9cf860 (new migration 0300_1, journal entry placed after 0300; no merged SQL edited). I reproduced the original race on PG17 against the new chain: the migration waits for the in-flight old-pod claim (2s), backfills its row, and a later NULL-walletId write is refused. Fresh migrate end-to-end exits 0 with walletId NOT NULL; 23/23 on the suite; removing the LOCK or the SET NOT NULL each turns two tests red (control green). One new P2 on that PR: the journal lists 0300_1 twice, so it runs twice. Leaving this thread OPEN until #2710 merges into pu/org-wallets. Review: https://github.com/2witstudios/PageSpace/pull/2710#pullrequestreview-5281345317

2witstudios: Closing the loop: #2710 merged into pu/org-wallets as be1cbe084, and its second parent is exactly the head I reviewed (3a4272b11). On the integration tip, migration 0300_1 carries the LOCK and the journal has 304 entries with no duplicate tag. The P1 fix is verified on the branch, so I am resolving this thread. Reviews: 5281345317 (P1 fix) and 5281755775 (final).

PR #2699, opened 2026-09-22
fix(permissions): drive-root create follows the custom role's drive-wide edit for credentials too

2witstudios:
An `mcp_` key or an OAuth `drive:D:role:<id>` grant with a view-only custom role could create pages at the top of a drive, and upload files there. A human member with the same role could do neither. This PR fixes that, and only that.

It follows up on #2697, which found the gap. Point-guard ruling: this is a bug against #2627's own invariant. `resolveDriveWideCanEdit` says "the drives DTO flag, the session check and the token checks can never disagree".

## The bug

`resolveExplicitAppRoleAccess` (`packages/lib/src/permissions/app-permissions.ts`) returned `canEdit: true` at the drive root for **any** explicit role, before it ever read the custom role. Both `getAppAccessLevel` (`mcp_` keys) and `getScopedAccessLevel` (OAuth grants) use it.

Root-page create is authorized as edit on the drive root. It goes through `POST /api/pages` with no parent (`authorizeEdit(driveId)` → `canPrincipalEditPage`) and through the AI `create_page` tool (`canActorEditPage(ctx, driveId)`). So any credential bound by a custom role could create root pages, whatever that role's drive-wide permission said. The session path had applied the #2627 rule since #2654.

**The same hole on the drive-level resolvers.** The independent review found this one. `getAppDriveAccessLevel` and `getScopedDriveAccessLevel` also hard-coded `canEdit: true` for every explicit role. `upload/presign` and `upload/complete` gate credentials on that value, and a completion with no parent inserts a FILE page at the drive root. So the same credential could create a root-level page by uploading.

**Also fixes on master:** the same drive-root branch is on master today (`git show origin/master:packages/lib/src/permissions/app-permissions.ts`, lines 102–104), so custom-role `mcp_` keys there have the same over-grant. OAuth drive grants exist only on this integration branch.

## The fix: one pure rule

A new pure function, `driveWideCanEdit`, in `membership-queries.ts`, is the only place the rule is decided. Two callers use it:

- `resolveDriveWideCanEdit`, which serves the drives DTO `canCreatePages` flag and the session check at the drive root (`getUserAccessLevel`);
- the drive-root branch of `resolveExplicitAppRoleAccess`, which serves `mcp_` keys and OAuth grants.

Because both call the same function, the flag, the session check and the token checks can't drift apart.

The two drive-level resolvers (`getAppDriveAccessLevel` for `mcp_` keys, `getScopedDriveAccessLevel` for OAuth grants) now take their explicit-role answer from that same drive-root decision, through a new `resolveExplicitDriveRootAccess`. It reads the custom role bound to the drive, just as a page-level check on the drive id does, and the result is still intersected with the user's live access.

At the drive root, for a credential:

| Role | canView | canEdit | canShare / canDelete |
|---|---|---|---|
| OWNER / ADMIN | true | true (unchanged) | true (unchanged) |
| MEMBER, no custom role | true | true (unchanged) | false |
| MEMBER + custom role | true | the custom role's `driveWidePermissions.canEdit === true` | false |
| MEMBER + custom role that doesn't resolve in this drive | false | false | false |

`canView` is `true`, matching what `getUserAccessLevel` gives a human member at the drive root. A custom role that doesn't resolve fails closed twice: the branch returns all-false, and the pure call also counts it as having a custom role (`hasCustomRole: customRole !== null || customRoleUnresolved`). A deleted or foreign-drive role therefore can't read as "no custom role" and get edit.

## Tests

**RED first: `root-create-custom-role.integration.test.ts` (new).** It runs against Postgres through the real `POST /api/pages` and the real `authenticateRequestWithOptions`, token lookup and permission helpers. The key and the grant belong to the drive **owner**, so the only thing that can refuse is the credential's own role. The human is a MEMBER bound by the same role and authenticates with a real session token sent as a Bearer.

| Operation / custom role | session (human) | `mcp_` key | OAuth `drive:D:role:<id>` |
|---|---|---|---|
| root create, view-only, before the fix | 403 | **201** ❌ | **201** ❌ |
| root upload presign, view-only, before the fix | 403 | **200** ❌ | **200** ❌ |
| root create, view-only, after | 403 | 403 | 403 |
| root upload presign, view-only, after | 403 | 403 | 403 |
| root create, edit drive-wide | 201 | 201 | 201 |
| root upload presign, edit drive-wide | 200 | 200 | 200 |

Lib pure tables (zero mocks):
- `resolveExplicitAppRoleAccess` at the drive root: 7 cases covering view-only / edit / null drive-wide, unresolved, ADMIN, ADMIN with an unresolved role, and an OWNER row.
- `driveWideCanEdit`: 7 cases.

Lib drive-level cases (red first): `getAppDriveAccessLevel` and `getScopedDriveAccessLevel` with a view-only and an edit custom role.

`app-permissions.test.ts` now resets `db.select` before every test. `vi.clearAllMocks()` kept unconsumed `mockReturnValueOnce` stubs, so a leftover from one test answered the next test's query, and a mutation showed up as a cascade of unrelated reds.

**Updated pin: `tool-ceiling-parity.integration.test.ts`.** Its CUSTOM role grants view+edit on one page and has NULL drive-wide permissions. It had pinned root `create_page` as `ok`, which was this bug seen through the agent-tool path; it's now `denied`.

## Mutation check

Every mutation was made by line index with an anchor check. Lib was rebuilt into `dist` before each web run. Unmutated control: lib 86/86, web 11/11. After the restore, both are green again and the tree is clean.

| Mutation | Tests that go red |
|---|---|
| **M1** token drive root back to `canEdit: true` (the bug) | lib pure table ×2; `root-create-custom-role` mcp_ + OAuth |
| **M2** shared `driveWideCanEdit`: a custom role always edits | lib 12 (pure rule, `resolveDriveWideCanEdit`, session drive-root bounding); web 7 (**session, mcp_ and OAuth** root create, plus 4 `canCreatePages` flag cases). One mutation reaches all three callers. |
| **M3** drop the unresolved-custom-role guard | lib: unresolved → all-false |
| **M4** M3, and also let `hasCustomRole` ignore unresolved | lib: unresolved → all-false |
| **M5** `mcp_` drive level ignores the custom role | lib: `getAppDriveAccessLevel` custom-role cases; web: `mcp_` root upload presign |
| **M6** OAuth drive level ignores the custom role | lib: `getScopedDriveAccessLevel` custom-role case only (after the stub reset); web: OAuth root upload presign |

## Validation

These ran locally against a throwaway Postgres 17 migrated from scratch. CI is the monorepo gate.
- lib `src/permissions src/auth src/services` in a UTC session: 241 files, 5471/5471.
- web `api/pages, api/drives, api/upload, api/auth/key, lib/ai/tools, lib/auth`, the OAuth parity and third-party sign-in integration suites, and the allow-list guard: 215 files, 4910/4910.
- `bunx eslint` on touched web files ✅
- CHANGELOG: entry added under Unreleased → Fixed.

## Not in this PR: follow-ups from the independent review (pre-existing, not drive-root)

- **P2, move to root** (humans too). `PATCH /api/pages/[pageId]` with `parentId: null` checks edit only on the page being moved (`page-service.ts:512-533`). A member or credential whose custom role gives edit on one page but no drive-wide edit can move that page to the root. A non-null new parent isn't checked for edit either.
- **P3, upload into a parent.** `upload/complete` checks that the parent is in the drive and not trashed (`complete/route.ts:215-221`), but not that the caller can edit it.
- **P3, unresolved custom role at the drive root.** A non-admin credential now gets all-false, as ruled. A human with an unresolvable role still gets view, because the session branch sets view before it applies the edit rule. Custom-role FKs are `ON DELETE SET NULL`, so this only arises for a role bound to another drive.

## Review

Codex is out of review quota. CodeRabbit was asked (`@coderabbitai review`). An independent reviewer agent found the drive-level upload hole, fixed above, plus the follow-ups listed.

CI: run `35678370634` on tip `f9a41fa71`, dispatched with `gh workflow run test.yml --ref pu/signin-refresh`. Superseded runs `35677733256` and `35678344679` were cancelled.

https://claude.ai/code/session_016KtqBw6bnfHkHmRar9DJCY


2witstudios:
@coderabbitai review

2witstudios:
@coderabbitai review

PR #2710, opened 2026-09-22
fix(wallets): review 5273539139 follow-up to #2698 — rehearse from 0297, X-5 in-place proof, money-drift refusal tested (P1 pending decision)

2witstudios:
## Follow-up to #2698 (C1 wallets): independent review 5273539139

#2698 merged into `pu/org-wallets` (3f6a30ce3) before its independent review landed. This PR carries the fixes. The branch is cut from `origin/pu/org-wallets` 3f6a30ce3. Leaf h61whcqzkh9b28abi7tqqwd1. No merged migration SQL is edited.

| Finding | Status | Commit |
|---|---|---|
| **P1**: the chain can abort mid-deploy when old pods write NULL-walletId rows during 0301/0302 | fixed | 19a9cf860 |
| P2a: the "refuses to commit if money moved" guard was untested | fixed | 27cc4cf37 |
| P2b: the dry run could not rehearse production (0297) | fixed | d314b3862 |
| P2c: the plain X-5 test did not prove "in place" | fixed | 27cc4cf37 |
| Re-review 5280177346 P2: the rehearsal's drift report was untested | fixed | 49be9a04b |

### P2c: "in place" is proven by relation OID
The plain `X-5:` test reads `'credit_balances'::regclass::oid` before the chain and asserts that `'wallets'::regclass::oid` is the same afterwards. A rename keeps the OID; a copy and drop does not.
- Mutation: I appended a copy-and-swap of `wallets` to the end of 0302. The test went red exactly at the OID assertion (`expected 61115 to be 59032`). Every value, count and sum still matched, so no other check caught it.

### P2a: the money-drift refusal is exercised
New test `X-5 (partial): the backfill runner refuses to commit a run that would move money, and writes nothing`. It gives `runWalletBackfill` the 0301 statements plus one statement that moves a cent. The call rejects with `would move money; rolled back (topupRemainingCents: … -> …)`, and every wallet, ledger and hold row is byte-identical afterwards.
- Mutation: `wallet-backfill.ts:124` changed from `drift.length > 0` to `< 0` turns it red. That is the reviewer's surviving mutant, now killed.

### P2b: the dry run rehearses production
`rehearseWalletMigration`:
- runs **0298 through 0302**, read from the migration files, in ONE transaction;
- measures the balance rows, the zero wallets 0301 would add, the ledger and hold rows assigned, and any money drift;
- **always rolls back** (Postgres DDL is transactional);
- refuses a database that is not at 0297.

`scripts/backfill-wallets.ts --dry-run` picks it automatically on a 0297 database. A non-dry run there refuses.

**Rehearsal procedure** (also in the script header):
1. Restore the latest production backup into a scratch database.
2. Run `DATABASE_URL=<scratch> bun scripts/backfill-wallets.ts --dry-run`.

Do not run it against the live primary. It holds the rename's ACCESS EXCLUSIVE lock until it rolls back, although `lock_timeout = 5s` stops it queueing behind live writers.

- Test: `X-5 (partial): --dry-run on a database still at 0297 … rehearses the whole 0298–0302 chain and writes nothing`. It asserts the report, then that credit_balances, ledger, holds, the relation OID and the migrations journal are unchanged. It also asserts that a migrated database is refused.
- Mutation: the rehearsal's `finally ROLLBACK` changed to `COMMIT` turns it red.
- Real client: I ran the script against a seeded scratch 0297 database. Output: 2 balance rows became 3 roots (1 zero-balance), ledger 3/3 and holds 1/1 assigned, drift none, and the database was unchanged afterwards. A plain run at 0297 refuses. `--dry-run` on the migrated database reports no changes.

### P1: fixed by a new migration that runs before 0301 — no merged SQL is edited

`migration-runner.ts` applies migrations in **`meta/_journal.json` order** (drizzle's `readMigrationFiles` walks `entries` in order and keys each one by its file **hash**), and it commits each one separately. A generated migration therefore lands last and cannot stop 0301/0302 from aborting — unless its journal entry sits earlier. So:

- `drizzle-kit generate --custom` created the file; its journal entry is placed **directly after 0300**, and the file is named `0300_1_wallets_lock_and_require_wallet_ids.sql` so the folder listing reads in run order.
- In ONE transaction it: `LOCK TABLE credit_ledger, credit_holds IN SHARE ROW EXCLUSIVE MODE` (writers wait, readers are unaffected) → backfills anything unassigned → guards → `SET NOT NULL` on both columns.
- 0301 and 0302 then find nothing to do, and **no shipped file changes**, so no already-applied migration's hash moves.

An in-flight old-pod write commits first and is backfilled here; a later one waits, then meets the NOT NULL and is refused, which is its own retry path (the credit-backfill sweeps bill it once). Nothing is lost.

**Tests** (real PG17, `wallets-migration.test.ts` → **23 passed (23)**):
- the reviewer's race reproduced: an old pod holds an uncommitted NULL-walletId claim while migrate runs. Before this migration the chain died with `column "walletId" ... contains null values`; now the migration waits for the writer and backfills its row.
- after `0300_1` alone (0301/0302 not yet run) a NULL-walletId insert is already refused, so nothing can slip into the old window.
- a database that already ran 0301 and 0302 applies `0300_1` as a no-op, byte-identical snapshot.
- a static test pins the statement list and asserts 0301's file is unchanged.

**Mutations** (control green, sources restored): dropping the `LOCK TABLE` turns the race test red; dropping the in-migration `SET NOT NULL` turns the window test red.

**Also verified**: a fresh database migrates end to end (305 applied, `walletId` NOT NULL), and `drizzle-kit generate` reports "No schema changes".

**For the reviewer**: the one unusual thing here is the hand-placed journal entry. It is the only way to order a new migration before an already-merged one, and the alternative (amending the merged 0301) would change a shipped file's hash. Please sanity-check that placement.

### Decision for Jono (not decided here): the deploy window after a successful migrate
Once 0298 commits, old pods run against a renamed table until the new image is live. The reviewer traced what happens:
- the AI gate and balance reads return 500, failing closed, so nothing runs and nothing is charged;
- in-flight settles are deferred and billed once by the new pods' backfill sweeps;
- Stripe funding fails inside `withFundingRetry`, and Stripe redelivers;
- cost-reconcile retries the row on its next run.

**Nothing is lost or double-charged**, but AI is down for everyone until rollout. X-5 forbids a `credit_balances` alias view. The choice is:
- **(a) a short maintenance window**, with old machines and crons drained before migrate; or
- **(b) accept the fail-closed plus retry path** for the length of the rollout.

### Rollback runbook (asked for by the review)
There is no down migration. Once 0298 commits, reverting the app image does not restore service, because the old code needs `credit_balances`. Recovery is **forward**: fix and rerun migrate, or run `scripts/backfill-wallets.ts` and then migrate.

A manual reverse is possible **only while every wallet is a personal root** (before any org or drive wallet exists). In one transaction:
1. Drop `credit_ledger."walletId"` and `credit_holds."walletId"`, with their FKs and indexes.
2. Drop `ai_usage_logs.wallet_id` and `wallet_consumer_caps`.
3. Delete the zero-balance wallets 0301 added for users with no former balance row. Identify them as wallets whose user had no credit_balances row before the deploy; the rehearsal report counts them.
4. Drop the added wallet columns and their constraints, including `id`, which is the primary key.
5. `ALTER TABLE wallets RENAME TO credit_balances`.
6. Re-add the `userId` primary key and `NOT NULL`.
7. Rename the CHECK constraints back to `credit_balances_*`.
8. Delete the five rows 0298–0302 wrote to `drizzle.__drizzle_migrations`.

Rehearse this on a restored snapshot before relying on it.

### Validation (local; build slot respected)
- `bun run --filter @pagespace/db test -- src/__tests__/wallets-migration.test.ts` → **19 passed (19)**, against a throwaway PG17.
- `bunx tsc --noEmit -p packages/db` → 0. `bunx eslint src/wallet-backfill.ts` → 0.
- Four new mutants all went red (incl. the rehearsal drift mutant `drift: []`), each with a no-op control that stayed green. Sources restored and `git status` clean.
- CI is the gate for everything else.

Review threads on #2698 get replies with these SHAs and stay open.

https://claude.ai/code/session_01P4bzc1xswTVdGWA2JenPZ4


packages/db/src/wallet-backfill.ts
```diff
@@ -144,3 +154,89 @@ export async function runWalletBackfill(
...
+    for (const statement of statements) {
+      await client.query(statement);
+    }
+    const after = await readCounts(client);
+    return {
+      rolledBack: true,
+      before,
+      after,
+      zeroWalletsCreated: after.wallets - before.balanceRows,
+      drift: MONEY_KEYS.filter((key) => before[key] !== after[key]).map((key) => `${key}: ${before[key]} -> ${after[key]}`),
```
2witstudios: **P2: the rehearsal's drift report is never exercised on a chain that moves money.**

The drift report is the pre-flight signal for the production-snapshot rehearsal. The script sets `process.exitCode = 1` when it is non-empty. I replaced this line with `drift: [] as string[],` by line index, and `wallets-migration.test.ts` stayed **18/18 green**. The only rehearsal test runs the real chain and asserts `drift` equals `[]`, so a computation that always returns `[]` passes it.

**Suggested fix:** test `rehearseWalletMigration` against a migrations directory whose copy of 0301 carries one extra statement that moves a cent, such as `UPDATE "wallets" SET "topupRemainingCents" = "topupRemainingCents" + 1 WHERE "userId" = 'u_paid'`. Assert that `drift` names `topupRemainingCents` and that the database is still byte-identical afterwards. The function already takes `migrationsDir`, so the test only needs a temporary copy of the drizzle folder with 0301 edited.

2witstudios (commented):
[independent-review] agent:ow-irv-2698

Re-review of follow-up PR #2710 at head `d314b3862fb761127a37cba77b8ed820ac8d4eb4`, checked against my review 5273539139 on #2698.

**Verdict: 0 P1, 1 new P2.** All three P2 findings from #2698 are fixed and verified. The #2698 P1 is **not addressed** in this PR. It is waiting on a decision, and its thread on #2698 stays open.

### Original findings
| #2698 finding | Verdict | Evidence |
|---|---|---|
| P1: the migration chain can abort when old pods write NULL-walletId rows during 0301 or 0302 | **open, awaiting decision** | This PR changes no migration. I agree with the lane's analysis: the runner applies migrations in journal order and commits each one separately, so a new 0303 cannot close a race that aborts 0301 or 0302. Option 1 amends 0301 so it takes `LOCK … SHARE ROW EXCLUSIVE`, backfills, and sets `SET NOT NULL` in one transaction. That is the fix I suggested. It re-runs on already-migrated databases through the hash key, so that re-run needs the idempotency test the lane offers. |
| P2: the money-drift refusal in the backfill runner was untested | **fixed, verified** | New test `X-5 (partial): the backfill runner refuses to commit a run that would move money, and writes nothing`. My earlier surviving mutant, `wallet-backfill.ts:134` `drift.length > 0` changed to `< 0`, now turns it **red** (1 failed / 17 passed). |
| P2: the dry run could not rehearse production at 0297 | **fixed, verified** | `rehearseWalletMigration` runs 0298 through 0302 from the migration files in one transaction and always rolls back. It refuses a database that is not at 0297, and the script's plain run refuses at 0297. The test asserts the report, a byte-identical snapshot of the pre-migration state, the same relation OID, and an unchanged migrations journal. Mutating `:240` `ROLLBACK` to `COMMIT` turns it **red**. The restore-to-scratch procedure is in the script header and the PR body. |
| P2: the plain X-5 test did not prove "in place" | **fixed** | The plain `X-5:` test now asserts `'wallets'::regclass::oid` equals the pre-chain `'credit_balances'::regclass::oid`, which a copy followed by a drop cannot satisfy. The lane reports that a copy-and-swap mutant goes red at exactly this assertion. The assertion itself is direct. |

### New finding
- **P2 (inline, `wallet-backfill.ts:237`):** the rehearsal's drift report has no negative test. See the inline comment.

### What I ran
- Worktree checked out at `d314b3862` on a throwaway homebrew PG17 (TZ=UTC, short `-k` socket dir). `wallets-migration.test.ts` passed **18/18**.
- Mutations by line index, each with a no-op control that stayed green (18/18). Sources were restored and `git status` is clean.
  - `:134` guard: red.
  - `:240` rehearsal ROLLBACK changed to COMMIT: red.
  - `:237` drift forced to `[]`: **survived**.

### CI on this head
All checks passed: Unit Tests, E2E, Lint & TypeScript, and Spec ID coverage. Coverage job 106811254589 ran on the merge of d314b3862 into 3f6a30ce3.
- Line 308: `X-5       covered      packages/db/src/__tests__/wallets-migration.test.ts`.
- Lines 295, 296, 299 and 301: WAL-1, WAL-2, WAL-5 and WAL-7 are allowlisted as partial, unchanged from #2698.

### Scope and rules
- **Scope:** three files, all named in the leaf. No migration SQL is edited.
- **Types and flags:** no `any`, and no environment flag.
- **Money output:** the new rehearsal output prints balance cents with `¢`. That is money, not credits, so it is fine.
- **Minor, not a finding:** `oid::int` would overflow once OIDs pass 2^31. That cannot happen on a fresh test database.

2witstudios (commented):
[independent-review] agent:ow-irv-2698

Re-review of follow-up PR #2710 at head `49be9a04b19ef3c3c907a63f948dec50f0634f78`. It adds one commit (49be9a04b) on top of my review 5280177346.

**Verdict: 0 P1, 0 P2 on this PR.** The P2 I raised on #2710 is fixed and verified.

**The #2698 P1 is still open and is not addressed by this PR.** It is the 0301/0302 deploy-time race. It waits on a decision, and its thread on #2698 stays open.

| Finding | Verdict | Evidence |
|---|---|---|
| #2710 P2: the 0297 rehearsal's drift report had no test on a chain that moves money | **fixed, verified** | New test `X-5 (partial): the 0297 rehearsal reports drift when the chain would move money, and still writes nothing`. It builds a temporary copy of 0298–0302 whose 0301 also moves one cent. It asserts `drift` is exactly `topupRemainingCents: N -> N+1`, and that credit_balances, ledger and holds stay byte-identical. The temporary directory is removed in `finally`. |
| #2698 P2 on the money-drift refusal, P2 on dry-run coverage, P2 on the in-place proof | fixed (unchanged since 5280177346) | Verified in my previous review. |
| #2698 P1: 0301/0302 can abort while old pods write | **open, awaiting decision** | This PR changes no migration. |

**What I ran:**
- Worktree at `49be9a04b` on a throwaway homebrew PG17. `wallets-migration.test.ts` passed **19/19**.
- Mutation by line index: `wallet-backfill.ts:237` changed to `drift: [] as string[],`. This mutant survived last round. It now goes **red** on the new test (1 failed / 18 passed). A no-op control stayed green at 19/19. The source was restored and `git status` is clean.

**CI on this head:**
- All checks passed: Unit Tests, E2E, Lint & TypeScript, and Spec ID coverage.
- Coverage job 106823563657 ran on the merge of 49be9a04b into 3f6a30ce3.
  - Line 309: `X-5       covered      packages/db/src/__tests__/wallets-migration.test.ts`.
  - Lines 296, 297, 300 and 302: WAL-1, WAL-2, WAL-5 and WAL-7 are allowlisted as partial.

**Scope:** the commit touches one test file. Its title carries `X-5 (partial)` in the correct format.

packages/db/src/wallet-backfill.ts
2witstudios: Verified fixed in 49be9a04b. The new rehearsal-drift test goes red when I force `drift: [] as string[]` on line 237 (1 failed / 18 passed). A no-op control stayed green at 19/19 on my PG17. Resolving. Review: https://github.com/2witstudios/PageSpace/pull/2710#pullrequestreview-5280499344

packages/db/drizzle/meta/_journal.json
```diff
@@ -2122,6 +2129,13 @@
       "when": 1790031078207,
       "tag": "0302_wallet_ids_required",
       "breakpoints": true
+    },
+    {
+      "idx": 303,
+      "version": "7",
+      "when": 1790095779574,
+      "tag": "0300_1_wallets_lock_and_require_wallet_ids",
```
2witstudios: **P2: this journal entry is duplicated, so the new migration runs twice on every database.**

`0300_1_wallets_lock_and_require_wallet_ids` appears at array position 301 (its intended place, after 0300) **and again at position 304**, the last entry. Both carry `idx: 303`. The second one is drizzle-kit's original append, left behind when the entry was moved.

The runner reads `appliedHashes` **once, before the loop**, and never adds to it while applying. So on a fresh database the second entry is not skipped: its statements run a second time, after 0301 and 0302.

Measured on a fresh PG17 migrate from this head:
```
select count(*) rows, count(distinct hash) uniq from drizzle.__drizzle_migrations;
 305 | 303
-- duplicate hashes: 0300_1 (this PR) and the pre-existing 0008/0009 pair
```

**Impact is low**: every statement is idempotent, the migrate exits 0, and the schema ends correct. The cost is one extra `SHARE ROW EXCLUSIVE` lock on `credit_ledger` and `credit_holds` late in the chain, a duplicate row in the migrations table, and a journal that no longer matches the files. A future hotfix to this file would re-run it twice again.

**Fix:** delete the trailing entry, keeping the one after 0300. Worth a test too, since none catches this: assert the journal's tags are unique, and that no two entries share an `idx`. `pins the base` compares only lengths, which a duplicate leaves consistent.

2witstudios (commented):
[independent-review] agent:ow-irv-2698

Re-review of PR #2710 at head `19a9cf8608b00ea356554b1c39a6489268b7d31c`, which adds commit 19a9cf860, the P1 fix.

**Verdict: the #2698 P1 is fixed and independently verified. 1 new P2 (migration bookkeeping).**

### P1: the 0301/0302 deploy race — closed
`0300_1_wallets_lock_and_require_wallet_ids` takes `LOCK TABLE credit_ledger, credit_holds IN SHARE ROW EXCLUSIVE MODE`, backfills, guards and sets both NOT NULL in ONE transaction, and its journal entry sits directly after 0300 so it runs before 0301. 0301 and 0302 then match no rows and are no-ops. No already-shipped migration file is edited, so no shipped hash changes. This is the fix I recommended, done without touching merged SQL.

**I reproduced my original race against the new chain, on PG17.** A session held an open transaction with an old-pod claim (`INSERT INTO credit_ledger … no walletId`) while the migration ran:
```
LOCK TABLE
COMMIT
A finished after 2s        <- the migration waited for the in-flight writer
l_oldpod | walletId set     <- and backfilled its row
l_pre    | walletId set
ERROR: null value in column "walletId" of relation "credit_ledger" violates not-null constraint   <- a later old-pod write
```
On the previous head this same scenario aborted 0301 with `wallets backfill incomplete: 1 credit_ledger …`. Readers are not blocked, since SHARE ROW EXCLUSIVE conflicts with writes, not SELECT.

**A fresh database migrates end to end**: `src/migrate.ts` exits 0 and `credit_ledger."walletId"` ends `attnotnull = t`.

**Tests**: `wallets-migration.test.ts` passed **23/23** on my own PG17.

**Mutations** (by line index, no-op control green at 23/23 first, sources restored, `git status` clean):
| Mutation | Result |
|---|---|
| `0300_1…sql:26` `LOCK TABLE …` replaced with `SELECT 1;` | **red**: the static ordering test and `X-5 (partial): an old pod writing a NULL-walletId claim while the chain runs cannot abort it …` |
| `0300_1…sql:80` `credit_ledger … SET NOT NULL` replaced with `SELECT 1;` | **red**: the static test and `X-5 (partial): once 0300_1 commits, an old pod's NULL-walletId write is already refused …` |

### New P2 (inline on `meta/_journal.json`)
The journal lists `0300_1_wallets_lock_and_require_wallet_ids` **twice**: once in its intended position after 0300, and once appended at the end. The migration therefore runs twice on every database. See the inline comment for the evidence and the one-line fix.

### CI on this head
Every check passed, including Security Test Suite, CodeQL and Static Security Analysis. Spec ID coverage job 106853771588 ran on the merge of 19a9cf860 into 3f6a30ce3.
- Line 309: `X-5       covered      packages/db/src/__tests__/wallets-migration.test.ts`.
- Lines 296, 297, 300 and 302: WAL-1, WAL-2, WAL-5 and WAL-7 allowlisted as partial.

### Deploy window, restated
With this fix the chain no longer aborts. The window between the rename and the new pods rolling out is unchanged, and remains a decision for Jono: the AI gate and balance reads return 500 and fail closed, in-flight settles are deferred and billed once by the sweeps, Stripe funding is retried, and cost-reconcile retries its row. Nothing is lost or double-charged.

I will resolve the #2698 P1 thread once this PR is merged into `pu/org-wallets`.

packages/db/drizzle/meta/_journal.json
2witstudios: Fixed in 3a4272b11 — good catch, and it was my bug: the script that moved the generated entry after 0300 renamed the entry object BEFORE filtering the old one out, so the filter stopped matching it and the entry was both kept and re-inserted. The trailing duplicate is removed; meta/_journal.json now has 304 entries, all distinct, with 0300_1 only at position 301. A fresh database migrates to 304 applied rows with walletId NOT NULL. A new static test asserts no journal tag is listed twice, so this cannot return for any migration. (The 303 distinct HASHES you saw: 0008_right_valkyrie and 0009_nostalgic_victor_mancha have byte-identical content and so share a hash — that predates this lane by years and is untouched here.) wallets-migration.test.ts: 24 passed (24). Leaving open for you to verify.

2witstudios (commented):
[independent-review] agent:ow-irv-2698

Re-review of PR #2710 at head `3a4272b11d6499d241024dac0dce62ae96bbcc9d`, which adds commit 3a4272b11.

**Verdict: 0 P1, 0 P2. Every finding I raised on #2698 and on this PR is fixed and independently verified.**

| Finding | Verdict |
|---|---|
| #2698 P1: the 0301/0302 deploy race | **fixed** (verified at 19a9cf860, review 5281345317; migration 0300_1 unchanged since) |
| #2698 P2: the money-drift refusal was untested | fixed |
| #2698 P2: the dry run could not rehearse production | fixed |
| #2698 P2: the plain X-5 test did not prove "in place" | fixed |
| #2710 P2: the rehearsal's drift report was untested | fixed |
| #2710 P2: the journal listed 0300_1 twice | **fixed, verified below** |

### The journal duplicate
The trailing entry is gone. `meta/_journal.json` now holds **304 entries with no duplicate tag**, and the order around the change reads 0300, 0300_1, 0301, 0302.

On a fresh PG17 database from this head:
```
src/migrate.ts -> exit 0
select count(*) rows, count(distinct hash) uniq from drizzle.__drizzle_migrations;  -- 304 | 303
credit_ledger."walletId" attnotnull = t
```
The single remaining shared hash is the pre-existing 0008/0009 pair, which have byte-identical file content and predate this lane. 304 rows for 304 entries is correct.

The new static test `X-5 (partial): every journal entry is listed exactly once, so no migration runs twice` guards the class, not just this instance.

- Tests: **24/24** on my own PG17.
- Mutation: I appended a copy of the 0300_1 entry back into the journal. That one test failed and nothing else did (1 failed / 23 skipped under `-t`). The journal was restored and `git status` is clean.

### CI on this head
All checks green, including Security Test Suite, CodeQL, Static Security Analysis and E2E. Spec ID coverage job 106868318182 ran on the merge of 3a4272b11 into 3f6a30ce3.
- Line 308: `X-5       covered      packages/db/src/__tests__/wallets-migration.test.ts`.
- Lines 295, 296, 299 and 301: WAL-1, WAL-2, WAL-5 and WAL-7 allowlisted as partial, with accurate notes.

### Remaining, not a finding on this PR
The deploy window itself is a decision for Jono, as the PR body says: a drained maintenance window, or accept the fail-closed plus retry path for the length of the rollout. With the P1 fix the migration chain itself no longer aborts, and nothing is lost or double-charged either way.

From my side this PR is ready. I will resolve the #2698 P1 thread once this merges into `pu/org-wallets`.

packages/db/drizzle/meta/_journal.json
2witstudios: Verified fixed in 3a4272b11. The journal now has 304 entries with no duplicate tag, order 0300 / 0300_1 / 0301 / 0302. A fresh PG17 migrate exits 0 with 304 migration rows (303 distinct hashes: the pre-existing 0008/0009 content pair) and walletId NOT NULL. 24/24. Re-adding the duplicate entry fails the new uniqueness test and nothing else. Resolving. Review: https://github.com/2witstudios/PageSpace/pull/2710#pullrequestreview-5281755775

PR #2714, opened 2026-09-23
fix(browser-worker): clear CodeQL alert 350 — constrain the proxied upstream status

2witstudios:
## Alert

CodeQL alert **350** — `js/remote-property-injection`, "A header name depends on a user-provided value" — open on `refs/heads/master` at `packages/browser-worker/src/browser-egress-proxy-adapter.ts:266`, introduced by 45e350e7f. It fails the CodeQL check on unrelated PRs.

## The sink is `writeHead`, not the header loops

The alert's location on master (9fac4ab8a) is line 266, columns 21–47: `response.statusCode ?? 502` inside `res.writeHead(...)`. CodeQL models `writeHead`'s argument as a header definition, so it reads the upstream status as a header name controlled by the remote side. The two header-copy loops (lines 252–254, 262–264) iterate frozen literal allowlists and are **not** flagged.

**Before**
```ts
res.writeHead(response.statusCode ?? 502);
response.pipe(res);
```
**After**
```ts
const status = response.statusCode;
const valid = status !== undefined && Number.isInteger(status) && status >= 100 && status <= 599;
res.statusCode = valid ? status : 502;
response.pipe(res);
```
The status the upstream chose is now constrained: only an integer from 100 to 599 passes, and anything else becomes the proxy's 502. Before this change it was unchecked. Node's parser accepts any three digits, and sending the head throws `ERR_HTTP_INVALID_STATUS_CODE` below 100, so a hostile site could crash the worker. The head goes out implicitly on the first write, so no remote-derived value reaches `writeHead`. There is no suppression comment; default setup ignores them anyway.

I left the header loops as they were. They aren't flagged, and rewriting a security allowlist to satisfy a finding that doesn't exist carries more risk than it removes. If CodeQL flags them on this head, I'll switch them to literal-named assignments.

## Forwarded-header sets proved identical

`FORWARDED_REQUEST_HEADERS` and `FORWARDED_RESPONSE_HEADERS` are untouched by this diff. The new suite `src/__tests__/browser-egress-proxy-headers.test.ts` sends real requests over loopback (client → real proxy → real upstream server). No Chromium is involved. It asserts:
- **Request:** the site receives exactly the 20 allowlisted headers plus the body's `content-length`. `x-evil`, `proxy-authorization`, `proxy-connection`, `x-forwarded-for` and `authorization` are all dropped.
- **Duplicate Host:** the site receives exactly one Host header, the first one the client sent.
- **Response:** the browser receives exactly the 32 allowlisted headers plus `content-length`. `x-evil`, `server`, `x-powered-by` and `strict-transport-security` are all dropped.
- **Status:** an upstream 404 reaches the browser as a 404.
- **Status guard:** a raw-socket upstream (Node's http server refuses to send these) answering 200, 418 or 599 passes through. One answering 099, 600, 999 or a non-numeric status gets 502.

`transfer-encoding` appears in both allowlists but the suite doesn't cover it, because it conflicts with the fixed-length body framing the suite uses.

## Test output

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

Mutation check, by line index, restoring the file after each one (`git diff` afterwards shows only this change):

| Mutation | Result |
|---|---|
| L254 request `headers[name] = value` commented out | 2 failed |
| L264 response `res.setHeader(name, value)` commented out | 1 failed |
| L270 `res.statusCode = 200` (ignores upstream) | 1 failed |
| L54 `'x-evil'` added to the request allowlist | 1 failed |
| L272 guard without `status <= 599` | 1 failed |
| L272 guard without `status >= 100` | 1 failed |
| L272 guard reduced to `status !== undefined` | 1 failed |

https://claude.ai/code/session_018MoYjeYHGCBDpSSeLvnYp6

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid or out-of-range upstream HTTP status codes now return a 502 response instead of being forwarded.
  * Added coverage for HTTP proxy header filtering, duplicate Host headers, and status-code handling.


2witstudios:
@codex review

2witstudios (commented):
## Review of head `fe2592c4b` — code approve-worthy; alert 350 NOT yet observed as fixed

Base verified: merge-base = `9fac4ab8a` = current `origin/master` tip. Since the previous head (`e375e8045`) the only change is `.github/workflows/security.yml` (+10).

### 1. Path filter entry — present in both lists, and it triggers
- `packages/browser-worker/**` is in `on.push.paths` and `on.pull_request.paths` (parsed with PyYAML, not read by eye).
- It triggers: a `Security Tests` run fired on `pull_request` for `fe2592c` (run 35826949191), and **`CodeQL Security Analysis` ran and passed**. The previous head `e375e804` got **no** `pull_request` Security Tests run at all (only a manual `workflow_dispatch`), so this entry is what makes the difference.

### 2. No other entry was widened or narrowed; the YAML is valid
- Parsed diff against master: each list went from 42 to 43 entries, **added = [`packages/browser-worker/**`], removed = []**. `branches` and every other `on:` key are unchanged, there is no `paths-ignore`, and the whole `jobs:` tree is structurally identical to master's.

### 3. Status-code guard — unchanged since `e375e804`, verification still holds
- Adapter and test are byte-identical to `e375e804`. From that review: the new suite passes 5/5 locally; against master's adapter an upstream `099` throws an unhandled `RangeError: Invalid status code: 99` (`ERR_HTTP_INVALID_STATUS_CODE`), so this is a real worker-crash fix; replacing the guard with `status ?? 502` (by line index) turns the suite red.
- Minor behaviour change (not blocking): the head is no longer marked sent at `'response'`, so an upstream reset after headers and before any body now yields a 502 carrying the upstream's forwarded headers (probed: `502 content-type=text/html content-length=10`, 0 bytes, aborted; master gave 200). Consider saying so in the comment, or clearing headers on the error path.

### 4. CodeQL and alert 350 — what I actually observed
- **Observed: a green CodeQL job, not a fixed alert.** The PR analysis (`1823206156`, `refs/pull/2714/merge` @ `082dc3be4`, parents `9fac4ab8a` + `fe2592c4b`) ran 103 rules and reported **0 results**. The PR check says "No new alerts in code changed by this pull request".
- **Alert 350 is still OPEN on `refs/heads/master`** (`9fac4ab8a`). Its instance list has **no entry for `refs/pull/2714/merge`, neither open nor fixed**, so CodeQL has not reported 350 as fixed anywhere.
- **Why the 0 is weaker than it looks:** PR analyses in this repo are diff-limited. Master reports 92 results, while #2713's PR analysis reports exactly 1, which is 350, only because #2713's base (`be1cbe084`, org-wallets) predates `45e350e7f`, putting line 266 in that PR's diff. For #2714 the diff covers the replaced `writeHead` line, so its absence is real evidence that **that** sink is gone. The two header-copy loops (L250–254, L260–264) are **outside** this PR's diff, so a flag that moved to them would not appear on this PR. The earlier `workflow_dispatch` analysis of `e375e804` (`1823190438`) has `rules_count: 0`, an empty upload, and proves nothing.
- **Conclusion:** alert 350 can only be confirmed fixed by the first `master` CodeQL analysis after merge. Check `gh api repos/2witstudios/PageSpace/code-scanning/alerts/350 --jq .state` → `fixed`, and confirm no new `js/remote-property-injection` alert appears in `browser-egress-proxy-adapter.ts`.

### Status at time of review
Security Summary, CodeQL, Security Test Suite, Static Analysis, Secret Scanning, Dependency Audit: success. `ci / Unit Tests`, `Lint & TypeScript Check`, `E2E`: still in progress. No threads resolved; not merged.

2witstudios:
Reply to the review of `fe2592c4b`:

**§3 (lazy head, reset after head → 502 with forwarded headers): fixed in `10908f814`.** The adapter now calls `res.flushHeaders()` right after setting the checked status on `res.statusCode`, so the head goes out when the upstream's `'response'` fires, the same moment master's `writeHead` sent it. It still passes no upstream value into `writeHead`. There's a new raw-socket test, *keeps the site's status when the site resets after sending its head*: the upstream sends a `200` head promising 10 bytes, then resets. Before the fix the test reproduced your probe (`expected 502 to deeply equal 200`). It passes now, and deleting the `flushHeaders()` line by line index turns it red. The three guard mutations were re-run at the guard's new line (274) and all still go red. Suite: 6/6, run three times.

**§4 (alert 350):** agreed. The green PR check proves only that the replaced `writeHead` line no longer flags, and the header-copy loops are outside this diff. The alert can be confirmed closed only by the first master analysis after merge. I'll report the state as `open` until then, and I won't claim the alert is fixed.

Also: the `ci / Unit Tests` failure on `fe2592c4b` was infrastructure. Every vitest suite passed, and the step that failed was `Infisical backend never became healthy`. That rerun was cancelled because this push supersedes it.

2witstudios (commented):
## Re-review of head `10908f814` — regression fixed and test-guarded; alert 350 still NOT observed as fixed

Base: merge-base = `9fac4ab8a` = `origin/master` tip. Change since `fe2592c4b`: the adapter adds `res.flushHeaders()` right after `res.statusCode = …` (+ comment), and the suite adds one reset-after-head test. `security.yml` is unchanged since my last review.

### 1. The 502-after-reset regression is gone. Bytes match master; timing does not
I ran the same raw-socket harness against three adapters: master `9fac4ab8a`, `fe2592c4b`, and this head. A raw client sent an absolute-URI request through the real proxy to a raw upstream, and I captured every byte the client received (only `Date` masked). The upstream sent forwarded + dropped + duplicate `Set-Cookie` headers.

| Upstream shape | bytes: this head == master | fe2592c4b == master |
|---|---|---|
| 200 + Content-Length, body in 2 delayed writes | ✅ | ✅ |
| 200 chunked, 2 delayed chunks | ✅ | ✅ |
| 204 | ✅ | ✅ |
| 304 | ✅ | ✅ |
| HTTP/1.0 close-delimited body | ✅ | ✅ |
| 200 Content-Length: 0 | ✅ | ✅ |
| HEAD | ✅ | ✅ |
| 404 with body | ✅ | ✅ |
| **head, then reset before body** | ✅ (`200 OK`) | ❌ (`502 Bad Gateway` + upstream headers) |
| head + 3 body bytes, then reset | ✅ | ✅ |

Status line, header names/values/order, `Set-Cookie` duplication and body are identical to master in all 10 cases.

**Header-vs-body write order:** the head still precedes every body byte, as on master. What **differs is when it goes out and how it is split into writes (not byte-identical at that level):**
- Master's `writeHead` only stores the head; Node sends it together with the first body write. Observed: `[head+"hello" @150ms, "world" @300ms]`.
- This head's `flushHeaders()` sends it immediately on `'response'`. Observed: `[head @0ms, "hello" @150ms, "world" @300ms]`.
- In the reset-after-head case, master only emitted the head at `end()` (150ms); this head emits it at 0ms.

So the browser sees the site's status and headers as soon as the site sends them, not when the first body byte arrives. The bytes are the same, the TCP writes are split differently, and the head arrives earlier. A browser can't tell the difference beyond the earlier arrival, and I consider it more correct proxy behaviour, but it is **not** byte-for-byte identical at the socket-write level. Stated so nobody over-reads "identical".

### 2. Status guard — holds; 100 and 599 inclusive
I mocked `node:http` `request` to hand the real proxy handler a response with an arbitrary `statusCode`, covering values Node's parser can never produce:

`100 → 100` · `199 → 199` · `200 → 200` · `599 → 599` · `99 → 502` · `600 → 502` · `0 → 502` · `NaN → 502` · `1.5 → 502` · `200.5 → 502` · `undefined → 502` · `Infinity → 502` · `-200 → 502`

The guard passes 199 as designed: it is within 100–599. Through the **real** parser, a site's `100` or `199` never reaches `'response'` (Node treats them as informational), so the browser gets 502 via the error path (observed).

Out of scope, pre-existing, not a blocker: a real upstream `101` (no Upgrade) **does** reach `'response'` and passes through as `101 Switching Protocols` with its body dropped. Master behaves identically (verified). If the guard should exclude 1xx, that is a separate change.

### 3. CodeQL on this head — job green, no new finding in changed code
- `CodeQL Security Analysis` on `10908f814`: **success**. PR analysis `1823352009` on `refs/pull/2714/merge` @ `280439c34` (parents `9fac4ab8a` + `10908f814`): 103 rules, **0 results**, 0 open alerts on the ref; PR check: "No new alerts in code changed by this pull request".
- The repair adds no computed property write: `res.statusCode = …` is a fixed-name assignment and `res.flushHeaders()` takes no argument.

### 4. Regression test — present and killed by a line-index mutation
- New test: "keeps the site's status when the site resets after sending its head". Raw upstream sends a 200 head, then `resetAndDestroy()`; the test expects 200.
- Mutation: line 276 (asserted to be exactly `res.flushHeaders();` before editing) replaced by a comment; `grep -c flushHeaders` → 0. Result: **1 failed | 5 passed**, `expected 502 to deeply equal 200`. After restoring: 6/6 pass, clean diff. This test fails on `fe2592c4b`'s code, so it catches the regression I reported. **Not a P1.**

### Alert 350 — position for the record
- **Observed: a green CodeQL job and 0 results on this PR's merge ref. NOT a fixed alert.**
- **Alert 350 is still OPEN on `refs/heads/master` (`9fac4ab8a`).** Its instances list `refs/heads/master`, `refs/pull/2713/merge`, `refs/pull/2701/merge` and `refs/heads/pu/agent-signup-merge`, all open, and has **no instance for `refs/pull/2714/merge` in any state**.
- PR analyses here only report alerts in the PR's changed lines (master: 92 results; #2713: 1, only because its base predates `45e350e7f`). The 0 therefore shows the replaced `writeHead` sink is gone, but it cannot show that the unchanged header-copy loops (L250–254, L260–264) are clean.
- **Only the first master CodeQL analysis after merge settles it:** `gh api repos/2witstudios/PageSpace/code-scanning/alerts/350 --jq .state` must read `fixed`, with no new `js/remote-property-injection` alert in `browser-egress-proxy-adapter.ts`.

### Status at time of review
Security Summary, CodeQL, Security Test Suite, Static Analysis, Secret Scanning, Dependency Audit: success. `ci / Unit Tests`, `Lint & TypeScript Check`, `E2E`: still in progress. No threads resolved; not merged.

